PR #181, opened 2014-11-04
Add spm support

sorrycc:
A nicer package manager: http://spmjs.io
Documentation: http://spmjs.io/documentation

http://spmjs.io/package/immutable

---

It is a package manager based on [Sea.js](https://github.com/seajs/seajs) which is a popular module loader in China.

spmjs focus in browser side. We supply a complete lifecycle managment of package by using [spm](https://github.com/spmjs/spm/tree/master).

> If you need ownership of immutable in spmjs, I can add it for your account after signing in http://spmjs.io.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,5 +62,14 @@
     "sequence",
     "iteration"
   ],
-  "license": "BSD"
+  "license": "BSD",
+  "spm": {
+    "main": "dist/immutable.js",
+    "ignore": [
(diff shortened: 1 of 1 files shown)


leebyron:
Going to close this since it can't be merged. I don't think I can handle the mental bandwidth of keeping more package managers up to date than we currently have.

PR #211, opened 2014-11-21
[Suggestion] for ordered groupBy

abergs:
I am grouping a sorted list, and want the map created by groupBy to reflect the order in the list.
Locally i just switched the map to a OrderedMap and the behavior was working as expected.

I am typing this out directly in GitHub so it might contains bugs. My local modifications were naive and always did OrderedMap, which might not be as lean/performant as you want.

Any thoughts?


leebyron:
Sorry for not seeing this before.

This seems like a good idea, but I'm a little bit concerned about applying the heavy weight of OrderedMap without option. I think it may be best to remain ordered though.

PR #227, opened 2014-12-04
Inherit from base types

tgriesser:
Here's a first pass at making it possible to inherit from `Map`, `List`, `Set`, `OrderedMap`, etc.

Mostly putting this here for discussion, as there are probably a few ways to attack this one, but this seems to be working quite well in my preliminary use. It also has the benefit of eliminating `Object.create(MapPrototype)` in favor of `new Ctor`, which gives better inspection in the chrome console (and [potentially perf?](http://jsperf.com/obj-create-vs-new/4))

The inherited types must initially be constructed with `new`, and when inheriting via class style, it's necessary to return from the constructor:

```js
class User extends Immutable.Map {
  
  constructor(value) {
    return super(value)
  }
 
  fullName() {
    return this.get('firstName') + ' ' + this.get('lastName');
  }

}
```

If you think this looks good, I could beef up the tests a bit.


tgriesser:
So I updated this to latest and ran that new `perf` suite, looks like this branch takes a pretty substantial hit ~50% or more on each of the tests. 

So unfortunately this doesn't look like it's the best approach, though figuring out some way to do this eventually would be really cool - the ability to have custom types for `Maps`/`OrderedMaps` has proved quite convenient in our application.

leebyron:
Yeah, I'm really sorry for not keeping you apprised, Tim. The benchmarking
suite was a pre-req for this to make sure we don't regress. I'm sure
there's a way to do it, and now we can measure different approaches!

Lee Byron
[ lee@leebyron.com | (317) 460-9114 | www.leebyron.com ]

tgriesser:
> Yeah, I'm really sorry for not keeping you apprised, Tim.

No worries at all! I've been getting along great with my fork, just wanted to get things up to date and give an update with my findings. I'm excited there's a benchmarking suite to be able to test as I go - I'll probably play around with a few different ideas here.

It also might be neat to explore how cljs handles analogous behavior with protocols, etc.

tgriesser:
Superseded by #368

Issue #249, opened 2014-12-17
Cursor.set(value: any): Cursor


Dashed:
Can we have a way to set current value of cursor?

```js
Cursor.set(value: any): Cursor
```

It's similar to `update(updater: (value: any) => any): Cursor`, but with being able to set `undefined`.

confiks:
I would like to be able to do this as well. Makes it easy to alter values stashed away in nested structures when passing cursors as props through React.

geirsagberg:
+1, this would be really handy. Shouldn't need a full update method call when I don't care about the current value of the cursor.

geirsagberg:
Added PR #516. 

On another note, would it make sense to have a matching cursor.get()-method? It would basically be an alias for deref(). I have the code, can add it to the PR if you want.

PR #250, opened 2014-12-17
Be able to set value of a cursor

Dashed:
Add:
```js
Cursor.set(value: any): Cursor
```

Similar to `update(updater: (value: any) => any): Cursor`, but with being able to set `undefined`.

Closes https://github.com/facebook/immutable-js/issues/249

### DIFF
--- a/contrib/cursor/__tests__/Cursor.ts
+++ b/contrib/cursor/__tests__/Cursor.ts
@@ -157,6 +157,25 @@ describe('Cursor', () => {
     );
   });
 
+  it('can set value of a cursor', () => {
+    var onChange = jest.genMockFunction();
+
+    var data = Immutable.fromJS(json);
+    var aCursor = Cursor.from(data, onChange);
+
+    expect(aCursor.set(
+        Immutable.fromJS({ a: 1 })
+      ).deref()).toEqual(
+      Immutable.fromJS({ a: 1 })
+    );
+
+    expect(onChange).lastCalledWith(
+      Immutable.fromJS({ a: 1 }),
+      data,
+      []
+    );
+  });
+
   it('creates maps as necessary', () => {
     var data = Immutable.Map();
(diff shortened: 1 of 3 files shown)


confiks:
I would like to be able to do this as well. Makes it easy to alter values stashed away in nested structures when passing cursors as props through React.

confiks:
This pull request includes a lot of code dealing with the case of setting a cursor to the value of `undefined`. I don't really see why that is needed, but I might be missing something.

Below, I'm proposing another version which is a lot simpler and as far as I've tested works correctly. If you think this would be suitable (@leebyron), I could create a new pull request.

```diff
 IndexedCursorPrototype.set =
 KeyedCursorPrototype.set = function(key, value) {
-  return updateCursor(this, function (m) { return m.set(key, value); }, [key]);
+  if(arguments.length === 1) {
+    return updateCursor(this, function() { return key; }, []);
+  } else {
+    return updateCursor(this, function (m) { return m.set(key, value); }, [key]);
+  }
 }
```

Dashed:
@confiks By all means, please go ahead. I'm interested to see what others have in mind for this use case.

wmertens:
@confiks I like your version, PR?

Dashed:
@confiks you can now just use `Cursor.setIn([], value)`; since that API wasn't available at the time of this PR.

I can look into updating this PR to have `Cursor.set(value)` be a sugar to `Cursor.setIn([], value)` when I get around to it.

Feel free to PR the sugar; and then I can close this one.

geirsagberg:
I created a PR https://github.com/facebook/immutable-js/pull/516 with the code from your comment, @confiks. I preferred the `updateCursor` syntax rather than `getIn`, because I feel it is more semantically correct.

On another note, would it make sense to have a matching `cursor.get()`-method? It would basically be an alias for `deref()`. I have the code, can add it to the PR if you want.

Dashed:
Closing in favour of #516

PR #270, opened 2014-12-31
fix: change require to immutable library name

istarkov:
1) declaration in typescript is `import Immutable = require('immutable');`
2) it's not easy work with such `require('../..')` create external browserify bundle with immutable and contrib/cursor both included

### DIFF
--- a/contrib/cursor/index.js
+++ b/contrib/cursor/index.js
@@ -15,7 +15,7 @@
  * If you wish to use it in the browser, please check out Browserify or WebPack!
  */
 
-var Immutable = require('../../');
+var Immutable = require('immutable');
 var Iterable = Immutable.Iterable;
 var Iterator = Iterable.Iterator;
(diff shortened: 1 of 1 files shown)


istarkov:
@leebyron 
I have an issue with browserify, for example if i create bundle `browserify -r immutable -r 'immutable/contrib/cursor' > vendor.js`
and try to use it in other bundle `browserify app.js -x immutable -x 'immutable/contrib/cursor' > app_bundle.js`
i've got an error `Cannot find module '../..'`

It looks like in this situation browserify can't detect that require('../..') is the same as require('immutable'),
and as immutable v3 working with cursors need to include additional module, i think this module can be seen as like external library, so it need to have full `require('immutable')` declaration inside

And in typescript definition i  see full require('immutable') but not in js code.
I think it's browserify bug, but it can be easy avoided if you allow this change.

Issue #226, opened 2014-12-03
updateIn doesn't update a record when given a notSetValue?


fson:
I have a nested structure like this:

``` js
var Immutable = require('immutable');
var Record = Immutable.Record;
var Map = Immutable.Map;

var State = Record({
  records: Map()
});
var Item = Record({
  name: null
});
var state = new State();
var notSetValue = new Item({name: 'foo'});
```

I try to set the name of the item at the key `records.foo` to "foo". If the key does not exist, it's first initialized to `notSetValue`. Looks like `updateIn` doesn't update the state record, if the return value of `updater` fn would be equal to `notSetValue`?

``` js
state.updateIn(['records', 'foo'], notSetValue,
               function (value) { return value.set('name', 'foo'); });
// => Record { newRecord: null, records: Map {} }
//                                            ^ no 'foo' here!
```

If I use a different `name` so it's not equal to `notSetValue`, it works correctly, however:

``` js
state.updateIn(['records', 'foo'], notSetValue,
               function (value) { return value.set('name', 'foobar'); });
// => Record { newRecord: null, records: Map { foo: Record { name: "foobar" } } }
```

I'd like to know if this is a bug in `updateIn` or if I'm using it wrong. It seems like an unexpected behavior.

EDIT: fixed `defaultItem` -> `notSetValue` in the example.

AsaAyers:
In the example above it seems like `notSetValue` and `defaultItem` are the same thing. by changing `notSetValue` to `defaultItem` I can reproduce this behavior.

AsaAyers:
https://github.com/facebook/immutable-js/blob/master/src/Map.js#L752-L754
```
    var existingValue = isNotSet ? notSetValue : existing;
    var newValue = updater(existingValue);
    return newValue === existingValue ? existing : newValue;
```

I believe the bug is here. `existingValue` is actually the `notSetValue`, but when comparing the new value to the old value it finds them to be the same, so it returns `existing`. I don't understand why this wouldn't just always return `newValue`. If they're `===` it seems safe to return.

leebyron:
Sorry for the delay in answering here.

I'm not sure what the value of `defaultItem` is, but if it causes the updating function to behave as an identity function, then no update will occur by design.

Since this task was opened, I've updated the documentation of `updateIn` to hopefully explain this behavior a bit more clearly.

AsaAyers:
[...]

PR #290, opened 2015-01-20
Fix facebook/immutable-js#226

AsaAyers:


leebyron:
Let me investigate this a bit more, I'm surprised no tests failed as a result of this change. It's currently documented to behave as coded before this change.

leebyron:
Yeah, this breaks tests for `updateIn`. I'll update the Travis CI script to ensure it builds the distribution before it runs tests to catch these.

If you run `grunt` before pushing, you'll see the affected test cases.

AsaAyers:
oh, I just ran `npm test` before I made my change and then `npm test` again after making my change. Thanks.

PR #295, opened 2015-01-26
Avoid making x properties non-enumerable

atalmadge:
I'm seeing cases where Object.keys({x:1, y:2}) is yielding ['y'].


leebyron:
I changed the new key to `"@"` which is even less likely to cause conflict. Thanks again for digging into this.

PR #297, opened 2015-01-27
makeCursor may return incorrect cursor type

### LINKED ISSUE #318: cursor.cursor('path') to immutable.List returns wrong cursor type
```js
var data = Immutable.fromJS({ a: [1, 2, 3] });
var cursor = Cursor.from(data);
var deepCursor = cursor.cursor('a');
```

deepCursor will be KeyedCursor, not IndexedCursor, due to the way `subCursor()` calls `makeCursor()` and how `makeCursor()` checks for optional arguments.

#297 fixes this issue.

jcranendonk:
Certain code paths call makeCursor with four arguments, with 'value' undefined. Since arguments.length is still four in this case, the value is not retrieved from rootData, and an incorrect cursor type may be generated.

Fixes #318 

(As a side effect, this change may also make makeCursor [eligible for VM optimizations](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#31-reassigning-a-defined-parameter-while-also-mentioning-arguments-in-the-body-typical-example).)

### DIFF
--- a/contrib/cursor/__tests__/Cursor.ts
+++ b/contrib/cursor/__tests__/Cursor.ts
@@ -49,6 +49,13 @@ describe('Cursor', () => {
     expect(deepCursor.deref()).toBe(data.getIn(Immutable.fromJS(['a', 'b'])));
   });
 
+  it('cursor return new cursors of correct type', () => {
+    var data = Immutable.fromJS({ a: [1, 2, 3] });
+    var cursor = Cursor.from(data);
+    var deepCursor = <any>cursor.cursor('a');
+    expect(deepCursor.findIndex).toBeDefined();
+  });
+
   it('can be treated as a value', () => {
     var data = Immutable.fromJS(json);
     var cursor = Cursor.from(data, ['a', 'b']);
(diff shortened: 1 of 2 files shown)


contrib/cursor/__tests__/Cursor.ts
```diff
@@ -49,6 +49,13 @@ describe('Cursor', () => {
     expect(deepCursor.deref()).toBe(data.getIn(Immutable.fromJS(['a', 'b'])));
   });
 
+  it('cursor return new cursors of correct type', () => {
+    var data = Immutable.fromJS({ a: [1, 2, 3] });
+    var cursor = Cursor.from(data);
+    var deepCursor = <any>cursor.cursor('a');
+    expect(deepCursor.findIndex).toBeDefined();
```
jcranendonk: Probably not the best way to test this. I'm open to suggestions.

tcoopman:
Any updates on this?

leebyron:
Thanks for adding this. I'm a bit behind on reviewing PRs, but I promise to dig into it soon.

jeffbski:
+1

jeffbski:
I created an alternate PR for this #381 which continues to use `argument.length` as the condition since I think it might be valid for value to be `undefined` in some circumstances. `subCursor()` simply needs to do an argument.length check as well and call makeCursor appropriately. I reused @jeroencranendonk-wf  's test.

leebyron:
Closing in favor of @jeffbski's fix.

PR #299, opened 2015-01-28
Add missing semicolon

gabelevi:
Noticed a missing semicolon when I was converting this file to a format that [Flow](https://github.com/facebook/flow) can understand


leebyron:
Sweet thanks

Issue #285, opened 2015-01-15
Map#map() always return a new Map even without changes


jeanlauliac:
Consider the following piece of code:

```js
(function () {
  var m = new Immutable.Map({foo: 1});
  var n = m.map(function (e) { return e; });
  console.log(n === m);
})()
```

It prints `false`. However, the map data hasn't changed, meaning we loose the benefit of immutability here. If no element changes, perhaps it would make sense to return the same `Map`. What do you think?

leebyron:
In theory I agree, however I'm not sure how to go about actually doing this.

leebyron:
Currently all of the higher order functions like `map` and `filter` first create iterators, then populate new instances of the same class via that iterator. The new instance has no reference to the previous nor way of making the comparison that nothing has changed.

However, if you have an idea on how to do this, I would very happily take a look at a pull request.

jeanlauliac:
So apparently `withMutations` correctly returns the same object if nothing changes. So one idea could be to use `withMutations` to map all the values. I'll give it a try.

facebook-github-bot-6:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

PR #303, opened 2015-01-31
Add specialized Map#map to keep ref. equality

jeanlauliac:
NOTE: this changeset contains an unfortunate amount of unrelated changes to the built files. That means a previous commit was done without running `grunt`.

NOTE: the positioning of the function in the file is to be discussed, I'm not too sure if we want to create a new pragma section or not. The documentation will probably need a little update as well (https://facebook.github.io/immutable-js/docs/#/Map/map)

---

The point is to keep the referential equality when no value changed. This let
one call `Map#map` without afterthoughts about optimization. For example, let's
imagine that we trigger an interface render when a `Map` changed; we want to
avoid the cost of re-rendering when no value changed. Without referential
equality, we'd have a write some kind of custom algorithm to compare the values
returned by `mapper`.

The performance of this function is no worse, as the generalized
`IterableImpl#map` always reifies the mapped sequence anyway. On the other hand,
the fact we're applying modifications to the existing `Map` mean we're
benefiting from the internal structure optimizations.

As a side note, I find it interesting to see how a custom `map` function looks
significantly simpler than going through the generalized operation on `Seq`.
Granted, we lose the unification of sequence transformation algorithms, but
the original problem this changeset addresses proves it's hardly possible to
totally abstract a `Map` as a `Seq`: a `Seq` loses the concept of concreteness.

A similar fix can be used for `filter`, at least.

See #285.


leebyron:
Filter may be more difficult as it would require changing the structure as well. 

Could you add a portion to the benchmarks so we can test the before and after of this?

I'll look into why your build got so noisy.

facebook-github-bot-6:

PR #307, opened 2015-02-01
Update link to Immutable.js.ts

nick:
Previous link doesn't work


leebyron:
This file is autogenerated. The readme and generation files need to be changed. Thanks for the report though, I'll look into this.

leebyron:
Fixed in latest docs

PR #309, opened 2015-02-02
Fix typo

dmnd:
Manually applied fix to dist since building created a lot of changes.


leebyron:
An updated dist has been build, thanks for the reminder.

PR #314, opened 2015-02-04
Instantiate classes with 'new', now that ES6 requires it.

ivan:
immutable-js instantiates a lot of classes without `new`, which is no longer legal.  As per ES6 February 2, 2015 Draft Rev 32, "Constructors defined using class definition syntax throw when called as functions" https://www.mail-archive.com/es-discuss@mozilla.org/msg34104.html

6to5 implemented the new behavior in  https://github.com/6to5/6to5/commit/3a11c7d46b5788175c9c1141ff50ed59ce26c40c, and I couldn't run immutable-js/src/ with 6to5 until making these changes.

I extracted the list of classes with
```bash
grep --no-filename -o -P '^(export )?class .*? ' src/**/*.js
```

and generated the diff with this shell program:
```bash
for i in Collection KeyedCollection IndexedCollection SetCollection Iterable KeyedIterable IndexedIterable SetIterable Iterator List VNode Map ArrayMapNode BitmapIndexedNode HashArrayMapNode HashCollisionNode ValueNode MapIterator ToKeyedSequence ToIndexedSequence ToSetSequence FromEntriesSequence OrderedMap OrderedSet Range Record Repeat Seq KeyedSeq IndexedSeq SetSeq ArraySeq ObjectSeq IterableSeq IteratorSeq Set Stack; do echo; echo $i; perl -p -i -e "s,(?<\!new|ion|nds)([\( ])(\\b$i\\(),\\1new \\2,g" src/**/*.js __tests__/**/*.js contrib/**/*.js; done
```

I've signed the CLA as ivan@ludios.org

### DIFF
--- a/__tests__/RecordJS.js
+++ b/__tests__/RecordJS.js
@@ -6,7 +6,7 @@ var Record = Immutable.Record;
 describe('Record', () => {
 
   it('defines a constructor', () => {
-    var MyType = Record({a:1, b:2, c:3});
+    var MyType = new Record({a:1, b:2, c:3});
 
     var t = new MyType();
     var t2 = t.set('a', 10);
@@ -16,7 +16,7 @@ describe('Record', () => {
   });
 
   it('can have mutations apply', () => {
-    var MyType = Record({a:1, b:2, c:3});
+    var MyType = new Record({a:1, b:2, c:3});
 
     var t = new MyType();
 
@@ -50,14 +50,14 @@ describe('Record', () => {
   });
 
   it('can be cleared', () => {
-    var MyType = Record({a:1, b:2, c:3});
+    var MyType = new Record({a:1, b:2, c:3});
     var t = new MyType({c:'cats'});
 
     expect(t.c).toBe('cats');
     t = t.clear();
     expect(t.c).toBe(3);
 
-    var MyType2 = Record({d:4, e:5, f:6});
+    var MyType2 = new Record({d:4, e:5, f:6});
     var t2 = new MyType2({d:'dogs'});
 
     expect(t2.d).toBe('dogs');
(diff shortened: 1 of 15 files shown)


leebyron:
This has pretty severe performance implications. I'm actually planning on moving away from the class semanic because of this recent change to the spec. I don't actually use the constructor function behavior in Immutable (you can call these functions with `new`, but the constructed object is discarded). Instead these should be viewed as *constructor functions*, rather than *class constructors*.

ivan:
Ah, thanks, after reading the code I see what you mean.

leebyron:
But I definitely appreciate you investigating this.

pluma:
I think the word you were looking for is "factories". "Constructor function" still implies it modifies `this` ;)

ivan:
Tangential, but noting anyway: you still can't use `immutable-js/src/` directly with webpack+6to5 even with this PR, because of a circular dependency between `src/Seq.js` and `src/Iterable.js`.  Se

ivan:
Tangential, but noting anyway: you still can't use `immutable-js/src/` directly with webpack+6to5 even with this PR, because of a circular dependency between `src/Seq.js` and `src/Iterable.js`.  Seq requires Iterable which requires Seq; thus Iterable gets `undefined` for Seq, KeyedSeq, IndexedSeq, and SetSeq.

I noticed this problem in Safari and IE11 but not Firefox and Chrome, because Firefox and Chrome have native Array iterators and thus don't call the `new Iterable` in `forceIterator.js`.

faassen:
I just ran into this. I am trying to subclass classes generated by Immutable.Record to add my own methods and I'd like to use it in the immutable way. But babel (formerly 6to5) complains that I cannot instantiate my classes without "new".

If you are going to move away from the class semantic with Immutable.js, is subclassing still supported? Or is subclassing the wrong approach for this use case?

leebyron:
I am going to move away from classes in v4. They will officially be factory functions and the syntax will make that more clear. Also as part of the road to v4+, I'm doing some refactoring to make the code truly ES6 compliant, it's currently not so and dependent on a particular build pass.

In the meantime, I highly discourage using the src/ directory directly unless you're using the build tools that come in the repo. That codebase is not safe to use with Babel. Instead you should always use the dist/ directory.

leebyron:
Closing this PR since new isn't the correct semantic for factory functions.

PR #344, opened 2015-02-24
Add named exports (use es6 shorthand prop names)

zachasme:
I'm using the _babel/6to5_ transpiler and would like to be able to import each class separately instead of the Immutable namespace, like this:

``` javascript
import { List, Map, Record } from 'immutable'
```

This PR adds named exports in addition to the default export in `src/Immutable.js` (I also took the liberty of using of object shorthand notation for building the export objects).

Please note that #314 also needs to be solved before transpiling with babel is going to work again.


leebyron:
Hey zachasme,

Can you explain what happens when you use the current export today?

The raw source of Immutable is not designed to be used directly as it currently has some delicate circular dependencies. It's designed to be transpiled to a single file for use. The current top level export lets us produce the most tightly compiled output that should work with the most consumers. 

I have a few projects that use babel myself and I have used this import pattern without trouble, so I'm curious if something has changed that would break this more recently.

zachasme:
I use the [ES6 Module Loader polyfill](https://github.com/ModuleLoader/es6-module-loader) to load modules that are transpiled to `System.register` format using babel (`babel --modules system`).

Using the current export today, trying to `import { Map } from 'immutable'` results in Map being undefined. The module loader polyfill handles circular bindings, but default exports can only be imported as `import x from 'module'`.

Using this PR and making sure classes are instantiated with `new`, I have no problems, however (except for the performance implications cf. #314 ).

leebyron:
Can you run `grunt build` and update the PR? This might be okay, but we will need to take a look at what it does to the built distribution file.

> and making sure classes are instantiated with new

There are exceptionally few classes in Immutable.js, instead most are factory functions. Using `new` along side factory functions likely has some additional memory weight without any added value.

zachasme:
Esperanto will build a module exporting the Immutable namespace on `exports.default` and each class on `exports.Map`. I see how this might be a problem with interoperability.

As an alternative, I've pushed ba07e4c2394300838436186372c63a951e01ea08 which adds `interop.js` as the entry point for building the distribution, leaving `Immutable.js` to be imported if you use the source directly.

From babel/babel@3a11c7d46b5788175c9c1141ff50ed59ce26c40c forward it will be necessary to use `new` until the classes are replaced with regular factory functions.

leebyron:
This approach is more interesting. 

However, I'm curious why you want to load the raw source directly rather than the compiled production distributed file? Babel and other transpilers have support for loading in a vanilla CommonJS module like this one in the way you want.

zachasme:
I'm using the source solely to get the named exports. Under normal circumstances I would indeed use the compiled distribution file.

But with this PR I can import the source until interoperability with cjs modules becomes unnecessary.

facebook-github-bot-6:

lacker:
Hey, thanks for submitting a pull request. Sorry for not getting closure on this earlier but I think this pull request has sort of timed out, so I am just going to close it. Please feel free to reopen if you think this is still a problem though.

PR #345, opened 2015-02-25
Fix typo

drabinowitz:
Currently the example in the `README.md` shows `assert(map1 === map2)`. This is confusing for readers since these two are not equal, but we `assert` that they are. This change adjusts the example to `assert` the reality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ has changed.
 ```javascript
 var map1 = Immutable.Map({a:1, b:2, c:3});
 var map2 = map1.set('b', 2);
-assert(map1 === map2);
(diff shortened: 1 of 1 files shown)


tgriesser:
In this case the two are actually equal though, `map1.set('b', 2)` is a no-op because the value of `'b'` is already `2`, so it's just smart enough to return the same value.

drabinowitz:
Of course my mistake thank you. I wonder if an inline comment wouldn't help to clarify that fact.

leebyron:
Sorry, the hope was the assert() was illustrating that point, but I suppose the text could be more verbose.

PR #347, opened 2015-02-25
Use load-grunt-tasks for load dependencies

zzarcon:
Automatically include all grunt dependencies `grunt-*` using **load-grunt-tasks**  instead of call `loadNpmTasks` for each one :)


leebyron:
This is interesting, thanks for putting this together. However, I prefer the explicitness of loading each dependency. It's easier to keep track of what npm dependencies are in use.

Issue #346, opened 2015-02-25
some API methods not available on cursor


tcoopman:
I was surprised to see that some methods are not available on Cursor. For example if you have this:

```js
var data = Immutable.Map({'a': Immutable.List()});
var cursor = Cursor.from(data, ['a'], newData => {
  data = newData;
});

cursor.push(....); // error
```

It may be that I don't understand cursors completely, but I would expect cursor.push to work. Is there any reason, or is it just not implemented?

hufeng:
maybe you can try this.

```javascript
var Immutable = require('immutable');
var Cursor = require('immutable/contrib/cursor');

var data = Immutable.fromJS({a: Immutable.List()});
var cursor = Cursor.from(data,newData => {console.log(newData);data = newData;});

cursor.withMutations((cursor) => {
  cursor.update('a', (list) => list.push(1));
});
```

tcoopman:
This code looks like it should work without `withMutations` but it doesn't. Why is that?

Also this code cannot really work for me because the `withMutations` needs to know the structure of the root Data. I don't want that (that's a reason why I use the cursor).

wmertens:
How about making cursors automatically proxy methods from the object it points to instead of hardcoding?
In general, it would be interesting to call root object and referenced object methods, and it would be useful to provide the _keyPath as well.

So an option would be to provide a list of names to the cursor and when it creates a subcursor it adds proxy methods for the names it can find. There would be a list for the root object and names for the referenced object, perhaps with different signature types like no path, keyPath, keyPath+first arg.
That way, you could have a single polymorphic Cursor class which would behave mostly as the referenced object.

The update function could be completely customizable, for example.

wmertens:
In fact, since conceptually the Cursor is a API-transparent proxy for the object it references, perhaps a cursor should subclass the referenced object prototype completely, and only add/override whatever methods the root object wants exposed, always calling them with the _keyPath as the first argument.

The cursor would be whatever object or primitive it points to, plus `update`, `get`, `getIn`, ... All changes would have to go through `update`, like `cursor.update(cursor.push(...))`.

To support cursors for a path that doesn't exist, you provide a default object. Since it's immutable, it can be shared between all subcursors.

Example API:
```jsx
root = Map(...);
root.getRootSize = function() { return this.size; };
cursor = Cursor.from(root, {default: new Map(), proxy: ["getRootSize"], update:handleUpdate})
return (<Foo data={cursor}/>);
```
in Foo:
[...]

PR #364, opened 2015-03-04
added List methods push, pop, unshift, shift to List IndexedCursor

tcoopman:
I tried to add some methods to the Cursor API because they are missing. This solves #346.

I'm not sure if this is the right way to do it, because all other methods extend both KeyedCursorPrototype and IndexedCursorPrototype and this only extends IndexedCursorPrototype. So all feedback is more then welcome!


tcoopman:
It looks like something went wrong with the dist. Everything has `src_....` in front of it.
I have no idea what caused this. I've just ran `grunt`.

tcoopman:
closing in favor of new and corrected pull request

PR #365, opened 2015-03-04
added List methods push, pop, unshift, shift to List IndexedCursor

tcoopman:
I tried to add some methods to the Cursor API because they are missing. This solves #346.

I'm not sure if this is the right way to do it, because all other methods extend both KeyedCursorPrototype and IndexedCursorPrototype and this only extends IndexedCursorPrototype. So all feedback is more then welcome!

### DIFF
--- a/contrib/cursor/__tests__/Cursor.ts
+++ b/contrib/cursor/__tests__/Cursor.ts
@@ -179,6 +179,59 @@ describe('Cursor', () => {
     expect(cursor.map((x: number) => x * x)).toEqual(Immutable.Map({a: 1, b: 4, c: 9}));
   });
 
+  it('can push values on a List', () => {
+    var onChange = jest.genMockFunction();
+    var data = Immutable.fromJS({a: {b: [0, 1, 2]}});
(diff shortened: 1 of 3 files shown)


natew:
This is big for me, really hinders ability to use of cursors until this is fixed! :+1: :+1:

tcoopman:
@natew If there are other methods that are missing, let me know, I'm happy to implement them too.

shivawu:
wow, this is so great, exactly the things I need. Can't wait for it

leebyron:
Nice work, thanks for your patience!

natew:
I'm not sure how to push regular JS items onto an immutable cursor List, this doesn't work:

```js
immutableStore.push.apply(null, jsarray);
```

I've tried a few variations I keep getting:

```js
Possibly unhandled TypeError: Cannot read property 'updateIn' of undefined
    at updateCursor (webpack:///./~/immutable/contrib/cursor/index.js?:279:37)
```

tcoopman:
@natew can you give a full testcase? Somewhere as a gist or something?

PR #367, opened 2015-03-05
Typos in URLs.

briandipalma:


briandipalma:
Removing parentheses around URLs.

leebyron:
This file is auto-generated. Must be a bug in the doc generator.

leebyron:
Fixed in latest docs

PR #371, opened 2015-03-06
Add key as third argument to custom merger

tgriesser:
Useful if you want to have different behavior depending on the key being compared / merged.

Test and docs updated.


leebyron:
Seems like a pretty obvious win. Thanks for your patience, this is great.

Issue #353, opened 2015-02-27
Immutable.Stack#last() returns undefined


Zenwolf:
**Version:**
3.6.2

**Bug:**
Stack should return the last value when `last()` is called.

**Example:**
```javascript
var l = Immutable.List([1, 2, 3, 4]);
var s = Immutable.Stack(l);
s.first(); // 1
s.last(); // undefined
s.size; // 4
s.get(3); // 4

Zenwolf:
I checked the docs, and it mentions both `first()` and `last()` functions for Stack. However, there are no special notes that state that `last()` should behave differently for the Stack object. Can you clarify if it is intended that Stack should not return a `last()` value?

leebyron:
Thanks for finding this bug! I'll dig in to fix it

Issue #332, opened 2015-02-15
Stack.map throws a TypeError


astorije:
The documentation [shows](http://facebook.github.io/immutable-js/docs/#/Stack/map) that `Stack` has a `map` method. However, calling the method throws a `TypeError`.

Here is a minimalist example run on the command line through Node.js: 

```bash
$ node --version
v0.10.25
$ npm --version
2.3.0
$ npm list immutable
[...]
└── immutable@3.6.2 

$ node
> function double (x) { return 2 * x; }
> var Immutable = require('immutable');

> Immutable.List.of(1, 2, 3).map(double);
List [ 2, 4, 6 ]

> Immutable.Stack.of(1, 2, 3).map(double);
TypeError: Object function cacheResultThrough() {
    if (this._iter.cacheResult) {
      this._iter.cacheResult();
      this.size = this._iter.size;
      return this;
    } else {
      return Seq.prototype.cacheResult.call(this);
    }
  } has no method '__iterate'
    at Stack.__iterate ([...]/node_modules/immutable/dist/immutable.js:3421:41)
    at IndexedIterable.mappedSequence.__iterateUncached ([...]/node_modules/immutable/dist/immutable.js:1138:23)
    at seqIterate ([...]/node_modules/immutable/dist/immutable.js:601:16)
    at IndexedIterable.IndexedSeq.__iterate ([...]/node_modules/immutable/dist/immutable.js:315:14)
    at IndexedIterable.reversedSequence.__iterate ([...]/node_modules/immutable/dist/immutable.js:1183:23)
    at IndexedIterable.mixin.forEach ([...]/node_modules/immutable/dist/immutable.js:4265:19)
    at Stack.pushAll ([...]/node_modules/immutable/dist/immutable.js:3331:22)
    at Stack.unshiftAll ([...]/node_modules/immutable/dist/immutable.js:3357:19)
    at new Stack ([...]/node_modules/immutable/dist/immutable.js:3274:22)
    at reify ([...]/node_modules/immutable/dist/immutable.js:1673:37)
```

I haven't tried other methods of `Stack` or the `map` method on other data types.

leebyron:
Thank you for this report! I'll look into it

tchak:
I am experiencing this problem also

PR #373, opened 2015-03-07
Fix Stack bugs #332 and #353

ellbee:
Immutable.Stack#last() returns undefined #353:

Stack.last() calls Stack.get(-1), but Stack.get() did not contain the wrapIndex call to handle negative indexes.

Stack.map throws a TypeError #332:

Stack.map was throwing an error due to the following line of the Stack.__iterate method in Stack.js:

return this.toSeq().cacheResult.__iterate(fn, reverse);

Note the missing parens after cacheResult. However, reinstating them led to an infinite loop where Stack.__iterate was calling Stack.__iterate recursively which is why I changed this line to use this.reverse().__iterate(fn). I also made a similar change to the Stack.__iterator method.

Tests have been added. 

The transpiler seems to have made some extra changes to dist/immutable.js, renaming some of the classes (Map and Iterator). I don't know if this is a problem. Maybe I missed something in the build step?


leebyron:
Thanks for this work! I appreciate you digging in and figuring out what to do

Issue #242, opened 2014-12-12
Expose rootData and/or keyPath of a Cursor


Dashed:
I usually find it useful to know the origin path for a cursor via `this._keyPath`. I usually don't want to keep track of both of them external to the cursor. 

I know they're meant to be private: https://github.com/facebook/immutable-js/blob/52db807330bb7786fbe20156aca80fbae47491c5/contrib/cursor/index.js#L43-L44

Can `rootData` and/or `keyPath` be officially public?

jeffbski:
+1

I think that this is very important to provide access to the keyPath, it enables a variety of great uses.

One use case is so that you can serialize the keyPath and updateValue in one piece of code, transmit it to another (possibly on the server), then apply the changes there. Full stack flux could use this approach.

```javascript
var structure = Immutable.fromJS({ a: { b: { c: 'hello' } } });
// somewhere deep in your code, you have a cursor that was passed down to you, maybe through
// subcursors, etc.
// var cursor = ...;  // _keyPath = ['a', 'b', 'c']
// I want to communicate to the server via web socket or ajax that data at cursor keyPath needs to be updated
socket.send({ update: { keyPath: cursor.keyPath, data: 'world' }});

// serialized keyPath and sent to server

// server validates and applies update to db or its own immutable structures and sends update to client
// validate and update server db at keyPath with data
// after all stores have updated, send notification with changed data to client

// client receives the update and applies to local structure
socket.on('update', (obj) => {
  var cursor = Cursor.from(structure, obj.keyPath);
  cursor.update(x => obj.data)
});
```

In the above use case, we did not want to update our local structure directly since we wanted it to be done in a central location after validation, then only if validation succeeds, server would indicate to client to update local cache and views.

dariocravero:
If I didn't get this wrong, another use case could be in traversing a tree-like structure. So through the cursor you could access the parent, siblings, etc. I was actually wondering as to why cursors weren't exposing that kind of functionality and until I found this issue I thought maybe I was trying to misuse them?

wmertens:
I'm working on an app where I need to edit parts of a document and using a cursor for the document part makes for a nicely modular approach.

So two use cases while traversing a tree:
- You may want to render a part of the tree differently depending on the path taken
- You may want to query location-specific calculated document metadata

In the first case you either have to pass the path along with the cursor or move the rendering decisions to the document viewer and have it pass flags to the editor. Neither option seems nice.

[...]

PR #379, opened 2015-03-09
Support tree-like data structures through Cursors

dariocravero:
Hi all,

Following up on #242, [we're using Cursors to browse tree-like data structures](https://github.com/facebook/immutable-js/issues/242#issuecomment-77404067).

This PR implements `parent`, `sibling`, `nextSibling` and `prevSibling` on `Cursor` to work with tree-like data structures.

Would that be a desirable feature for `Cursor` or should it live in a more specific implementation?

Thanks :)
Darío

### DIFF
--- a/contrib/cursor/__tests__/Cursor.ts
+++ b/contrib/cursor/__tests__/Cursor.ts
@@ -294,4 +294,71 @@ describe('Cursor', () => {
     );
   });
 
+  it ('can get its parent', () => {
+    var data = Immutable.fromJS({a:{b:{c:1}}});
+    var c1 = Cursor.from(data, ['a']);
+    var c2 = Cursor.from(data, ['a', 'b']);
+
+    expect(Immutable.is(c1.parent(), data)).toBeTruthy();
+    expect(Immutable.is(c2.parent(), c1)).toBeTruthy();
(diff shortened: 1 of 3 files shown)


leebyron:
This should live in a separate sort of thing. Cursor is meant to be a very transparent view over a Map or a List, not to represent a new kind of data structure. What you're describing sounds more like Zippers which we do not have any tools for at the moment. Feel free to start to flesh that out!

dariocravero:
Good stuff! Thanks for the heads up :) So, would that data structure be
accepted in, say, contrib? Or should I do my own package for it?
Is it ok to base it off Cursors and have it use its internals? I guess that
my biggest concern with it is that internals like _rootData and _keyPath
can change. Perhaps these could be public? Or am I missing something? Thanks

leebyron:
Sorry for missing your comment here. I highly suggest starting a project for this outside this library. In a future version, the `contrib` folder will be replaced by separate github projects.

What you're onto here is similar to zippers and lenses and probably benefit from being able to evolve as such rather than being tied to the fate of a (not the best) implementation of data cursors.

dariocravero:
That's fine :) Thanks for your answer on this!

PR #380, opened 2015-03-09
Minor Readme Update - putting Phil's name first. 

hughfdjackson:
Phil's (may he rest in peace) contribution is massive by comparison. His name ought to come first (at least.. I think so)


leebyron:
Good call, Hugh!

PR #381, opened 2015-03-09
subCursor (cursor.cursor('a')) returns improper type

jeffbski:
Due to defect in the code, calling cursor() on cursor can return improper type

```javascript
var data = Immutable.fromJS({ a: [1, 2, 3] });
var cursor = Cursor.from(data);
var deepCursor = cursor.cursor('a');
deepCursor instanceof IndexedCursor; // fails since is KeyedCursor
```

Originally reported here https://github.com/facebook/immutable-js/pull/297 I decided that it might be a safer fix to continue using argument.length rather than checking undefined since I think it might be valid to use undefined in some circumstances.

So this fix goes back to using argument.length in the makeCursor check, but properly calls it with only 3 arguments from subCursor if subCursor is not called with value.

### DIFF
--- a/contrib/cursor/__tests__/Cursor.ts
+++ b/contrib/cursor/__tests__/Cursor.ts
@@ -49,6 +49,13 @@ describe('Cursor', () => {
     expect(deepCursor.deref()).toBe(data.getIn(Immutable.fromJS(['a', 'b'])));
   });
 
+  it('cursor return new cursors of correct type', () => {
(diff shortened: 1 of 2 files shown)


jeffbski:
This issue is also mentioned in #318

wmertens:
:+1:

leebyron:
Thanks for your patience, this is a great fix.

jeffbski:
Great! Thanks.

Issue #300, opened 2015-01-29
Have Cursor.isCursor(maybeCursor)?


Dashed:
Pretty much all collections have a static method to check if an object is that particular type. AFAIK, there's no official way to do this `Cursor.isCursor(maybeCursor)`.

This is needed for a `shouldComponentUpdate` in React whose props/state may hold cursors.

At the moment, this is used to identify cursors:

```js
function isCursor (maybeCursor) {
  return maybeCursor && typeof maybeCursor.deref === 'function';
}
```

Is there a better way to do this externally? Or if it can be baked in as a static function? :smile:

jeffbski:
+1

leebyron:
This is a bit counter to the purpose of Cursors, which seek to be invisible and act as the types they wrap.

For shouldComponentUpdate, I recommend using `Immutable.is()` which will automatically handle Cursor unwrapping for comparisons.

If you want to deref regardless of Cursor for more specific comparisons, you can use `obj.valueOf()`. Regular types return themselves in `valueOf`, but Cursors return their `deref()`.

Gozala:
> This is a bit counter to the purpose of Cursors, which seek to be invisible and act as the types they wrap.

> For shouldComponentUpdate, I recommend using Immutable.is() which will automatically handle Cursor unwrapping for comparisons.

Given the discussion here https://github.com/omniscientjs/omniscient/issues/76 I agree with @leebyron that it would make more sense to defer comparison to immutable type implementation or cursor implementation than try to handle each individual type in a custom way.

Gozala:
I also have to say that I was proposing to make a fork, but mainly to resolve issue if rendering on animation frame as om does.

Current problem with cursors can be illustrated as:

```js
const not = x => !x;


ToggleButton({ onClick: event => cursor.update(‘isChecked’, not)})
```

If onClick is invoked even number of times with in the same animation frame result will be unexpected, the problem is render won't run there for event handler will refer to cursor that has out of date knowledge of the state.

I do believe this is a reason why om has two ways to defer cursor https://github.com/omcljs/om/wiki/Cursors#consistency-and-observability `(deref cursor)` to get current state one `(om/value cursor)` to get rendered state.

I think adding such a capability to cursors would likely be necessary to handle async rendering.

Gozala:
Deleted my response as I realised I posted it to a wrong thread.

wmertens:
[...]

PR #386, opened 2015-03-12
is(): test primitives when dereferencing

wmertens:
Repeating the tests instead of having to recurse and handling a recursion guard

See https://github.com/facebook/immutable-js/issues/300#issuecomment-78452758 for a problem description


leebyron:
We will still need to cover the case where valueOf returns a null-ish primitive which throws when calling value.equals. I think that's the key issue here.

leebyron:
Also, could you add test cases that this solves for and ensure you run "grunt" as part of the pull request?

wmertens:
@leebyron done. The tests I added pass on this PR and throw on current master

Note that the tests were actually not testing, see the expectIs* helpers :grinning:. I had to fix the 0, -0 equality

wmertens:
@leebyron anything else needed to get this merged?

__tests__/Equality.ts
```diff
@@ -12,14 +12,18 @@ describe('Equality', () => {
 
   function expectIs(left, right) {
     var comparison = Immutable.is(left, right);
+    expect(comparison).toBe(true);
     var commutative = Immutable.is(right, left);
-    return comparison && commutative && comparison === commutative;
+    expect(commutative).toBe(true);
+    expect(comparison === commutative).toBe(true);
```
leebyron: this line isnt necessary since you already made the expectations above that both values are `true`, which means there is no possible scenario that this expectation could fail.

__tests__/Equality.ts
```diff
@@ -36,7 +40,9 @@ describe('Equality', () => {
     expectIs(NaN, NaN);
     expectIs(0, 0);
     expectIs(-0, -0);
-    expectIsNot(0, -0);
+    // Note: Unlike Object.is, Immutable.is assumes 0 and -0 are the same value,
+    // matching the behavior of ES6 Map key equality.
+    expectIs(0, -0);
```
leebyron: Thank you for fixing this!

leebyron:
Thank you again for your patience and great work. I'm happy this issue is fixed and very happy you found and fixed the issue with the tests.

(btw, I fixed that inline concern as I merged)

Issue #338, opened 2015-02-19
OrderedMap can't be created from List of Lists


ivan-kleshnin:
```javascript
let {OrderedMap, List} = require("immutable");

console.log(OrderedMap(List([
  ["A", "B"]
]))); // OrderedMap { A: "B" } -- ok

console.log(OrderedMap([
  List(["A", "B"])
])); // OrderedMap { undefined: undefined } -- should be the same OrderedMap { A: "B" }

console.log(OrderedMap(List([
  List(["A", "B"])
]))); // OrderedMap { undefined: undefined } -- should be the same OrderedMap { A: "B" }
```

P.S.
Same problem with `Map`

PR #387, opened 2015-03-12
Create Map from List of Lists #338

ellbee:
Addresses issue #338. Adds the ability to create a Map and OrderedMap from a List of Lists (if this is something that should actually be supported). Ran the perf tests and all seems to be good.


leebyron:
There's no perf test covering this case yet. We will first need good perf coverage of this use case. I imagine creating a new Seq on every iteration is probably quite expensive.

ellbee:
Re the perf tests, I was just trying to make sure that I hadn't caused a performance regression for existing cases as I had the same worries about creating a Seq each time vs straight up array access. I'll get perf coverage added, I guess it should follow the 2, 8, 32, 1024 pattern?

ellbee:
Ah sorry @leebyron, I misread the Map perf test and as you hinted it turns out I wasn't doing a regression test at all! I have added construction with arrays and Lists to the test and not surprisingly Seq did indeed turn out to be expensive. Here's the perf test (only the 1024 case shown, Old is current branch of master):

```
Map builds from an array of 1024
  Old:     1,678     1,692     1,706 ops/sec
  New:       833       841       849 ops/sec
  compare: 1 1
  diff: -50.27%
  rme: 0.9%

Map builds from a List of 1024
  New:       260       262       265 ops/sec
```
Whoops. So I think this is the most correct, if not fastest, thing to do:

```javascript
  __iterate(fn, reverse) {
      return this._iter.__iterate(function(entry ) {
        // Check if entry exists first so array access doesn't throw for holes
        // in the parent iteration.
        if (entry) {
          validateEntry(entry);
          var iterable = !Array.isArray(entry) && isIterable(entry);
          var k = iterable ? entry.get(0) : entry[0];
          var v = iterable ? entry.get(1) : entry[1];
          return fn(v, k, this);
        }
      }, reverse);
  }
```

The isArray check prevents a larger drop off when using plain arrays.

```
Map builds from an array of 1024
  Old:     1,639     1,656     1,674 ops/sec
  New:     1,570     1,596     1,622 ops/sec
  compare: 1 1
  diff: -3.67%
  rme: 1.35%

Map builds from a List of 1024
  New:     1,135     1,146     1,158 ops/sec
```

This might also be an option, although I suspect it has edge cases e.g. get function doesn't have the correct semantics. It is significantly faster for constructing from List:

```javascript
  __iterate(fn, reverse) {
      return this._iter.__iterate(function(entry ) {
        // Check if entry exists first so array access doesn't throw for holes
        // in the parent iteration.
        if (entry) {
          validateEntry(entry);
          var iterable = typeof entry.get === 'function';
          var k = iterable ? entry.get(0) : entry[0];
          var v = iterable ? entry.get(1) : entry[1];
          return fn(v, k, this);
        }
      }, reverse);
  }
```

```
Map builds from an array of 1024
  Old:     1,635     1,653     1,671 ops/sec
  New:     1,652     1,667     1,682 ops/sec
  compare: 1 1
  diff: 0.86%
  rme: 0.98%

Map builds from a List of 1024
  New:     1,521     1,534     1,547 ops/sec
```

leebyron:
Excellent work, thank you for being so detail-focused!

PR #391, opened 2015-03-15
toString(): Print keys in JSON

wmertens:
This way, it's obvious when numeric keys are being mixed with string
keys, like when importing from JSON and inserting by key.


leebyron:
Thanks, this makes tons of sense. I wholey agree that being explicit about key types when stringed helps explain the difference from objects, and naked strings look far too much like property names, which they are not.

PR #411, opened 2015-03-25
Implement sub-record construction / update preserving record types

Gozala:
This is a fix for #408

Issue #407, opened 2015-03-25
Collection records


Gozala:
I've being using Records for modelling application state and soon enough run into a case where I'd like to specify a record field that must be a collection of other Records. Something along these lines:

``` js
const Suggestion = Record({ url: '', title: '' })

const Completer = Record({
  selected: -1
  entries: List()
})
```

Here I'm unable to specify that `Completer` entries should be a `Suggestion` records. I would love if there was a way to declare that somewhat along these lines:

``` js
const Suggestion = Record({ url: '', title: '' })

const Completer = Record({
  selected: -1
  entries: Suggestion.List()
})
```

leebyron:
I think you may be mis-interpretting the initial value for the Record class. It is intentional that you're unable to specify what type each field is because type checking will quickly expand into a very large (and possibly expensive) extension to this codebase in a really unrelated matter.

What the object provided to Record describes is the list of valid keys on that Record and the *default values* for each key, not their type. For example: to articulate the difference between `entries`, a list of `Suggestion` records, having a default value of empty list `List()` or `null` (e.g. should `entries` be nullable?) you could write either:

```js
const Completer = Record({
  selected: -1 // Number
  entries: List() // List<Suggestion>
})
```

or

```js
const Completer = Record({
  selected: -1 // Number
  entries: null // ?List<Suggestion>
})
```

Does this make sense?

leebyron:
I'm very interested in layering type information atop Record, I just don't think this library is the right place for this to live. I see a few ways forward:

1. Integration with static type checking tools like `Flow` and `Typescript`. This would allow the additional type information to be removed at runtime, offering zero overhead and allow for full program analysis ahead of time.

2. A separate library which supplies runtime type checking for Record (and presumably the rest of Immutable.js). This will come at a serious runtime cost, but will enable the strongest sense of safety. That cost (both runtime cost and library size) is not one that every user of Immutable.js should pay, which is why I think this needs to be it's own library if desired.

leebyron:
I now understand from your pull-requests that you're interested not just in type validation but parsing an arbitrary JS body into a tree of Immutable collections. That's also super interesting and something I would love to see, but probably should not be specific to Record, but apply to all Immutable collections.

Gozala:
> I now understand from your pull-requests that you're interested not just in type validation but parsing an arbitrary JS body into a tree of Immutable collections. That's also super interesting and something I would love to see, but probably should not be specific to Record, but apply to all Immutable collections.

[...]

Issue #412, opened 2015-03-25
Can't run tests on node 0.12.0


Gozala:
All `*.ts` tests fail on node 0.12.0 (at least on my machine) with a pretty similar stack traces:

```sh
> immutable@3.6.4 test /Users/gozala/Projects/immutable-js
> ./resources/node_test.sh

Using Jest CLI v0.4.0
 FAIL  __tests__/ArraySeq.ts
TypeError: /Users/gozala/Projects/immutable-js/resources/jestPreprocessor.js: undefined is not a function
  at BitMatrixImpl.release (/Users/gozala/Projects/immutable-js/node_modules/ts-compiler/node_modules/typescript/bin/tsc-unlocked.js:815:32)
  at PullInstantiatedTypeReferenceSymbol.PullTypeSymbol.wrapsSomeTypeParameterIntoInfinitelyExpandingTypeReference (/Users/gozala/Projects/immutable-js/node_modules/ts-compiler/node_modules/typescript/bin/tsc-unlocked.js:31567:26)
  at PullInstantiatedTypeReferenceSymbol.getGenerativeTypeClassification (/Users/gozala/Projects/immutable-js/node_modules/ts-compiler/node_modules/typescript/bin/tsc-unlocked.js:46402:30)
  at PullTypeEnclosingTypeWalker.getGenerativeClassification (/Users/gozala/Projects/immutable-js/node_modules/ts-compiler/node_modules/typescript/bin/tsc-unlocked.js:32236:36)
  at PullTypeResolutionContext.oneOfClassificationsIsInfinitelyExpanding (/Users/gozala/Projects/immutable-js/node_modules/ts-compiler/node_modules/typescript/bin/tsc-unlocked.js:32867:71)
  at PullTypeResolver.typesAreIdenticalInEnclosingTypes (/Users/gozala/Projects/immutable-js/node_modules/ts-compiler/node_modules/typescript/bin/tsc-unlocked.js:39263:29)
  at PullTypeResolver.signatureTypeParametersParametersAndReturnTypesAreIdentical (/Users/gozala/Projects/immutable-js/node_modules/ts-compiler/node_modules/typescript/bin/tsc-unlocked.js:39504:52)
  at PullTypeResolver.signaturesAreIdenticalWorker (/Users/gozala/Projects/immutable-js/node_modules/ts-compiler/node_modules/typescript/bin/tsc-unlocked.js:39492:75)
[...]

PR #413, opened 2015-03-25
Implementation of Record Lists

Gozala:
This is a fix for #407. Please note that this change is based on #411 so it will appear to contain somewhat unrelated changes actual changes related to only this feature are:
https://github.com/Gozala/immutable-js/compare/sub-records...Gozala:list-record


Gozala:
For now I just did some manual testing as I'm having issues with running tests locally (see #412 for details) and I need to crash. Will submit actual test for this feature tomorrow.

leebyron:
Considering that Record.List's constructor returns a List. I'm not convinced this warrants being a new type of collection. Instead what it looks like you want to do is build a smarter kind of `fromJS` which is aware of a schema of Records, Lists (and Maps, Sets).

I would recommend re-framing this as a standalone function which takes two arguments: the "shape" of the data to construct and the input js object data that returns a tree of Immutable collections in the appropriate shape.

Gozala:
> Considering that Record.List's constructor returns a List. I'm not convinced this warrants being a new type of collection. Instead what it looks like you want to do is build a smarter kind of fromJS which is aware of a schema of Records, Lists (and Maps, Sets).

That is not entirely correct, `RecordType.List()` returns a subclass of the `List`, which is indeed a List but it's `.push()` `.pop()` etc have different behavior in that they do add instances of `RecordType`. This does enables you to define polymorphic operations over lists that don't need to be `RecordType` specific that this list consists of. In other words parsing JS at the construction is a part of it indeed, but there is more to this than just that.

I do agree though that other types (but Maps) should probably have it as well.

facebook-github-bot-4:

lacker:
Hey, I think this pull request has essentially timed out. Sorry! I am going to close it but if you are still interested in it enough to rebase and poke around to see if it still makes sense then please feel free to reopen.

PR #415, opened 2015-03-25
Add function for intersecting an unknown number of sets

matthewwithanm:
This commit adds a `Set.intersection()` function for getting the intersection of an unknown number of sets. This isn't a problem with `union`, since the identity element for that is an empty set (which is easy to create):

``` javascript
Immutable.Set().union(...allTheThings);
```

However, with intersection, it's the universal set and Immutable doesn't have a representation of that.


leebyron:
If there is truly an unknown number of sets, would it be better to provide an array/list of Sets instead of using var-args?

leebyron:
I'm curious when you would use this API over set1.intersection(set2)

matthewwithanm:
> If there is truly an unknown number of sets, would it be better to provide an array/list of Sets instead of using var-args?

Yeah, maybe that would be better…I wasn't sure if it would be confusing to deviate from `Set#intersect()`'s signature.

> I'm curious when you would use this API over set1.intersection(set2)

Sorry for the bad explanation. The purpose of this API is to handle cases where you have an unknown number of sets *that's potentially zero.*

facebook-github-bot-4:

kozlitinaelja:
Hi there. I am starting handling backlog. I know it is been a while since you've submitted PR, sorry for such delay. The are some conflicts. Could you rebase or close the issue is obsolete? 😅

matthewwithanm:
I guess it's still relevant? Honestly, I haven't been using Immutable lately so I'm not sure!

PR #416, opened 2015-03-26
Speed up TypeScript compilation when running test suite

### LINKED ISSUE #399: Tests are slow
The test suite runs rather slow, and I was wondering whether it is possible to speed it up.

I’ve tested different variants on MBPs from 2010 and 2008 (sorry, don’t have more recent hardware available atm):

CPU           | current state | [no typescript][] | [jasmine instead of jest, no typescript][]
---------------|-------------------|----------------------|--------------------------------------------
Core2Duo |     4m20.332s |       0m51.636s |     0m27.554s
i7 (2010)   |       2m1.904s |       0m24.786s |     0m23.579s

[no typescript]: https://github.com/davidaurelio/immutable-js/tree/no-typescript-in-tests
[jasmine instead of jest, no typescript]: https://github.com/davidaurelio/immutable-js/tree/no-jest-no-typescript

- The biggest offender is TypeScript, so the first question is: Do the tests really need to use typescript?

- The QuickCheck-style tests (with jasmine-check) take their time. I think it might be worth considering moving those to a separate suite. That way, there’d be a really fast running test suite that gives quick feedback (< 5 seconds) during development, and the jasmine-check-suite that can be run on occasion and by CI. 

- I don’t know why jest is slow on old hardware, but I guess that matters only to me when working on my ancient machine.

robertknight:
Speed up the test suite by updating to the current version
of the TypeScript compiler which is ~5x faster than the 1.0
version that was being used previously and caching compiled
TypeScript files.

 * Update type definitions and flatten.ts for compatibility
   with TS 1.4
 * Remove ts-compiler dependency and use the now-public TypeScript
   compiler API directly.

Fixes #399

With these changes a full run of the test suite on my laptop goes from ~100s to ~20s on first run and ~10s for subsequent runs where the test files have not changed.

*There is currently a test failure in slice.ts which I'm looking into*

### DIFF
--- a/__tests__/flatten.ts
+++ b/__tests__/flatten.ts
@@ -8,6 +8,8 @@ import I = require('immutable');
 import jasmineCheck = require('jasmine-check');
 jasmineCheck.install();
 
+type SeqType = number | number[] | I.Iterable<number,number>;
+
 describe('flatten', () => {
 
   it('flattens sequences one level deep', () => {
@@ -29,13 +31,13 @@ describe('flatten', () => {
   })
 
   it('flattens only Sequences (not sequenceables)', () => {
-    var nested = I.Seq.of(I.Range(1,3),[3,4],I.List.of(5,6,7),8);
+    var nested = I.Seq.of<SeqType>(I.Range(1,3),[3,4],I.List.of(5,6,7),8);
     var flat = nested.flatten();
     expect(flat.toJS()).toEqual([1,2,[3,4],5,6,7,8]);
   })
 
   it('can be reversed', () => {
-    var nested = I.Seq.of(I.Range(1,3),[3,4],I.List.of(5,6,7),8);
+    var nested = I.Seq.of<SeqType>(I.Range(1,3),[3,4],I.List.of(5,6,7),8);
     var flat = nested.flatten();
     var reversed = flat.reverse();
     expect(reversed.toJS()).toEqual([8,7,6,5,[3,4],2,1]);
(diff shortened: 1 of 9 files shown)


leebyron:
There are some failing tests because of this change. What's the reason for that?

__tests__/slice.ts
```diff
@@ -31,7 +31,7 @@ describe('slice', () => {
   it('slices a sparse indexed sequence', () => {
     expect(Seq([1,,2,,3,,4,,5,,6]).slice(1).toArray()).toEqual([,2,,3,,4,,5,,6]);
     expect(Seq([1,,2,,3,,4,,5,,6]).slice(2).toArray()).toEqual([2,,3,,4,,5,,6]);
-    expect(Seq([1,,2,,3,,4,,5,,6]).slice(3, -3).toArray()).toEqual([,3,,4,,,]); // one trailing hole.
+    expect(Seq([1,,2,,3,,4,,5,,6]).slice(3, -3).toArray()).toEqual([,3,,4,,]); // one trailing hole.
```
leebyron: I guess TypeScript changed their rules around trailing comma removal in 1.4?

__tests__/flatten.ts
```diff
@@ -29,13 +31,13 @@ describe('flatten', () => {
   })
 
   it('flattens only Sequences (not sequenceables)', () => {
-    var nested = I.Seq.of(I.Range(1,3),[3,4],I.List.of(5,6,7),8);
+    var nested = I.Seq.of<SeqType>(I.Range(1,3),[3,4],I.List.of(5,6,7),8);
```
leebyron: I would have accepted `any`, but I applaud your specificity.

leebyron:
Nice work!

robertknight:
> I guess TypeScript changed their rules around trailing comma removal in 1.4?

Looks to me like it was simply a bug that got fixed.

Thanks for the quick review!

andrewluetgers:
I'm having trouble getting started with contribution to the project. My first time trying to run tests failed. I tried with the tag prior to this change and all was fine. I just filed an issue for it https://github.com/facebook/immutable-js/issues/431

Issue #394, opened 2015-03-16
Filtered sequence + take + react 0.13 hangs


kentor:
Wasn't sure if I should file this here or at the react project.

Very bizarre, this code crashes in latest Chrome and Firefox for me:

```javascript
var data = Immutable.List([1])
  .toSeq()
  .filter(_ => true)
  .take(1)

React.createElement('div', null, data);
```

Using react 0.13 and immutable-js 3.6.4

JSBin: http://jsbin.com/qoyimuneqo/2/edit?html,js,output (might crash tab, firefox handles it better for me)

Taking out the line with `filter` or the `take` will work as usual.

edit: removed 6to5/jsx dependency

kentor:
From my debugging, it gets stuck in the while loop.

```javascript
  } else if (node) {
    var iteratorFn = getIteratorFn(node);
    // Entry iterators provide implicit keys.
    if (iteratorFn) {
      if (iteratorFn !== node.entries) {
        var iterator = iteratorFn.call(node);
        var step;
        while (!(step = iterator.next()).done) {
          if (ReactElement.isValidElement(step.value)) {
            validateExplicitKey(step.value, parentType);
          }
        }
      }
    } else if (typeof node === 'object') {
```

leebyron:
Thanks for the report! I'll dig into this

PR #417, opened 2015-03-26
fix slice iterator bug #394

ellbee:
Fixes issue #394. An iterator created from a sliceSeq would get into an infinite loop on the second call to next() because it would run the code that skips to the first item each time next() was called. Test added.


leebyron:
This is almost the fix we want - but we also want the creation of the iterator to be effect-free. The first call to `next()` should cause the underlying seq to fast-forward to the starting position.

I believe the fix should probably be replacing `!==` with `>` so the second iteration skips over this condition.

Thanks for expanding the test case and finding and fixing this issue!

leebyron:
I added my requested change as well as a few more tests to your PR before merging.

Thanks again!

Issue #419, opened 2015-03-27
set.intersect() fails inside withMutations if sharing elements with set.add/intersect


yetanotherhacker:
Running this in node

```
Immutable.Set(['a', 'd']).withMutations(function(set) {
      set.add('b');
      set.union(['c']);
      set.intersect(['b', 'c', 'd']);
})
```

will fail with a `TypeError: Cannot read property '1' of undefined`. It seems that `intersect()` does not play well when `add()` or `union()` calls before it that share similar elements.

yetanotherhacker:
I thought I'd check up on this - could I get feedback / acknowledgement that this has been seen?

leebyron:
Seen. Certainly there shouldn't be a TypeError, I'll look into that. But this is also not legal code. Union and Intersection always return new collections - even within a withMutations block, so this should result in the Set ['a', 'b', 'd'].

yetanotherhacker:
I see. Thanks!

magnus-madsen:
A possible suggestion, inside src_Set__Set.prototype.intersect, should the inner call back return false to stop the iteration?

    src_Set__Set.prototype.intersect = function() {var iters = SLICE$0.call(arguments, 0);
        if (iters.length === 0) {
            return this;
        }
        iters = iters.map(function(iter ) {return SetIterable(iter)});
        var originalSet = this;
        return this.withMutations(function(set ) {
            originalSet.forEach(function(value ) {
                if (!iters.every(function(iter ) {return iter.includes(value)})) {
                    set.remove(value);
                    // <--- return false here?
                }
            });
        });
    };

magnus-madsen:
Suggestion, inside src_Set__Set.prototype.intersect, the values to be removed inside the callback could be locally accumulated and then removed. For example:

src_Set__Set.prototype.intersect = function() {var iters = SLICE$0.call(arguments, 0);
        if (iters.length === 0) {
            return this;
        }
        iters = iters.map(function(iter ) {return SetIterable(iter)});
        var originalSet = this;
        return this.withMutations(function(set ) {

            var acc = [];

            originalSet.forEach(function(value ) {
                if (!iters.every(function(iter ) {return iter.includes(value)})) {
                    acc.push(value)
                }
            });

            acc.forEach(function (value) {
                set.remove(value);
            })
        });
    };

facebook-github-bot-4:
[...]

PR #420, opened 2015-03-27
get set intersect working after add or union in withMutations

yetanotherhacker:
This addresses issue #419.


leebyron:
Hmm, this seems like a mitigation rather than a fix. A map shouldn't have nulled entries like this. Let me dig deeper, but thank you for your initial exploration.

facebook-github-bot-4:

kozlitinaelja:
@leebyron, is there any decision was made regarding this PR?

leebyron:
I'd like to get to the bottom of this. The original fix is patching a symptom and probably isn't the only place this could appear.

Issue #310, opened 2015-02-02
Adding Iterable#getPath('foo.bar') to the library?


jviereck:
In my current project I have to do a lot of deep lookups. At the moment immutable-js aids this usecase by using the `Iterable#getIn(...)` method like this:

```js
var nested2 = nested.mergeDeep({a:{b:{d:6}}});
nested2.getIn(['a', 'b', 'd']); // 6
```

Instead of passing an array to the `getIn` method, I personally find passing a string more convenient, like:

```js
nested2.getPath('a.b.d'); // 6
```

where the implementation of `Iterable#getPath(...)` might look like this:

```js
mixin(Iterable, {
    ...
    getPath: function(searchKeyPathStr, notSetValue) {
      return this.getIn(searchKeyPathStr.split('.'), notSetValue);
    },
    ...
}
```

How do you feel about adding a function like `getPath` to the library? I am happy to create a PR in case there is enough support for this idea.

PS: The idea for `getPath(...)` originates from my experience with the SproutCore library and their SC.Observable object: http://docs.sproutcore.com/symbols/SC.Observable.html

Dashed:
Unlikely this would be in the library. This has been suggested before.

See: https://github.com/facebook/immutable-js/issues/237

leebyron:
Thanks @Dashed. Yes, the tricky thing here is that not everyone agrees on what "path" encoding should be used. Also note that encoding path as a string means that you cannot use it to access when a non-string key is used. Finally, what would `'a.1.b'` become? `"1"` and `1` are different keys! You either exclude numeric strings, or you exclude the ability to index into a List!

Despite the issues with this, I know that it's a valuable shorthand. My suggestion to those who want this shorthand is to use a locally scoped function that turns it into a key path, so they can use the semantics they care about. Since the function is so small, it's probably fine for it to live inline in your application instead of needing to live on npm.

```js
function path(pathString) {
  return pathString.split('.').map(key => parseInt(key) || key);
}
```

Then you can use it for any `*In` function in the library:

```js
nested.getIn(path('a.b.d.1.2.3'));
```

jviereck:
Thanks for the detailed replay @leebyron and @Dashed - make totally sense to me.

Issue #237, opened 2014-12-10
Support "JSON pointer" in Cursor.from and cursor.get


aaronshaf:
Instead of (or in addition to):

``` javascript
var cursor = Cursor.from(data, ['a', 'b'], newData => {
  data = newData;
});
```

Support:

``` javascript
var cursor = Cursor.from(data, '/a/b', newData => {
  data = newData;
});
```

See http://tools.ietf.org/html/rfc6901 and https://github.com/manuelstofer/json-pointer as an example.

gajus:
What about using https://www.npmjs.com/package/lodash.topath (https://lodash.com/docs#toPath)?

It is a tiny dependency and could be used to convert string value to an array path. This convention is already broadly used within lodash, e.g. https://lodash.com/docs#get

leebyron:
This isn't a question of which dependency to take, but rather that this library seeks to remain dependency-less. 

Also, performance is a primary concern, and in past performance tests we found this parsing step to be a serious perf regression on getting values. 

If it's important to you to use a path notation, then more power to you, it's easy to compose functions together to do just that, but I don't think it's the right trade-off to introduce a performance regression for those who don't need this feature

PR #433, opened 2015-04-08
Path string for the keyPath

andrewluetgers:


andrewluetgers:
Thats odd, that test passes for me.

leebyron:
Thanks for your exploration on this.

This concept has come up a few times before (#237, #310) and has some serious flaws that keep it from being part of the core library.

* General disagreement on key path format. Some want dots, some want JSONPath, and there are other competing standards. I don't think it's the place of this library to make a call on this.

* Unclear how to support keys which include the character `.`. It also encourages client code to build these key paths, and it becomes easy for a client to write the code `keys.join('.')` which could result in some pretty crazy and hard to debug errors if one of those keys happened to have a `.` in it.

* Ambiguous support for indexing into Lists. Should `"a.1.3"` translate to `["a", "1", "3"]`, or `["a", 1, 3]`, or `[a, "1", 3]`, or `["a", 1, "3"]`. All are legitimate and different interpretations.

* Missing support for complex keys. If my Map is keyed on some complex key type, I can't use keyPath. It's a bit arbitrary for the feature to exist but only support a subset of the library's capabilities.

* General concern about extending the polymorphic shape of these functions. Passing a string where a keyPath is expected is actually a pretty easy mistake to make - usually you intended to use a different method (e.g. passing a single key to any of the `*In` methods when you meant to use the non-deep variant). The purpose of TypeScript and Flow is to help detect these kinds of errors. This change will remove that pitfall detection.

Since the implementation of this is usually very specific to a certain application's assumptions and also usually very simple (in your case, just a `split`), I recommend including a helper function in your own library code. It's slightly more verbose, but also gives you complete control.

```js
// Vanilla Immutable.js
m.getIn(['a','b','c']);

// Your proposal
m.getIn('a.b.c')

// Using a helper function
m.getIn(p('a.b.c'));
```

Where `p()` can be whatever you want, even just `p = pathStr => pathStr.split('.')` as proposed here, but also could be a JSONPath parser, or as I proposed in #310, `p = pathStr => pathStr.split('.').map(k => parseInt(k, 10) || k);`

andrewluetgers:
All good points. However the fact that it's come up several times should add a little weight to perhaps looking at it again. But I get it. 

I don't expect a change in this policy but in my defense I was aware of all of these issues for the most-part. Complex keys into lists (as opposed to js arrays) was a new one on me. The array of strings interface is a klunky interface to work with which is why the issue is being brought up. The string notation in that sense is a nice-to-have not a be all end all for every use case. It's intended as an option. For complex keys you can always fall back to the array notation. As for debugging you have access to the parsed keyPath. "Passing a string where a keyPath is expected is actually a pretty easy mistake to make" yes I ran into this frequently and it seems it should not be a mistake, an array of a single string vs just a single string, interface wise it makes sense to me but then again i liked jQuery's interface some people think its an crime to have polymorphic interfaces, I think its nice. In my mind why even bother with get vs getIn just have one function that does both. But thats a matter of taste. 

I'll use a helper and be done with it then.

leebyron:
>  In my mind why even bother with get vs getIn just have one function that does both. But thats a matter of taste.

It's also different functionality! Since Maps can have complex keys, the array itself could be a key into a Map, so having one function that does both would limit these uses.

pedroteixeira:
Any reasons for not using json pointer syntax? It already defines how to escape the seperator chat.

leebyron:
All the same reasons, @pedroteixeira. Separator escaping is one of many issues. JSON path (and anything with escaping rules) generally still has a "whoops, forgot to escape that" pitfall anyhow.

pedroteixeira:
I was just curious if there will be the support for getting into `{a:{"b.c": 1}}`? I was thinking about JSON Pointer, which uses `a/b/c` and in this case, would encode the path to `a/b~1c/`. 

For example: even with this string support, I'll continue translating to arrays using a custom function, given a string and underlying schema - because you would have getIn accepting strings but updateIn would have have to assume too much in order to solve ambiguity. Even getIn can be ambiguious and probably not deterministic, if you have both number and strings in a Map for instance.

Just an idea: Is it possible to add new methods to Immutable objects prototype? Isn´t this kind of stuff better as optional plugins, defining new methods? Perhaps immutable-js would benefit a lot from having ways of having contrib libs, adding new helper functionality.

leebyron:
Your suggestion on contrib libs is having access to adding things to the prototype?

Certainly anyone could mutate the Immutable.js prototypes today, I wouldn't recommend doing that instead of using functions, but it would certainly work.

just-boris:
Hello.

I am using Immutable.js and also found lack of this feature. @leebyron points are fair enough but there is a lot of projects which don't have `.` in key names and also use numbers only as array indexes. So, that corner cases is not an issue for them and they deserve to have prettier API.

Helper function can be solution, but not enough because then you have to import this function into every file in your project, rather than just accessing values with built-in method.

As for me, adding extra methods like `getPath`/`setPath`/`updatePath` seems good solution. It doesn't change anything in existing features, but gives better experience for popular use cases.

I considered to create external wrapper around Immutable, but then figured out, that it is not possible because Immutable itself know and uses only built in data-structures, and I can't create extended version of them. So it is possible only with changes in core.

PR #435, opened 2015-04-08
Fix readme

pomerantsev:
It was probably just a typo.


leebyron:
Not necessarily a typo, it just made an assumption that you had imported a local Seq variable from the immutable module. I agree this is more clear though. Thanks!

PR #436, opened 2015-04-08
Ability to define callback as default value for Record

netgusto:
Had to fix a test for Map on my system, but Travis thinks otherwise it seems.


leebyron:
This conflicts with the work happening on the 4.0 branch (see: https://github.com/facebook/immutable-js/commit/5001072516a65b1245b528f88556ae4d1e4871fe).

But perhaps there's another way to solve your problem. Could you actually explain the problem you're solving? I understand what this code is doing, but I'm just not quite sure why you would want to do it.

netgusto:
I have a Record, and would like the ability to express a default value that is not static, but the result of an expression evaluated when default values are computed (during the record initialization).

Typical example: default value with a string representing the current time:

```javascript
class MyRecord extends Immutable.Record({
    reference: null,
    status: Status.AWESOME,
    availableSince: function() { return moment().toISOString(); }
});
```
Of course, when a MyRecord instance is reified, if a value is given for 'availableSince', the expression should not be evaluated; also, the value may be overridden a later time (later than the initialization) on the same record.

Is there another, already implemented way to do this I'm not aware of ?

leebyron:
Ah, okay, I see what you're saying.

Perhaps a better way to do this is with a custom factory function:

```js
var _MyRecord = Immutable.Record({
  reference: null,
  status: Status.AWESOME,
  availableSince: null
});
function MyRecord(initialValues) {
  if (!initialValues[availableSince]) {
    initialValues[availableSince] = moment().toISOString();
  }
  return _MyRecord(initialValues);
}
```

leebyron:
If you're using the class syntax, this would be a constructor function which calls `super()`

netgusto:
Well OK, I guess this might work yes.

Thanks. I still think that it might be a nice featurette, mainly for the greater expressiveness it conveys.

PR #445, opened 2015-04-14
Explicit sliceSize comparison to 0; is NaN for an unbounded slice.

benmosher:
I made a guess as to where this test should go, I'm not confident I got it right.

Without this change, I'm seeing an exception on a (re)generator-fed Seq if I skip(1), as it does not create the parent iterator.


leebyron:
Thank you for investigating this and for this solution! I had to make some changes, but the approach is the same and the test is very close to the one you wrote.

PR #446, opened 2015-04-16
fromJS: should return the same result if used with or withour converter

alexeyraspopov:
Just simple example of using `fromJS`:

    var data = Immutable.fromJS({ data: { n: 13 } });

    data.toString(); // "Map { "data": Map { "n": 13 } }"

But, if I use converter (no matter how), I'm getting another result:

    var data = Immutable.fromJS({ data: { n: 13 } }, (k, v) => v);

    data.toString(); // "Seq { "data": Seq { "n": 13 } }"

It's break my code and I need to write additional conversion `toMap`.

If you agree with importance of this PR I can add unit tests for this case.

Thanks,
Alexey Raspopov

### DIFF
--- a/src/fromJS.js
+++ b/src/fromJS.js
@@ -17,10 +17,10 @@ export function fromJS(json, converter) {
 
 function fromJSWith(converter, json, key, parentJSON) {
   if (Array.isArray(json)) {
-    return converter.call(parentJSON, key, IndexedSeq(json).map((v, k) => fromJSWith(converter, v, k, json)));
+    return converter.call(parentJSON, key, IndexedSeq(json).map((v, k) => fromJSWith(converter, v, k, json)).toList());
   }
   if (isPlainObj(json)) {
-    return converter.call(parentJSON, key, KeyedSeq(json).map((v, k) => fromJSWith(converter, v, k, json)));
+    return converter.call(parentJSON, key, KeyedSeq(json).map((v, k) => fromJSWith(converter, v, k, json)).toMap());
   }
   return json;
 }


leebyron:
This is actually counter to the purpose of the converter function. If you omit it, a default converter is used which converts Arrays to Lists and Objects to Maps, however the converter function is there to allow for overriding this behavior.

For example, if I wanted Objects to produce OrderedMaps and Arrays to produce Sets, I could write a converter function to do so. However, after this change, my OrderedMaps would get converted back into Maps and my Sets back into Lists.

I'm curious if you think this functionality is presented in a flawed way - or if the documentation around it just didn't make these cases clear enough?

leebyron:
I'm also curious for what purpose you're using the converter function? Your example shows just using an identity function - but presumably that's not particularly useful in practice?

alexeyraspopov:
My case is OrderedMap of Records.

```javascript
var data = {
  todos: {
    'bc72a': { id: 'bc72a', text: 'rule the world' },
    'f45g9': { id: 'f45g9', text: 'do something' }
  }
}
```

```javascript
var Todo = Immutable.Record({ id: '', text: '', completed: false });

var immData = Immutable.fromJS(data, function(k, v){
  if(k === 'todos'){
    return v.toOrderedMap().map(function(t){
      return Todo(t.toObject());
    });
  }

  return v;
});
```

alexeyraspopov:
Just to clarify my point

> This is actually counter to the purpose of the converter function. If you omit it, a default converter is used which converts Arrays to Lists and Objects to Maps, however the converter function is there to allow for overriding this behavior.

I agree with this behavior, but in case of complex structures, if I want to change only one or two of children I expect the default behavior for other structures.

alexeyraspopov:
Any updates? @leebyron

leebyron:
Sorry for the delay, I've been out for a while. I'll get back to this really soon! I'm excited to get this right

alexeyraspopov:
@leebyron I just want you to remember that this PR was created 3 month ago :) Can we decide what to do next?

leebyron:
This PR still has the primary concern of coercing the initial inputs into concrete structures first, which is at best a serious performance penalty but at worse an unexpected coercion.

If you could find a way to handle this such that intermediate concrete structures are not created, I think you would be onto something.

alexeyraspopov:
I haven't been using Immutable for all this time, but still feel guilty because of not finishing this. @leebyron feel free to close this to have less noise in PR list.

PR #449, opened 2015-04-17
hasIn does not take a notSetValue

donabrams:
hasIn(searchKeyPath) {
    return this.getIn(searchKeyPath, NOT_SET) !== NOT_SET;
},


leebyron:
Yep! Thanks for the fix.

PR #452, opened 2015-04-19
Fix isPlainObj function for objects from different context

Elergy:
We can't create immutable object with Immutable.fromJS if source object has been created in different context (e.g. we want to create immutable object in chrome extension's popup, but source for immutable object is received from background page).

We have plain JS object in the background page:

```
var data = {
    item1: {
        name: 'Name 1'
    },

    item2: {
        name: 'Name 2'
    }
};
```

In the popup.js script we want to get this object and create immutable version of it:

```
chrome.runtime.getBackgroundPage(function(page) {
    var data = page.data;
    var immutableData = Immutable.fromJS(data);
});
```

but we can't because 'data' has been created in different context (in the background page). 

```
immutableData === data; //true
```

If we serialize and deserialize data we can have immutable version:

```
var dataAfterSerialization = JSON.parse(JSON.stringify(data));
var newImmutableData = Immutable.fromJS(dataAfterSerialization);

newImmutableData.toString() === 'Map { "item1": Map { "name": "Name 1" }, "item2": Map { "name": "Name 2" } }'; //true
```

Full example is here - https://github.com/Elergy/Immutable-extension-test


src/fromJS.js
```diff
@@ -36,5 +36,9 @@ function fromJSDefault(json) {
 }
 
 function isPlainObj(value) {
-  return value && (value.constructor === Object || value.constructor === undefined);
+  return value &&
+    (value.constructor === Object ||
+      value.constructor === undefined ||
+      Object.prototype.toString.call(value) === '[object Object]'
```
Elergy: slowest action in the end. For applications which have only one context it won't be a performance-breaker.

leebyron:
The problem with this approach is that it becomes very expensive. `isPlainObject` is called many times through Immutable.js and changing it from an O(1) to an O(N) operation to support an obscure use case is probably the wrong tradeoff.

I would strongly suggest using a custom version from `fromJS` which supports exactly the characteristics you're looking for if the version that comes with Immutable doesn't work for you.

If your data that's been created in a different realm is safe to convert (e.g. has no complex or exotic objects - every object can be safely replaced with Map), then you can use a very simple implementation of fromJS:

```js
function fromJSGreedy(json) {
  return Array.isArray(json) ?
    Immutable.List(Immutable.IndexedSeq(json).map(fromJSGreedy)) :
    Immutable.Map(Immutable.KeyedSeq(json).map(fromJSGreedy));
}
```

facebook-github-bot-4:

happypoulp:
@leebyron what about using `Object.prototype.toString` instead of `=== Object`?

```javascript
function isPlainObj(value) {
  return value && (Object.prototype.toString.call(value) === '[object Object]' || value.constructor === undefined);
}
```

If this test is robust enough and don't hurt performance it could be a nice fit.

By the way, I encountered the same issue using a JS SDK that was returning a plain JS Object created from another frame.

lacker:
I think this pull request has timed out, so I am going to close it. I think it would be an improvement to have something `O(1)` here that just works a bit better so if you are interested in making that work then please feel free to reopen!

PR #460, opened 2015-05-03
Add zipAll support to IterableImpl

mduvall:
This commit adds `zipAll` support which zips over the iterables and uses the length of the longest iterable. There is a check to ensure that the invariant of the length of the longest iterable is always the size of the zipped value and a test to make sure leaves the holes in remaining zipped values are set to `undefined`.


src/Operations.js
```diff
@@ -767,9 +774,11 @@ export function zipWithFactory(keyIter, zipper, iters) {
         steps = iterators.map(i => i.next());
         isDone = steps.some(s => s.done);
       }
+
```
mduvall: Added white space for breathing room, can remove if it is inconsistent in style.

src/Operations.js
```diff
@@ -726,9 +726,15 @@ function maxCompare(comparator, a, b) {
 }
 
 
-export function zipWithFactory(keyIter, zipper, iters) {
+export function zipWithFactory(keyIter, zipper, iters, zipAll) {
   var zipSequence = makeSequence(keyIter);
-  zipSequence.size = new ArraySeq(iters).map(i => i.size).min();
+  var iterSizes = new ArraySeq(iters).map(i => i.size);
+  var zipSeqSize = !!zipAll ? iterSizes.max() : iterSizes.min();
+  if (zipAll) {
```
mduvall: Any opinions here? I was about to delve into the sequence's iterator but thought that might be much but went the route of normalizing the iterators lengths by just throwing on the `undefined`s at the end.

leebyron: Yeah this concat could be expensive. I think the logic to decide when to end the iteration should exist within the iterator.

src/Operations.js
```diff
@@ -767,9 +774,11 @@ export function zipWithFactory(keyIter, zipper, iters) {
         steps = iterators.map(i => i.next());
         isDone = steps.some(s => s.done);
```
leebyron: I *think* that this is the critical point. a zipAll is done when *all* steps are done instead of *some* steps are done.

mduvall: Yeah that sounds right to me, let me see if I can get that to work.

facebook-github-bot-4:

PR #466, opened 2015-05-07
Reword comment to not use "require"

zertosh:
browserify does a [quick](https://github.com/substack/node-detective/blob/75946ee23cb7abb974a2053235ecb2663b879c07/index.js#L4) regexp match for `\brequire\b` before deciding on whether to build an AST to get the require calls. By rewording this comment so that the word "require" is nowhere in the source, you'd be improving everyones build times. Though Browserify has `noparse` option, this is an easy win.

I'm not sure why uglify decided to remove some line breaks for me but not for whoever ran the build last - but I'm not going to question it.


leebyron:
#seemslegit

PR #475, opened 2015-05-17
Add renameKeys to Map

english:
Similar to [`clojure.set/rename-keys`](https://clojuredocs.org/clojure.set/rename-keys), it returns a `Map` with the keys in `keyMap` renamed to the values in `keyMap`.

I find this function really handy when putting a `Map` through a series of transformations and can only imagine its a common use case.

```js
Map({ a: 1, b: 2, left: 'alone' }).renameKeys({ a: 'y', b: 'z' });
// => Map { "y": 1, "z": 2, "left": "alone" }
```


leebyron:
Perhaps better to use `mapKeys` for this?

leebyron:
Also - is this function trivial enough that it may be more appropriate to be defined when used?

```js
function renameKeys(keyMap) {
  keyMap = Iterable(keyMap);
  return myMap.mapKeys(k => keyMap.get(k, k));
}
```

english:
You're right, that's a much nicer implementation.

I can see your point that it's trivial to write if needed with `mapKeys`. I guess it's your call; I'm sure it would be a useful API, but you're the one that will be maintaining it :)

PR #496, opened 2015-06-08
Fix isSuperset error on Array arg

matthewwithanm:
According to [the docs][1], `isSuperset` should accept an Array. This adds a check similar to the one in `isSubset` to make sure that's true (:

I wasn't sure where to add the test so I put it in `Set.ts`. Let me know if there's a better place!

[1]: https://github.com/facebook/immutable-js/blob/013d52b58d6c7497a6c791cd5950acc32a7340b8/type-definitions/Immutable.d.ts#L2070


leebyron:
Thanks!

matthewwithanm:
np (:

Issue #459, opened 2015-05-03
Calling toJS() (and other methods?) on Objects with _capacity: NaN results in infinite loop


batashi:
Example:

```javascript
var list = Immutable.List([1, 2, 3, 4, 5]);
var newList = list.slice(0, NaN) // newList._capacity = NaN and newList.size = NaN
list.toJS() // causes infinite loop
```

I'm not sure whether the issue lies with the newList object being incorrectly formatted or toJS() not being able to parse that kind of object. It is probably the former as I believe I've gotten an infinite loop calling some method on that kind of object other than toJS().

leebyron:
Thanks for the report!

facebook-github-bot-4:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

leebyron:
Fixed by #511

PR #502, opened 2015-06-13
Fix infinite loop in slice from #459

saromanov:
Hi!
In the Issue #459 was described problem with infinite loop after calling ```slice```.
I found, that this problem not only with List, but with all types, which call function ```wholeSlice``` from the file TrieUtils.js. Infinite loop starts every time when ```slice``` parameters begin or end is not integer
```javascript
var vec = Immutable.List([7,4,5,8]);
result = vec.slice(0,"A");
console.log(result); //infinite loop
//Same thing in console.log(vec.slice(NaN,NaN));
//or console.log(vec.slice("A","B"));
//or result.toJS()
```
I fixed this and now, every time when slice gets non integers parmeters, it returns valid value
```javascript
var vec = Immutable.List([7,4,5,8]);
console.log(vec.slice(0,"A")); //List [7,4,5,8]
```


saromanov:
Hm, it looks that node v0.10.38 is not supported ```Number.isInteger()```. In the 0.12.4 it forked fine. Of course, it can be replaced with ```Math.round(num) === num ``` or something like this, but it will be not ES6 harmony style :)

leebyron:
To detect NaN you can just do `value !== value`, NaN is the only value which fails that test.

Infinite recursion is bad news, so it would be nice to invariant here, however what I'd love to see is a test that `_capacity` can never become NaN in the first place.

saromanov:
Unfortunately, it looks, what test case probably is not complete in the place. in the ```slice.ts```, checks only case with NaN, but not checks another types(float, string). Detect only NaN is not enough. Also, this problem occurs after toJS method and this case also missed in the ```slice.ts```.
I think, i will provide this test cases and test case with ```__capacity```

leebyron:
Thanks again for this investigation which inspired #511, fixing #459

PR #510, opened 2015-06-18
Trailing commas in dist/immutable.js

ialexryan:
See commit for five locations. These are bad, especially in IE8 etc.

### DIFF
--- a/dist/immutable.js
+++ b/dist/immutable.js
@@ -4814,7 +4814,7 @@
 
     keySeq: function() {
       return this.valueSeq();
-    },
+    }
 
   });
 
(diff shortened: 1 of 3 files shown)


jeffchan:
Thanks for fixing this. I'd love to see this merged, so closure compiler doesn't give any warnings.

leebyron:
Sorry for not spotting this sooner - this is great but the /dist/ files are auto-generated as output from /src/, so if landed as is, this work would be quickly replaced when the next PR is landed.

Could you ensure this change is applied to /src/ instead of /dist/?

leebyron:
Merged (and squashed) in 35658506c294121ad3990745cccbe0f1980b3504. Thanks!

PR #511, opened 2015-06-18
`List::setSize()` and `List::slice()` now handle non-numbers more gracefully

steveluscher:
As @batashi pointed out in #459, some list methods (`slice()` and `setSize()`) were not resilient against `Infinity` and non-numeric arguments. This PR introduces a specification by way of some new `List` tests, and an implementation to make it so.


leebyron:
Merged in a3b83611a32912bccb779057b35367366cc5c417

PR #520, opened 2015-06-26
fix #515 Empty set union with Ordered Set returns Ordered Set

### LINKED ISSUE #515: Empty Set unioned with non-empty OrderedSet returns OrderedSet
```javascript
> Immutable.Set([1]).union(Immutable.OrderedSet([1]))
Set { 1 }
> Immutable.Set().union(Immutable.List([1]))
Set { 1 }
> Immutable.Set().union([1])
Set { 1 }
> Immutable.Set().union(Immutable.OrderedSet([1]))
OrderedSet { 1 }
```

Expected `union` to always return a Set, as written in the docs.

ellbee:
Added an extra check to the Set constructor so that it does not short circuit and return argument if that argument is an Ordered Set. Test added.


leebyron:
Thank you for this as well, I extended your approach to apply to Map as well. The same issue could have affected Map vs OrderedMap

PR #521, opened 2015-06-28
fix #480 filter.take regression

### LINKED ISSUE #480: filter().take() regression in v3.7.3
There seems to be a regression in v3.7.3 that causes `take(n)` to fill the resulting array with up to `n` `undefined`s.

```
$ npm install --save immutable@3.7.2
immutable@3.7.2 node_modules/immutable
$ node
> var Immutable = require('immutable')
undefined
> Immutable.Map({a: 1, b: 2}).valueSeq().filter(function (n) {return n > 1}).take(10).toJS()
[ 2 ]
$ npm install --save immutable@3.7.3
immutable@3.7.3 node_modules/immutable
$ node
> var Immutable = require('immutable')
undefined
> Immutable.Map({a: 1, b: 2}).valueSeq().filter(function (n) {return n > 1}).take(10).toJS()
[ 2, , , , , , , , ,  ]
```

Strangely enough, it doesn't seem to happen without filtering beforehand.

Thanks in advance,
Tim

ellbee:
```javascript
var a = Immutable.Seq([1,2,3,4]).filter(function(val) {
  return val < 3;
}).take(10).toArray();
```

take(10) in the preceding code will create a sliceSeq with size 10, which then causes toArray() to construct an array of 10 elements instead of the expected 2. I think that in this case the sliceFactory called by take() should not set the size for the sliceSeq, because the filterSeq it is slicing does not have a defined size.

My attempted fix is to reinstate a line of code that was changed in commit bcd51c0 which checks if the sequence to be sliced has a valid size before setting sliceSeq.size.


leebyron:
Ah, thank you for your investigation and straight-forward fix with comments. Excellent work!

PR #531, opened 2015-07-03
fail test

skv-headless:
I've expected map behaviour like for other Iterable object instead it returns undefined.


leebyron:
I'm working on v4.0 which will include a major reworking of `Record` to no longer be `Iterable` nor have most of the collection API. For example, `map` in this case doesn't make a lot of sense.

Of course, the current behavior of returning undefined is pretty awful, but the new behavior will be that `map` is not even defined :)

Issue #540, opened 2015-07-08
.setSize() retains values previously discarded with .slice()


vlinder:
```js
Immutable
    .fromJS([1,2,3,4,5])
    .slice(0,2)
    .setSize(5)
    .forEach(function(val){console.log(val)});
```
Outputs:
```
1
2
3
4
undefined
```

One would not think that you get the values removed by slice back again by setting size. Somehow the last value is the only one which actually gets removed.

From the documentation of setSize()
>If size is greater than this List's size, the new List will have undefined values for the newly available indices.

Same output when using using .setSize() twice instead of slice().
```js
Immutable
    .fromJS([1,2,3,4,5])
    .setSize(2)
    .setSize(5)
    .forEach(function(val){console.log(val)});
```

Might be browser specific?
### Tested with
* Chrome Version 43.0.2357.130 (64-bit) on OS X 10.10.
* Firefox 38.0.5 on OS X 10.10
* Firefox 39.0 on OS X 10.10
* Chromium 43.0.2357.132 on Arch Linux

leebyron:
Thank you for the report

PR #542, opened 2015-07-08
Added tests for issue #540, List.setSize() and List.slice() keeps tru…

vlinder:
…ncated values.

These tests currently fail until someone fixes the issue.


leebyron:
This has been fixed, and I've merged in your tests to ensure it

PR #544, opened 2015-07-09
Added ImmMap and ImmSet aliases for ES6 destructs

sampepose:
ES6 destructuring is handy, but I don't want to override built-in ES6 Map and Set types. This adds an alias to allow destructuring.

Bad:
`var {Map, Set} = require('immutable'); // overrides ES6 Map and Set`

Good:
`var {ImmMap, ImmSet} = require('immutable');`

This parallels Hack immutable type naming conventions.


leebyron:
I'd like to be not prescriptive about this and let individuals decide for themselves how they would like to name these if they use them in concert with the ES6 collections.

Using your example of ES6 destructuring, I would recommend:

```js
var { Map: ImmMap, Set: ImmSet } = require('immutable');
```

leebyron:
Using ES6 module syntax, the same can be accomplished with:

```js
import { Map as ImmMap, Set as ImmSet } from 'immutable';
```

leebyron:
Closing this, hoping the suggestions above are sufficient. Let me know if not.

Issue #486, opened 2015-05-28
get or getIn on List using a key that is a string, object, or void 0 does not return correct notSetValue


dashed:
``` js
const {List} = require('immutable');

const foo = List();

const NOT_SET = {};
console.log(foo.get('bar', NOT_SET) !== NOT_SET); // => true (expected false)
console.log(foo.getIn(['bar'], NOT_SET) !== NOT_SET); // => true (expected false)

// this regression breaks .hasIn()
console.log(foo.has('bar')); // => false
console.log(foo.hasIn(['bar'])); // => true (expected false)
```

Dashed:
Using `immutable@3.7.3`.

leebyron:
Thank you for the report!

Dashed:
I tried other types of keys:

```js
const {List} = require('immutable');

const foo = List();

[
    'bar',
    true,
    false,
    null,
    void 0,
    {},
    42,
    42.5
].forEach(function(key) {

    const NOT_SET = {};

    let ret = false;

    ret = (foo.get(key, NOT_SET) !== NOT_SET) || ret;
    ret = (foo.getIn([key], NOT_SET) !== NOT_SET) || ret;

    ret = foo.has(key) || ret;
    ret = foo.hasIn([key]) || ret;

    if(ret) {
        console.log('broken for key', key);
    }
});
// broken for key bar
// broken for key undefined
// broken for key {}
```

leebyron:
Thanks for your investigation!

mhagmajer:
+1

mhagmajer:
```js
Immutable.List([7]).getIn(0) // throws error whereas 7 was expected
Immutable.List([7]).getIn('0') // returns 7 as expected

leebyron:
It looks like the second should probably also throw an error. getIn expects an Iterable collection of keys. Strings are Iterable but often not considered collections because that leads to confusion. 

Would you please open a separate issue for this other issue?

mhagmajer:
Thanks @leebyron, I realized that I was calling `getIn` incorrectly. I created a new issue at https://github.com/facebook/immutable-js/issues/940 for strings in getIn.

PR #545, opened 2015-07-10
Fix List notSetValue not working for several key types

darrenburns:
As discussed in #486, using keys which can't be coerced to numbers in Lists ignored notSetValue. This pull request should hopefully solve that problem.


src/List.js
```diff
@@ -57,7 +57,7 @@ export class List extends IndexedCollection {
 
   get(index, notSetValue) {
     index = wrapIndex(this, index);
-    if (index < 0 || index >= this.size) {
+    if (index < 0 || index >= this.size || index !== index) {
```
leebyron: Rather than doing this, we can flip the if check and only do the get if it's safe, rather than short-circuiting.

src/List.js
```diff
@@ -371,6 +379,9 @@ export function emptyList() {
 
 function updateList(list, index, value) {
   index = wrapIndex(list, index);
+  if (index !== index) {
+    throw new Error('Cannot set item at unknown index "' + index + '" on ' + listName(list));
```
leebyron: No other set method throws like this at the moment, so this would be a breaking change. I think this may be the right thing to do long term, but for now, let's make it silent.

src/List.js
```diff
@@ -65,6 +65,14 @@ export class List extends IndexedCollection {
     return node && node.array[index & MASK];
   }
 
+  has(index) {
```
leebyron: This seems identical to the `has` implementation of `IndexIterable` which is already in the prototype chain.

src/TrieUtils.js
```diff
@@ -57,7 +57,12 @@ export function ensureSize(iter) {
 }
 
 export function wrapIndex(iter, index) {
-  return index >= 0 ? (+index) : ensureSize(iter) + (+index);
+  if (index === '' || isNaN(+index)) {
```
leebyron: We actually want to implement (as best we can without being breaking) the ECMA rules for detecting and coercing of "array index".

leebyron:
I made the changes pointed out in the inline comments atop your rev, in case you want to inspect what changed.

Thanks again for investigating this!

leebyron:
See https://github.com/facebook/immutable-js/commit/8031a8b8b6a74701f6c189ad62af41643b60f5e0 for final code

PR #547, opened 2015-07-14
Update `bundle` task to build each immutable class individually.

ryan953:
Bundle each class into it's own file so consumers can use only the parts that they need. Minified versions are included as well.

I've left the tests running against the full-build, and havn't split up the type file.


leebyron:
I think getting the type to work correctly is probably the next step in getting this right. I like this direction.

I'd also like to find a way to be able to require `immutable/Map` rather than `immutable/dist/Map` without causing too much turmoil in the build script and library developer experience.

leebyron:
Thanks for starting to explore this!

ryan953:
@leebyron What do you think is the best way to split the typedefs up? I started looking at importing `typescript-services` and some of the other stuff that the gh-pages branch has but it feels like overkill. How do you feel about manually splitting `Immutable.d.ts` into parts and then recombining it during the build step?

clintwood:
@leebyron, @ryan953, its worth looking at [`rollup`](https://github.com/rollup/rollup) and specifically whats discussed [here](https://github.com/rollup/rollup#a-better-approach) not as a tool for building Immutable.js but rather as an approach for packaging in general... Using ES2015 modules like is [done here](https://github.com/lodash/lodash/tree/3.10.0-es) makes it easy for modern tooling like `rollup` to pull only what is needed to reduce distribution footprint of web apps.

Another thing, providing an ES6+ package like `lodash` have done enables the module user to build their distribution based on their target environment which for cases when using recent `iojs` distributions eliminates a lot of unnecessary cruft from ES5 polyfilling. ;)

leebyron:
Thanks for the suggestion, @clintwood - I'll look into that.

@ryan953 - I'm not totally sure, but that's something we'll need to solve in order to make this really work for everyone.

klaemo:
This PR would enable something like this:
```js
const Map = require('immutable/map')
// or
import Map from 'immutable/map'
```
Basically the way `lodash` does it?! That would be pretty sweet :)
I was surprised by the size of my bundled app after including immutable ;)

milesj:
It's been a year now. Is there any reason why individual modules do not exist yet...?

lacker:
I think this pull request has essentially timed out. I am going to close it but if someone still actively wants to work on this then please feel free to reopen it!

andrewmclagan:
Would love an update if anyone has time, where this is at >>?

Issue #538, opened 2015-07-06
Implementing the transducers protocol


acdlite:
Would you accept a PR to update ~~Map and~~ List (and any other types that make sense) to support the transducer protocol? Like this: https://github.com/jlongster/transducers.js#the-transducer-protocol It should be a minimal addition.

Sorry if this has been asked and answered already. I did a search and only found this comment https://github.com/facebook/immutable-js/issues/23#issuecomment-60796072

leebyron:
Go for it!

KyleAMathews:
:100:

acdlite:
@leebyron Great! I'll add it to List, Set, and Stack. Keyed collections like Map don't make sense, unless I mistaken.

Dashed:
@acdlite Any Immutable collection that is iterable should work (this includes Map).

For Map, the merge operation would probably be suitable for the `@@transducer/step` protocol.

DaleJefferson:
@acdlite What are the advantages in using transducer.js over using immutable.Seq?
If transducer.js is better than Seq should it be replaced? Or do they do different things?

dnalot:
Stumbled on this and wondered why it stagnated? The PR looks superficially solid besides needing feedback for tests,

leebyron:
I still think this one is a pretty cool idea - though I've been out of the loop on how often transducers are being used.

mhelmer:
Seems like the PR got closed due to lack of tests. If `transducers.js`or `transducers-js` was added as a `devDependency`, it wouldn't be hard to test it using `transduce`, `into` and so on. Could definitely look into it.

leebyron:
Better late than never

PR #550, opened 2015-07-16
Add transducer protocol to Immutable Data Structures

donabrams:
Completed as per https://github.com/cognitect-labs/transducers-js/issues/20. cc: https://github.com/facebook/immutable-js/issues/538.

I ran some tests locally (see https://gist.github.com/donabrams/3f42e7017170b52593f5), but I was using an implementation (https://github.com/transduce/transduce) to test this rather than some standalone code. I'm not sure what tests would be appropriate to commit here.


leebyron:
Hey @donabrams let's update this and get it working!

It looks like the transducer protocol suggests that it use symbols when available, should this do the same?

You can look at how Immutable implements the Iterator protocol for an example.

donabrams:
It looks like the double at strings were used at the time, since symbols weren't ready yet. Only Gonzala used symbols:
[`Symbol.for("transducer/step")`](https://github.com/search?utf8=%E2%9C%93&q=%22Symbol.for%28%5C%22transducer%2Fstep%5C%22%29%22&type=Code&ref=searchresults) vs. [`"@@transducer/step"`](https://github.com/search?utf8=%E2%9C%93&q=%22%5C%22%40%40transducer%2Fstep%5C%22%22&type=Code&ref=searchresults)

leebyron:
Ok, we can omit Symbols for now and add them later. This is cool.

Could you please include some simple test cases? That will help ensure these never break in the future, but would also document the minimal use cases.

donabrams:
Yeah, I originally drew a blank when deciding what tests I'd write for this. Since these interfaces are declarative but imply that these data structures hold certain properties, I'd think the right test would be around the property invariants. Really, there should be a generative/invariant test for anyone implementing transducer protocol but I couldn't find one :( I'll update this w/in the next few days.

keeth:
Would love to see this merged.. anything an outsider could do to help move it along?

lacker:
This pull request seems like it has gotten stalled in "needs tests" state, and I theorize it is stuck there indefinitely, so I am going to close it. If someone is interested in working on this then I think it could definitely still be a useful area to work on though.

PR #551, opened 2015-07-16
added method keyPath to Cursor

andrejewski:
A lot of people including myself need a way to generate a key path between a root and node of a deeply nested data structure. This method does that. See the tests for examples.

I also see you are moving toward pulling Cursor out of Immutable. That's awesome. Maybe it would be best to move keyPath to be a static method on Immutable since getIn() and friends need a keyPath, or I'll just put it into a separate module.

### DIFF
--- a/contrib/cursor/__tests__/Cursor.ts
+++ b/contrib/cursor/__tests__/Cursor.ts
@@ -17,6 +17,43 @@ describe('Cursor', () => {
 
   var json = { a: { b: { c: 1 } } };
 
+  describe('keyPath', () => {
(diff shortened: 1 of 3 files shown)


vojtatranta:
+1

lacker:
Hey, so right now we wish Cursor were actually separate from Immutable itself. We are considering kicking it out into its own repo or something like that. So if you're interested in maintaining it separately let me know! - otherwise I am going to close this pull request since Cursor-in-the-immutable-main-repo is something we would like to deprecate.

PR #553, opened 2015-07-18
Typescript: declare Immutable as a module too

jmalonzo:
I think this helps in certain situations where Immutable was imported via a <script/> tag, as the module is imported as Immutable and not "immutable".


leebyron:
Thanks! Could you finish your sentence? When is this useful?

Also, if you could please squash your commits then I can merge this!

PR #555, opened 2015-07-21
Add ability to group cursor operations

mikaelbr:
First of all, sorry for the wall of text. I'm trying to cover the argumentation, why this is needed and how this can open for more powerful patterns, so we can have a discussion around it. A lot of times there is a need to group operations on cursors, much like withMutations, but not with passed values but the actual cursors. Like so:

``` js
var data = Immutable.fromJS({'a': 1});
var c1 = Cursor.from(data, ['a'], onChange);
var c2 = c1.groupedOperations(cursor => cursor.set(2));
```

In contrast to 

``` js
var data = Immutable.fromJS({'a': 1});
var c1 = Cursor.from(data, ['a'], onChange);
var c2 = c1.withMutations(cursor => cursor.set(2)); // Would fail
// This would also fail as c1 points to a scalar
c2 = c1.withMutations(value => 2); // Would fail

// Would instead have to go "one level up" to point to a immutable structure:
c1 = Cursor.from(data, onChange);
c2 = c1.withMutations(value => value.set('a', 2)); // Would work
```

By having `groupedOperations` (or with another name?) you could have seperate operations/functions and group/compose them to one, and behave as if it was one with only one triggered `onChange`.

Problem is, if you have the following:

``` jsx
var square = cursor => cursor.update(input => input * input);
var timesTwo = cursor => cursor.update(input => input * 2);
```

you could do:

``` jsx
var c2 = square(c1);
```

and that would work as expected. You could also do:

``` js
var c2 = square(timesTwo(c1));
```

but that would trigger `onChange` twice, which is in many cases less than optimal. How ever, with `groupedOperations`, you can pass in composed functions and only trigger one update:

``` js
var c2 = c1.groupedOperations(compose(square, timesTwo));
```

As mentioned, you could do something seemingly similar with `withMutations`, but as `withMutations` pass on the cursors underlaying data and not the cursor it self, this would be "semantically different", and our functions would have to be different with composed and non-composed editions (one handling cursors, and one handling the values the cursors are pointing at).

Alternative to `groupedOperations` name can be `bulk` or `grouped`.

### DIFF
--- a/contrib/cursor/__tests__/Cursor.ts
+++ b/contrib/cursor/__tests__/Cursor.ts
@@ -289,6 +289,38 @@ describe('Cursor', () => {
     expect(typeof mapped2.get(0).deref).toBe('function');
   });
 
+  it('can have grouped operations with cursor apply with a single callback', () => {
+    var onChange = jest.genMockFunction();
+    var data = Immutable.fromJS({'a': 1});
+
+    var c1 = Cursor.from(data, onChange);
+    var c2 = c1.groupedOperations(cursor => {
+      // Passed value should be a cursor.
+      expect(typeof cursor.deref).toBe('function');
+
+      return cursor.set('b', 2).set('c', 3).set('d', 4);
+    });
+
+    expect(c1.deref().toObject()).toEqual({'a': 1});
+    expect(c2.deref().toObject()).toEqual({'a': 1, 'b': 2, 'c': 3, 'd': 4});
+    expect(onChange.mock.calls.length).toBe(1);
+  });
+
+  it('passes cursors to grouped operations even if cursors point to scalars', () => {
+    var onChange = jest.genMockFunction();
+    var data = Immutable.fromJS({'a': 1});
+
+    var c1 = Cursor.from(data, ['a'], onChange);
+    var c2 = c1.groupedOperations(cursor => {
+      // Passed value should be a cursor.
+      expect(typeof cursor.deref).toBe('function');
+      return cursor.set(2);
+    });
+    expect(c1.deref()).toEqual(1);
+    expect(c2.deref()).toEqual(2);
+    expect(onChange.mock.calls.length).toBe(1);
+  });
+
   it('can have mutations apply with a single callback', () => {
     var onChange = jest.genMockFunction();
     var data = Immutable.fromJS({'a': 1});
--- a/contrib/cursor/index.d.ts
+++ b/contrib/cursor/index.d.ts
@@ -286,6 +286,16 @@ declare module 'immutable/contrib/cursor' {
      * callback is triggered with the final value.
      */
     withMutations(mutator: (mutable: any) => any): Cursor;
+
+    /**
+     * Every time you change a cursors value you trigger the onChange callback
+     * from Cursor creation. In addition you can get scalar values as values
+     * passed to the updater. This can sometimes be unfortunate if you want to
+     * group a set of operations on a cursor. `groupedOperations` will allow
+     * you to group a set of operations on a cursor and only trigger the
+     * change callback after the updater function is invoked.
+     */
+    groupedOperations(updater: (value: Cursor) => Cursor): Cursor;
   }
 
 }
--- a/contrib/cursor/index.js
+++ b/contrib/cursor/index.js
@@ -198,6 +198,16 @@ IndexedCursorPrototype.withMutations = function(fn) {
   });
 }
 
+KeyedCursorPrototype.groupedOperations =
+IndexedCursorPrototype.groupedOperations = function(fn) {
+  var silentCursor = makeCursor(this._rootData, this._keyPath);
+  return updateCursor(this, function (m) {
+    var result = fn(silentCursor);
+    if (result && result.deref) return result.deref();
+    return result;
+  });
+}
+
 KeyedCursorPrototype.cursor =
 IndexedCursorPrototype.cursor = function(subKeyPath) {
   subKeyPath = valToKeyPath(subKeyPath);


leebyron:
Thanks for your exploration and explanation, but I'm still a little bit confused.

Why could you not just slightly change your example to:

```
var square = input => input * input;
var timesTwo = input => input * 2;

var c2 = c1.update(square);

var c2 = c1.update(square(timesTwo(c1));
var c2 = c1.update(compose(square, timesTwo));
```

This gives you shorter and more versatile definitions for "square" and "timesTwo" -> it operates on the value of the cursor rather than the cursor itself, then you can use the pre-existing "update" function to group operations.

Is there a more real-world use case that you're trying to use that leads to this being necessary?

mikaelbr:
I get the confusion, and what you propose is kind of a viable alternative. But the problem with this is that you have to invoke the update function through cursor updates, and that would potentially allow `cursor.updates` to "leak through your system". I think it's a question of principal/separation here.

Let's say I have a React component:

```js
var Comp = React.createClass({
  render: function () {
     // Partially apply action to get cursor as input (or any other props.
     // Component doesn't care about what type of input it is.
     let onClick = partial(compose(clickAction, logAction), this.props.data);
     return <button onClick={click}>Click me</button>
  }
});
```

```js
// Some other place/file, a collection of functions that just
// manipulate state cursors point at. Separate from the
// Component, and holds the "whole truth" of state
// manipulation.
function clickAction (cursor) {
  return cursor.update(....)
}

function logAction () { /* ... */ }
```

Alternativly, my component would have to be implemented as

```js
var Comp = React.createClass({
  render: function () {
     // Component has to care about whether it's a cursor or not
     let onClick = partial(this.props.data.update, compose(clickAction, logAction));
     return <button onClick={click}>Click me</button>
  }
});

// and clickAction

function clickAction (someSortOfInput) {
  return someSortOfOutput;
}
function logAction () { /* ... */ }
```

Both examples will work, and having more generic/versatile functions as the latter example is smart in many cases, but I'd argue that sometimes you'd want to have the entire truth of your state update in these sort of actions, and have them specific to state-update.

Summarised, I think this

```js
var newCursor = manipulate(oldCursor);
```

is in many cases (in regards to component "actions") better separated with more clear intent for the `manipulate` function, rather than

```js
var newCursor = oldCursor.update(manipulate);
```

Another potential drawback of the second example is multiple inputs. So if a "action" (state update function) requires multiple data as arguments (depends on some sort of data in a immutable tree), you'd have to have the cursors as input anyways, as a mere `cursor.update` wouldn't do.

With something like `groupedOperations` you could potentially group operations across cursors as well, by finding the common parent cursor, updating that by manipulating the parent cursor inside the `groupedOperations` body. If that made any sense.

leebyron:
Hmm, I'm still a bit confused - I don't see where "groupedUpdates" fits into your examples. 

Because of the nature of cursor updates, an "onclick" does need to know about cursors, since calling `cursor.update` and ignoring the result implies the top-level mutation and side-effects that we use cursors for. 

To me, the 2nd example you illustrate there looks like a cleaner implementation. The component on-click is explicit about calling the cursor's side-effecting `update` method while the library functions are more generic since they operate on values rather than operating on cursors.

leebyron:
To give more isolated feedback, if I were doing a code review and came across code that looked like:

```js
function clickAction (cursor) {
  return cursor.update(someSortOfInput => someSortOfOutput)
}
function logAction (cursor) {
  return cursor.update(someSortOfInput => someSortOfOutput)
}
```

With the intention of being reusable or compostable, then I would recommend isolating the `cursor.update` calls outside of these functions. It's the inner function within `cursor.update` which is reusable and composable.

leebyron:
Regardless, I think we may be able to solve this using `update` without you needing to change how your methods work today.

When calling `cursor.update(value => ...)`, `value` will be the unwrapped collection. At this point you can call whatever you want on it, including `update` without triggering a change event on the cursor's root.

```js
let onClick = partial(this.props.data.update, clickAction);
let onLog = partial(this.props.data.update, logAction);
let onClickAndLog = partial(this.props.data.update, compose(clickAction, logAction));

function clickAction (cursor) {
  return cursor.update(someSortOfInput => someSortOfOutput)
}
function logAction (cursor) {
  return cursor.update(someSortOfInput => someSortOfOutput)
}
```

This will still work as expected, each event handler will only result in one change event, since the values passed to `clickAction` and `logAction` are regular immutable collections in all three cases

mikaelbr:
I realise that it's hard to understand my point, and I'm having a hard time formulating why I think this is a very nice to have, but bear with me. Also, this discussion has kind of derailed to be about "actions" and whether or not they should have cursors or data as input, but I think that's kind of the integral part of if `groupedOperations` is needed or not. I think, if it's sensible to have actions which takes cursors as input, this feature is needed.

> Hmm, I'm sorry but I'm still a bit confused - I don't see where "groupedUpdates" fits into your examples.

Thinking of it, it's not really `groupedUpdates` or `groupedOperations`, but it's essentially just an `update` where you're passed the `cursor` you want to update instead of the underlying immutable object (or scalar value).

> Because of the nature of cursor updates, an "onclick" does need to know about cursors, since calling cursor.update and ignoring the result (all React event handlers) implies the top-level mutation and side-effects that we use cursors for.

It wouldn't ignore the result, but return the resulting cursor(s). Actions should be pure, or at least abstract side-effects to lifted functions (returning accessors to async data/promises/observables or what ever).

> To me, the 2nd example you illustrate there looks like a cleaner implementation. The component on-click is explicit about calling the cursor's side-effecting update method while the library functions are more generic since they operate on values rather than operating on cursors.

My examples are fairly contrived and the distinction between the two isn't too clear, but in more real-world examples the actions would probably be larger and probably some use `set`, `remove`, `push` etc, and/or a combination thereof. I think having the ability to pass in the cursors it self (or the immutable objects for that sake) as arguments is more flexible in the sense that you can more dictate the signature of your actions, instead of immutable.js cursors doing it. You can design your internal API with more flexibility and "intent". You could work around that with partial applications and binding of your actions to pass inn additional data, but I'd say that would be semantically weird and the API would be way harder to use/not clean.

> With the intention of being reusable or compostable, then I would recommend isolating the cursor.update calls outside of these functions. It's the inner someSortOfInput => someSortOfOutput functions within cursor.update which are reusable and composable

In the current examples, I'd agree with you. But I didn't necessarily intend to make them reusable across immutable structures, native structures and scalar values in a sense. But reusable across components getting passed cursors as props, if that makes sense. What I tried to argue, was that it held a more complete picture what is happening/shows better intent.


> Regardless, I think we may be able to solve this using update without you needing to change how your methods work today.

Yeah, you're right, and that's what I've done previously, and recommended others to do. However, this doesn't work if cursors point to scalar values. If I have a cursor that points to an integer, I'd get passed an integer as argument to the update functions. This is part of the "problem". By having a update function that is passed a cursor, you can compose without having to know if the cursor points to a scalar or not.

I get your scepticism and would be vary of feature creep myself, but I think the ability of passing cursors to your actions can allow you to have a better internal API. Therefor a method like this could help to make actions much more composable.

aknuds1:
I agree with Mikael! The uniformity of cursors is much preferable to have to deal with either immutables or scalars.

lacker:
Thanks for sending in this pull request. This has been stalled for a while so I think it is time to just close this pull request. It might be that this feature is useful, but it seems like we would need a more concise case for it to make it core-library-worthy.

PR #561, opened 2015-07-24
Correct location of isIterable function

matthewwithanm:
Just a small docs fix. The `isIterable` function is on `Iterable` (not `Immutable`).


leebyron:
Merged in 825e29ef98012d210f817905994529be09c45de2, thanks!

PR #567, opened 2015-07-29
Typescript: export Immutable as a module

jmalonzo:
This is helpful in cases where immutable.js was imported via a script tag, as "immutable" will not
be available in that environment (but Immutable is).


jmalonzo:
@leebyron I've updated the comment and squashed the commits. Cheers.

leebyron:
Thanks!

leebyron:
For future reference, once you squash your commits, you can just run `git push -f` to update the same PR without having to make a new one.

jmalonzo:
@leebyron thanks, will do!

PR #571, opened 2015-07-30
change findIndex in Lists default value

ddaza:
FindIndex should return something besides `-1` specially since the `get(-1)` gets the last value of the list. It just makes it confusing and not intuitive.

### DIFF
--- a/__tests__/List.ts
+++ b/__tests__/List.ts
@@ -398,7 +398,7 @@ describe('List', () => {
   it('finds values using findIndex', () => {
     var v = List.of('a', 'b', 'c', 'B', 'a');
     expect(v.findIndex(value => value.toUpperCase() === value)).toBe(3);
-    expect(v.findIndex(value => value.length > 1)).toBe(-1);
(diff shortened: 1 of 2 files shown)


csaintc:
I agree with this, once the tests are fixed. It is strange that I can't use the int returned from findIndex to get a provided default value from a list.

```
const fridge = new List(['bacon', 'eggs', 'cheese']);
const index = fridge.findIndex((food) => {
  return food === 'vegetables';
});
fridge.get(index, 'Food not in fridge.')
```

outputs: `cheese`

lacker:
I think this is intuitive because the regular find-index code in javascript does exactly this. Also, it would be a pretty big breaking change, so it would have to be quite valuable before it was worthwhile. I think that means the conclusion here is "wontfix", so I am going to close this pull request. Thank you for submitting it though - I really appreciate the spirit and the effort here.

ddaza:
Completely forgot about this! Thanks for taking the time to respond.

After giving it some thought, I agree with the `findIndex()` response being the same as the JS response.  I think the confusion came when using it with the `get` we need to be careful combining the 2 functions.

Issue #588, opened 2015-08-14
Immutable.Map().get() does not treat integers the same as {}[]


jlucas91:
Hi all,

I'm fairly new to JS so I might be missing something, but it seems like Immutable.Map().get() should treat integers the same way as standard object access. For example:

> Immutable.Map({1:2}).get(1)
undefined

> var x ={1:2}
undefined
> x[1]
2

You need to convert to a string to use .get() which seems error prone and cumbersome.
> Immutable.Map({1:2}).get('1')
2

leebyron:
The issue here is not in Immutable, but JS objects. JS object properties are *always* strings.

```js
var x = { 1: 2 }; // { "1" : 2 }
var keys = Object.keys(x); // [ "1" ]
```

When you access a property using square-bracket notation, that value first gets coerced into a string before looking it up in the JS Object.

leebyron:
Map, on the other hand, does not coerce to string first. Maps (ES6 Maps and Immutable.js Maps) allow you to store any kind of value as keys, strings, numbers, other Objects.

jlucas91:
Ah, that makes sense. Thank you for the clarification!

samwgoldman:
Just from watching this repo, I feel like this confusion is pretty common. I looked to see if there was anything in the docs calling it out as a common pitfall, but couldn't find anything. I'd be happy to contribute the documentation, but I'm not sure where it should go. The README? The docs for the [Immutable.Map constructor](http://facebook.github.io/immutable-js/docs/#/Map/Map)?

leebyron:
I would love to entertain that kind of change to the docs!

alvaromb:
Sorry to re-open this issue, but I have a question. If we fetch a JSON from an API and create a ``Map`` from it, all keys will be converted to string, am I right? There is no easy way to automatically parse a JSON to an Immutable structure without having all object keys as strings really?

leebyron:
That's mostly right. I should point out that "all keys will be converted to string, am I right?" is not quite right - specifically there is no conversion happening as JSON object keys are *always* strings.

For example:

```js
// Examples of valid JSON
{ "foo" : "bar" }
{ "true" : true }
{ "123" : 456 }

// Examples of invalid JSON
{ foo : "bar" }
{ true : true }
{ 123 : 456 }
```

leebyron:
Added 59028b5e8552339acbdaab0c3f6547bd131f8f54 to help clarify further

alvaromb:
[...]

PR #589, opened 2015-08-15
Document common pitfall for Maps constructed of JS objects

samwgoldman:
Ref #588

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -404,6 +404,19 @@ declare module Immutable {
    *     var newMap = Map({key: "value"});
    *     var newMap = Map([["key", "value"]]);
    *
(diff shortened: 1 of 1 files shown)


samwgoldman:
(I wonder if this confusion is why the ES6 Map doesn't have this kind of constructor.)

Dashed:
@samwgoldman Can you also put a similar warning for `fromJS()`?

leebyron:
Thank you for writing this! I added it to fromJS as well in ecae145be82a137f1b7f2d2fad67ad90e9c65ee1

> (I wonder if this confusion is why the ES6 Map doesn't have this kind of constructor.)

It definitely is one of the primary reasons that it does not. I think it's also one of the reasons it's not often used :)

PR #615, opened 2015-09-08
Add set difference method to Map Fixes #601

### LINKED ISSUE #601: Set difference on Maps
It would be very useful for me to have set difference operations on Maps as well as sets. For example something like

``` javascript
const map = Immutable.Map({a: 1, b: 2, c: 3});
const set = Immutable.Set([a, b]);
console.log(map.subtract(set)).toJS(); // prints {c: 3}
```

rawls238:
@leebyron @derekdreery

### DIFF
--- a/__tests__/Seq.ts
+++ b/__tests__/Seq.ts
@@ -97,4 +97,9 @@ describe('Seq', () => {
     expect(dig.toJS()).toEqual(['d', 'i', 'g']);
   });
 
+  it('Does not infinite loop when an undefined number is passed to take', () => {
+    var list = Immutable.Seq([1, 2, 3, 4, 5]);
+    expect(list.take(NaN).toJS()).toEqual([]);
+  });
+
 });


mpuyat:
@rawls238 In the case that you would do the difference across 2 maps, such as:

```
const mapOne = new Immutable.Map({a: 1, b: 2});
const mapTwo = new Immutable.Map({a: 1});
console.log(mapOne.subtract(mapTwo));  // {a: 1, b: 2}
console.log(mapOne.subtract({1: 'a'}));  // {b: 2}
```

It seems like the behavior is sort of unexpected because you're using the `includes` method, which will use the values of `mapTwo` to perform the "difference."

@leebyron care to shed some light on what the semantics for comparison across 2 maps should be? Seems like the intuitive thing to do is use the keys.

rawls238:
My original intention was to have the key based. It's an oversight in the implementation if it is not, I think that this kind of differencing only makes sense over keys and we should exclude values. If we really wanted to support difference over values I think that that behavior should be specified by the consumer using a separate flag or something.

Issue #605, opened 2015-08-27
feature request:  merge deep and delete on a Immutable.Map (intersection of keys)


mercuriete:
as another guy states in that bug..
https://github.com/facebook/immutable-js/issues/455

i think it would be nice to have this feature.

for example, i am creating a web with ReactJS and i am trying to show a list of items.

i need to quick short cut the rendering if i know that the item doesnt change at all.

Of course i could mergeDeep and then delete the keys that are not anymore there but i think we should have a simple operation as intersection is.

thanks for develop ImmutableJS :).

clouddra:
+1 for key intersection of Map

leebyron:
Merge is another way of saying "union" and a union definitionally never deletes anything. Does `deleteIn` suffice? Perhaps you could provide some code example of what you wish you could write which doesn't work at present?

frank-weindel:
This would be very useful to me right now. I'm working on a settings modal that uses one immutable map for the current active settings and another one that is the pending settings. If the active settings and the pending settings are the same (via Immutable.is), no changes have been made. When the user goes to save their changes I'd like to do this kind of deepSet between the active settings and the pending settings. The pending settings get priority, any keys that exist in active but are missing in pending get removed from active. For all the other keys it works like a normal merge.

crunkyedwards:
+1 for this feature. I have exactly the same scenario as frank-weindel.

mercuriete:
I will try to explain my scenario.
supose you have a REST endpoint that returns N items. And this list changes all the time very quickly.
So you have a lot of item that was previously in the map but  some dissapear and some are new.

I wanted to have the intersection of both sets/maps plus the new ones.

So in the end what i wanted it wasn't an intersection.

I wanted something like mergeDeep and delete those items that aren't there anymore.

si-harps:
This feature would be very useful. I currently have 2 similar structures stored in ordered maps, let call them schema and document. 

I need to be able to deep merge the document with all its values into schema but delete all keys (nested or otherwise) that don't exist in schema.

Currently I use mergeWith and delete keys recursively down the tree but a new member of orderedMap (eg. intersect) would be awesome!

I did consider using orderedSet for this but it doesn't match my needs entirely. By the way, thanks so much for immutable.js - incredible library!

lukaswelinder:
This would definitely be a nice feature to have. One way this can be done:

```javascript
import { Map, Set } = from 'immutable';

// Shared properties, expected result:
[...]

PR #616, opened 2015-09-08
Add subtract function for Maps

marcelpuyat:
Refs: https://github.com/facebook/immutable-js/issues/601 https://github.com/facebook/immutable-js/issues/605 https://github.com/facebook/immutable-js/pull/615/


rawls238:
I think that it's quite strange that you put up a separate PR for this, but I believe that this implementation is better than the one I had :)

lacker:
Hey, so there are basically two ways to implement a "diff" type function. You can just do it with an iterable over the whole structure, or you can get into the structure itself. The first way is simpler, and it's what this PR does, but people can do it themselves already based on the existing API. The second way actually gives some performance improvement. So I don't think we should go this direction, I think we should either not include "diff" in the API at all, or have an optimally-efficient mechanism. Thanks for submitting this PR though, I really appreciate the effort here.

Issue #569, opened 2015-07-29
Surprising result for Iterable#includes


myitcv:
With reference to: https://facebook.github.io/immutable-js/docs/#/List/includes

Consider:

``` javascript
var Immutable = require("immutable");

var o1 = Immutable.fromJS({x: 5})
var o2 = Immutable.fromJS({x: 5})

console.log("o1 === o2", o1 === o2);

var l = new Immutable.List([o1]);

console.log("l.includes(o1)", l.includes(o1));
console.log("l.includes(o2)", l.includes(o2));
```

which gives the output:

```
o1 === o2 false
l.includes(o1) true
l.includes(o2) true
```

The last line surprised me because my assumption was that every function in the API would, by default, use reference equality tests as opposed to value equality (with the obvious exception of [`is(...)`](https://facebook.github.io/immutable-js/docs/#/is)). 

`.includes` is in effect returning `true` iff `Immutable.is(e, o2)` for any `e` in `l`. 

Have I missed an important part of the docs here?

leebyron:
Closing this aging issue - sorry for not getting to your question

`includes()` does indeed use `Immutable.is` which implements value equality for immutable collections

PR #617, opened 2015-09-09
Document usage of Immutable.is() in docs for Iterable.includes and Iterable.has

marcelpuyat:
Refs: https://github.com/facebook/immutable-js/issues/569


leebyron:
Great, thank you

Issue #619, opened 2015-09-10
Symbols as entries in Map throw "Cannot convert a Symbol value to a string"


dashed:
Running in io.js:

```js
const Immutable = require('immutable');

// Symbols as entries in Map throw:
// TypeError: Cannot convert a Symbol value to a string

Immutable.fromJS({foo: Symbol('bar')});
Immutable.Map().set('foo', Symbol('bar'));

// these examples don't throw
Immutable.Set([Symbol('foo')]);
Immutable.fromJS(['foo', Symbol('bar')]);
Immutable.List([Symbol('foo')]);
Immutable.Set().add(Symbol('hi'));
Immutable.Map().set(Symbol('bar'), 'foo');
```

PR #620, opened 2015-09-10
Prefer explicit string conversion

dashed:
ES6 Symbols will throw a runtime exception if implicitly coerced into a string.

----

Closes https://github.com/facebook/immutable-js/issues/619

-----

Unsure if this is the right patch.


leebyron:
Thanks! Sorry for letting this sit - did you verify that this fixes the issue for you?

Dashed:
Yep verified. I probably should've put in a test for it.

leebyron:
I'd like to test this. Could you include a test or use case that fails before this patch and works after?

Dashed:
@leebyron I wasn't quite sure how to test this such that:

```js
Symbol() + ''; // this throws
String(Symbol()); // but this doesn't throw
```

I added a test case that I was working from; but didn't get far. I added a polyfill, core-js, which may be undesirable. 

----

I think an alternative approach to testing would be to mock the String constructor using Jest; but I'm not quite sure how to do this.

puradox:
What's the status of this PR? I'm running into the same problem and would like to see this merged.

leebyron:
Tests look good, just need to rebase and merge. Will get to that shortly.

dashed:
This needs more work. I won't be able to look into it for the next few weeks...

puradox:
@dashed Can you explain what needs more work? I can lend a hand :+1:

dashed:
@puradox https://github.com/facebook/immutable-js/pull/620#issuecomment-166125273

dashed:
Yay! Thanks @leebyron ! 🎉

runru:
Recover my Gmail account reginaladnier50@gmail.com

Issue #618, opened 2015-09-09
Sub-cursor updates are concurrent


jameshopkins:
``` js
const data = Immutable.fromJS({a: 1, b: 2});

const cursor = Cursor.from(data, newData => {
  console.log(newData)
});

cursor.set('a', 2); // Map {"a": 2, "b": 2}
cursor.set('b', 3); // Map {"a": 1, "b": 3}
```

The second `.set()` call disregards the value which the first `.set()` call sets. I require the second `.set()` to return `Map {"a": 2, "b": 3}`
## Context

Two sub cursors (which don't share the same parent cursor), are passed to a React component as two individual `props`. I wish to update the values of both sub cursors in the same function call within that React component.

jurassix:
In older versions of Immutable you needed to actually redefine the cursor on each change.

Try something like:
```javascript
const cursor = Cursor.from(onTransaction);
function  onTransaction(newData, oldData) {
      cursor = Cursor.from(newData, onTransaction);
}
```

jurassix:
[From the source it looks like you just use `withMutations`](https://github.com/facebook/immutable-js/blob/cursor/__tests__/Cursor.ts):

```javascript
  it('can have mutations apply with a single callback', () => {
    var onChange = jest.genMockFunction();
    var data = Immutable.fromJS({'a': 1});

    var c1 = data.cursor(onChange);
    var c2 = c1.withMutations(m => m.set('b', 2).set('c', 3).set('d', 4));

    expect(c1.toObject()).toEqual({'a': 1});
    expect(c2.toObject()).toEqual({'a': 1, 'b': 2, 'c': 3, 'd': 4});
    expect(onChange.mock.calls.length).toBe(1);
  });

```

charypar:
I think the problem is more like this, in code:

```
const data = Immutable.fromJS({a: 1, b: 2});

const cursor = Cursor.from(data, newData => {
  console.log(newData)
});

const cursor1 = cursor.get('a');
const cursor2 = cursor.get('b');

// pass cursor1 and cursor2 to different places, then

cursor1.set('a', 2); // Map {"a": 2, "b": 2}
cursor2.set('b', 3); // Map {"a": 1, "b": 3}
```

The cursors are independent, changing one has no effect on the other, which makes them less useful than they could be. You could say they are not "live" - once you derive a cursor from a parent, changes to the parent (because they need to overwrite the parent), don't affect the child any more.

[...]

PR #622, opened 2015-09-11
Allow for access to synchronised root cursor data across derived cursors

jameshopkins:
Currently, in order to give a derived cursor access to up-to-date root-level data, it must be adjoined to the root-level cursor via a chained sequence. This characteristic is severely limiting when used in applications where sub-cursors are distinctly independent, and therefore establishing a chain in this way isn't feasible, nor appropriate.

A typical example of this is in a React application where multiple sibling sub-cursors are passed into a component as `props`. The [issue](https://github.com/facebook/immutable-js/issues/618) was [nicely summarised](https://github.com/facebook/immutable-js/issues/618#issuecomment-138918319) by @charypar recently. Issues include:
* reliance on the component's render cycle in order to propagate root data updates to cursors.
* skipping a subtree in `shouldComponentUpdate` will result in the subtree having out-of-date cursors

## Solution
Instead of each cursor holding it's own reference to the root data, every cursor now shares a common reference to the root data.

### DIFF
--- a/contrib/cursor/__tests__/Cursor.ts
+++ b/contrib/cursor/__tests__/Cursor.ts
@@ -103,12 +103,13 @@ describe('Cursor', () => {
     expect(data.toJS()).toEqual(json);
 
     // as is the original cursor.
-    expect(deepCursor.deref()).toBe(1);
+    expect(deepCursor.deref()).toBe(3);
     var otherNewDeepCursor = deepCursor.update(x => x + 10);
-    expect(otherNewDeepCursor.deref()).toBe(11);
+    expect(otherNewDeepCursor.deref()).toBe(13);
+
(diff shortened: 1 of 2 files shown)


jameshopkins:
Given the apparent [enthusiasm of the authors for others to extend the cursor outside of Immutable.js](https://github.com/facebook/immutable-js/wiki/Upgrading-to-Immutable-v3#cursors-in-contrib), and due to the integral change in semantics that my PR would introduce, I've decided to fork the cursor implementation as it's own [project](https://github.com/redbadger/immutable-cursor) in the hope it can be extended further, independently of future Immutability.js releases. I will however, be actively tracking this project, and fixing any issues/including additions as and when they come about.

PR #623, opened 2015-09-13
throw error if NaN is passed into take Fixes #552

### LINKED ISSUE #552: Seq.take() without arguments causes `Maximum call stack size exceeded` exception.
Test case - http://jsfiddle.net/1ccutco5/1/
I think take should throw exception with info about usage, or return 1 element(i issued this with takeLast() and, for me, this name sounds like function already should return last element without any arguments).

rawls238:
@leebyron Fixes #552

### DIFF
--- a/__tests__/Seq.ts
+++ b/__tests__/Seq.ts
@@ -97,4 +97,9 @@ describe('Seq', () => {
     expect(dig.toJS()).toEqual(['d', 'i', 'g']);
   });
 
+  it('Does not infinite loop when an undefined number is passed to take', () => {
+    var list = Immutable.Seq([1, 2, 3, 4, 5]);
+    expect(list.take(NaN).toJS()).toEqual([]);
+  });
+
 });


leebyron:
Hey, thanks for this PR! A few pieces of feedback:

* Immutable.js strives to match existing JavaScript behavior. Since `take(N)` is shorthand for `slice(0, N)`, the behavior should match. Instead of throwing an error, `slice(0, NaN)` always returns an empty collection.
* The change to the type definition file seems incorrect. You should always provide a number to take, a type-checker like TypeScript should catch if you have not - but your change would prevent it from doing so.
* Can you add a test case illustrating the expected outcome of code reproing #552 to ensure this doesnt regress? In fact, I would suggest you do this first, as I suspect the issue may have been resolved in the last release on npm.

rawls238:
Aha, it is indeed fixed for the case where NaN or null is passed through in the most recent version. However, it still breaks for the case where take() with no args is called because of the lack of the relevant definition in the typescript (which was what was reported in the issue), though we could simply say that we don't support this kind of behavior since it's ambiguous as to what the take function is supposed to do?

leebyron:
Yeah I think the type definition should continue to enforce that `take(N)` accept one argument. There are already functions called `first()` and `last()` which accepts no arguments that's designed to be used for returning one element and have clearer names than `take()` for that behavior.

leebyron:
In that case, it's up to you on what to do next. I would be happy to accept a PR which adds a regression test case illustrating proper behavior for `take()` and `take(NaN)` if you're up to it.

rawls238:
I think it makes sense to close this and the corresponding issue since I can't write the test case for the take() case since the jestPreprocessor fails before the test can execute :). Would be happy to write a simple test case for the NaN case if you think that's worth adding

leebyron:
I do think the NaN case is worth adding :) - no pressure if you don't want to. But I would be happy to add that test

rawls238:
awesome, added the appropriate test case

leebyron:
Great, thank you!

PR #630, opened 2015-09-20
WIP: Dash docset

glenjamin:
I'm a big https://kapeli.com/dash user, and I've been meaning to do some work to generate a docset for immutable for a while.

I've got something working now, so figure I'd share it. I'm unsure whether it makes sense for this to sit inside the main repository or not going forwards.


glenjamin:
Here's a screenshot of the result in action, for reference

screen shot 2015-09-20 at 21 52 08

PR #644, opened 2015-10-02
Updated immutable to the latest version

jv-PintoBobcat:
People like to open the console and tinker with the documentation right there.  Might as well give them the latest version.


rawls238:
:+1: to this, seems like a lot of issues are being filed from this

leebyron:
Thanks

PR #646, opened 2015-10-02
add insert function on List Fixes #402

### LINKED ISSUE #402: Feature request: insert()
For convenience: `insert()` would be nice. Basically just a convenience wrapped around `slice()`, similar to `remove()`.

rawls238:
@leebyron


dist/immutable.d.ts
```diff
@@ -180,6 +180,7 @@ declare module Immutable {
      */
     delete(index: number): List<T>;
     remove(index: number): List<T>;
+    insert(index: number, value: T): List<T>;
```
leebyron: Can you add a docblock above this similar to the one above `delete` which describes the method in similar language to other documentation in this file?

rawls238: yup, will do

rawls238: added

leebyron:
Nice work, thanks

Issue #592, opened 2015-08-19
Inconsistency between .d.ts and implementation?


myitcv:
`v3.7.4`

The following TypeScript code compiles...

``` typescript
import * as Immutable from "immutable";
...
console.log(Immutable.IndexedSeq<number>([1,2,3]));
```

...but fails at runtime:

```
Uncaught Error: Immutable.IndexedSeq is not a function
```

The following TypeScript code does not compile...

``` typescript
// Property 'Indexed' does not exist on type 'typeof Seq'
console.log(Immutable.Seq.Indexed<number>([1,2,3])); 
```

...but works at runtime:

```
ArraySeq{_array: [1, 2, 3], size: 3}
```

Suggests the `.d.ts` definition is incorrect to me?

matthewwithanm:
@myitcv Can you give #647 a try?

leebyron:
Fixed in master, will be released soon!

PR #647, opened 2015-10-03
Update type definitions to match nested Indexed, Keyed, Set variants

matthewwithanm:
Re: #592

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -1192,6 +1192,33 @@ declare module Immutable {
      */
     function of<T>(...values: T[]): Seq.Indexed<T>;
 
+
+    /**
+     * `Seq` which represents key-value pairs.
+     */
+    export module Keyed {}
+
+    /**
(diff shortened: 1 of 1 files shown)


myitcv:
@matthewwithanm - what does the resulting documentation look like (not sure how to regenerate)?

matthewwithanm:
Good question! I hadn't checked /:

The doc generation is in the gh-pages branch and currently chokes on the new file…I've opened #648 to teach the visitor about qualified names. After that, these things look pretty much like you might expect (or at least pretty much like I expect):

# Nav

screen shot 2015-10-03 at 9 16 10 am

# Content

screen shot 2015-10-03 at 9 17 20 am

myitcv:
I'll defer to @leebyron - just so long as the docs (whatever they look like) match the implementation I'm happy :+1:

leebyron:
This is great, thank you for doing this.

matthewwithanm:
😊 np

PR #648, opened 2015-10-03
Handle QualifiedNames

matthewwithanm:
QualifiedName nodes don't have a `text()` method.

This change (or a similar one) is needed to generate the docs after #647. lmk if there's a better way to implement it!


leebyron:
Nice! Thanks for this

Issue #632, opened 2015-09-24
Set.contains is slow


fab1an:
Using `contains` to search in a set is incredibly slow, the time is spent in the `Iterable.is` comparison method.

This is irritating since `contains` is referenced as alias to `includes` in the documentation.

A benchmark:
https://jsperf.com/immutabe-set-vs-normal-object/14

![screen shot 2015-09-24 at 12 20 41](https://cloud.githubusercontent.com/assets/55547/10071006/b0bd5190-62b6-11e5-90f0-bede5d6654d4.png)

jv-PintoBobcat:
That is really strange.  Here's the code for it: https://github.com/facebook/immutable-js/blob/9a5897ec0bd623b9e41e0015c75e5f7a46a3494e/src/IterableImpl.js#L459

fab1an:
The difference is that `contains` uses the `Iterable` mixin: https://github.com/facebook/immutable-js/blob/9a5897ec0bd623b9e41e0015c75e5f7a46a3494e/src/IterableImpl.js#L147

and `includes` uses the `SetIterable`mixin: https://github.com/facebook/immutable-js/blob/9a5897ec0bd623b9e41e0015c75e5f7a46a3494e/src/IterableImpl.js#L678

leebyron:
Thanks for the report! You're right that the alias is getting the wrong method in this case.

fab1an:
I tried adding a Unit-Test, but I don't really understand how the mixins work. How would such a test be written? (that calling `.has` equals calling `.contains` etc.)

tomwidmer:
I guess you could 'spy' on .has using a jasmine spy and check that .contains calls it. Something like:

    var spy = spyOn(mySet, 'has');
    expect(mySet.contains('foo')).toBe(true);
    expect(mySet.has).toHaveBeenCalled();

or something.

PR #660, opened 2015-10-16
Fixes Set#contains performance bug

danielstreit:
This is a fix for #632 A performance bug with Set#contains

`contains` is initially set to `includes` on the `Iterable.prototype` [here](https://github.com/facebook/immutable-js/blob/master/src/IterableImpl.js#L459). SetIterable.prototype overrides `includes` [here](https://github.com/facebook/immutable-js/blob/master/src/IterableImpl.js#L678). But, `SetIterable` does not implement `contains`, it still points to the super. This pr fixes that.

http://jsperf.com/immutabe-set-vs-normal-object/18

screen shot 2015-10-15 at 10 32 13 pm


leebyron:
Awesome, thank you for this!

PR #665, opened 2015-10-19
Fix typo in comment describing List#forEach

bzalasky:
Changes "regardless of if they WHERE explicitly defined," to "regardless of if they WERE explicitly defined."


leebyron:
Thanks!

Issue #505, opened 2015-06-18
Remove homogeneous collection API from records


leebyron:
Records should not have the map/filter/reduce and other API methods useful for homogeneous collections like List and Map. Their operation does not do what is expected.
#220 #268

negativetwelve:
I just ran into this issue and spent a while trying to debug it. Is there any way we can add this to the docs so that future developers don't spend a while debugging this issue? I'd be more than happy to make that change

andrinealver:
:+1:

anjianshi:
I found all methods that called `recordType.constroctor()` has problem, includes `merge` (actully `mergeIntoCollectionWith`).

```js
var R = Immutable.Record({});
var r = new R();
r.merge({});   // no problem, return "r" itself
r.merge({a: 1});  // return undefined
```

AlexGalays:
Why remove them altogether though ? Record.map seems very useful for instance.

leebyron:
Note when tackling this, ensure `isImmutable()` still returns true for record instances

PR #668, opened 2015-10-21
Make it obvious that a Record is a Map

johnnyreilly:
and so shares the API of Map.  I only discovered this by accident so it seems potentially worth emphasing?

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -1077,7 +1077,7 @@ declare module Immutable {
   /**
    * Creates a new Class which produces Record instances. A record is similar to
    * a JS object, but enforce a specific set of allowed string keys, and have
-   * default values.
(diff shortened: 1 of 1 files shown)


rawls238:
Given this issue https://github.com/facebook/immutable-js/issues/505 I don't think we want to emphasize this since it should not share an identical API as Map

johnnyreilly:
That's a fair point. That said, the docs don't presently detail how one changes a record (using set, update etc) at all. That should be corrected and guidance given I think.

ir-fuel:
Johnny has a point. Right now the docs only show definition, construction and field removal. I wasn't aware of the other methods until I stumbled upon this issue.

leebyron:
This is actually just an implementation detail, and one that's likely to change in the future. It's a "Map" in the sense that it exposes `get` and `set` methods, but the HOF methods like `filter` and `forEach` make no sense on a Record.

PR #682, opened 2015-10-29
Add subtract method to Immutable.Map

mattbanister:
I'm setting this PR up for this issue: https://github.com/facebook/immutable-js/issues/601

I'm having a problem with the default grunt task though.  There's one test failing for me in master with the following message:

```
Waiting on 1 test.../Users/matt/Documents/Projects/immutable-js/__tests__/zip.ts 97:14 Module 'Immutable' has no exported member 'IndexedIterable'.
 FAIL  __tests__/zip.ts
```

Any idea why that might be the case?

Also I understand that I should add the artifacts to this PR.  Does that include everything from the dist folder?  immutable.d.ts, immutable.js and immutable.min.js?   I just wanted to confirm and I'll update the PR.

Thanks
Matt


mattbanister:
Okay looks like that might have been a problem with the "infinite lists" test in the zip tests file.  I added an additional commit that fixes the build task.

lacker:
I think this method of taking the difference is too inefficient - a user of this library could just do it themselves if you are removing things one by one in `withMutations`. So if this method were included in the library people would expect it to behave more optimally. This PR is also pretty old, so I am just going to close it. If you are interested in working on this more then please feel free to reopen!

PR #685, opened 2015-11-03
Fix Record documentation.

lucsky:
The example showing how undefined values in a Record are ignored was incorrect, it was actually using a defined value.


leebyron:
The documentation was actually already correct as written, but I agree it was not very clear about *why* this is the case. I'll make a change which better describes why the key "x" is going to be ignored.

lucsky:
Aaaahh, yeah, I got it now. Sorry for the noise.

PR #709, opened 2015-11-29
Remove homogeneous collection methods from Record Fixes #505

### LINKED ISSUE #505: Remove homogeneous collection API from records

rawls238:
@leebyron this seems to have caused a bunch of confusion, there are numerous issues people have filed around confusion about strange behavior when using these methods on Record.


dist/immutable.js
```diff
@@ -3799,6 +3799,8 @@
     return EMPTY_ORDERED_SET || (EMPTY_ORDERED_SET = makeOrderedSet(emptyOrderedMap()));
   }
 
+  var UNSUPPORTED_OPS = ['map', 'reduce', 'filter', 'sort', 'groupBy', 'sortBy', 'reduceRight', 'every', 'some'];
```
rawls238: let me know if I've left anything out here or you think that some of these should be supported

dist/immutable.js
```diff
@@ -3922,6 +3924,9 @@
   RecordPrototype.withMutations = MapPrototype.withMutations;
   RecordPrototype.asMutable = MapPrototype.asMutable;
   RecordPrototype.asImmutable = MapPrototype.asImmutable;
+  UNSUPPORTED_OPS.forEach(function(op)  {
+    RecordPrototype[op] = undefined;
```
rawls238: I wonder if it may make sense to fail loudly with these for now (i.e. explicitly throw an error stating the operation is unsupported) as opposed to silently removing them by simply setting them as undefined

rawls238:
ping @leebyron

leebyron:
Thanks for this (and for your patience) I'm working out a roadmap for immutable now that includes this idea and want to ensure we don't introduce something breaking along the way, so lend me some time to work things out. Thanks so much for your help addressing this!

lacker:
This pull request is super old and I think @leebyron has some master plan here anyway. Since it's probably not even worth rebasing I think we should just close this pull request. Thank you for submitting it, though, I really appreciate the effort.

PR #717, opened 2015-12-16
Using `typings` in `package.json`

xogeny:
This pull request shows what is required to make `immutable` easily usable with TypeScript >= 1.6.

The key thing is adding the `typings` field to `package.json` and pointing it at the `index.d.ts` file.  This `index.d.ts` file is different from `immutable.d.ts` because it *does not* have a `module` wrapping it.  The `module` part is apparently implied when using the `typings` field.

A few notes.  First, I'm not necessarily advocating for accepting *this* pull request.  I just wanted to show what was necessary for me to get it working.  With this simple addition to `immutable`, you can just do this in TypeScript:

```
import immutable = require('immutable')
```

After doing an `npm install immutable`.  There is no need to use `tsd` or `DefinitelyTyped` at all.

Another thing to note is that I've copied and pasted most of `immutable.d.ts` to create `index.d.ts`.  I recognize this is not very DRY.  I just couldn't figure out a quick and simple way to use a `<reference.../>` element to avoid this.  But I'm pretty sure it could be done.

### DIFF
--- a/dist/immutable.d.ts
+++ b/dist/immutable.d.ts
@@ -1126,7 +1126,7 @@ declare module Immutable {
    *
    */
   export module Record {
-    interface Class {
+    export interface Class {
       new (): Map<string, any>;
       new (values: {[key: string]: any}): Map<string, any>;
       new (values: Iterable<string, any>): Map<string, any>; // deprecated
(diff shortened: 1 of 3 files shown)


leebyron:
Closing since related changes have landed, but thanks for your advocation!

PR #725, opened 2015-12-19
Inheritance 

andykog:
It would be nice to have ability of creating extendable immutable classes, so we can make things like:
```js
function ExtendableMap(d) {
   Object.assign(this, Immutable.Map(d));
}
ExtendableMap.prototype = Object.create(Immutable.Map.prototype, {
   constructor: { value: ExtendableMap, enumerable: false, writable: true, configurable: true }
});

class Form extends ExtendableMap {
   constructor(initialValue = "") {
      super({ value: initialValue });
   }
   setValue(z) {
      return this.set("value", z);
   }
}

class SpecialValidatedForm extends Form {
   setValue(z) {
      return 
         super.setValue(z)
         .set("valid", z != "");
   }
}
```
Performance loss doesn't seem to be considerable.
I can make additional commits, if that makes sense.


leebyron:
Thanks for your investigation into this!

We've done some pretty thorough investigation into this idea a couple times now, the most in depth by @tgriesser and found some pretty serious performance burden by this, regardless of whether you actually use the inheritance feature. Ultimately we decided not to support this, and philosophically prefer a functional style or at least composition over inheritance anyhow.

Here's how your same example might work using a functional style:

```js
function setFormValue(map, z) {
  return map.set("value", z);
}

function setSpecialValidatedFormValue(map, z) {
  return setFormValue(z).set("valid", z != "");
}
```

Here's how your same example might work using object composition:

```js
class Form {
   constructor(initialValue = "") {
      this._map = Map({ value: initialValue });
   }
   _create(newMap) {
      if (newMap === this._map) { return this; }
      var newForm = new this.constructor();
      newForm._map = newMap;
      return newForm;
   }
   setValue(z) {
      return this._update(this._map.set("value", z));
   }
}

class SpecialValidatedForm extends Form {
   setValue(z) {
      return this._update(
         super.setValue(z)._map.set("valid", z != "")
      );
   }
}
```

PR #726, opened 2015-12-20
Add explicit dependency on jest-cli, to keep npm3 happy

glenjamin:
Without this, `jest-cli` is not installed when using `npm@3`, and you can't run the tests.


leebyron:
Awesome thanks! I just rebased and landed this

PR #727, opened 2015-12-20
Short-hand constructor for Map, taking flat arguments

glenjamin:
The typescript part of this is a bit suspect, but it compiles!

When constructing a map I'd quite like to create via an API like:
```js
Immutable.Map.of('key', 'val1', 'key2', 'val2');
```

This is to reduce the visual and allocation overhead introduced when using array-of-arrays - which as far as I can tell is the only way to construct succinctly with non-string keys.


leebyron:
This is pretty cool, but I'm concerned about the type signature. Unfortunately this would only allow making Maps where the keys and values are of the same type.

Maybe it would be better to just explicitly type this as `Map<any, any>`?

glenjamin:
That could be simpler, I'll make that change now.

glenjamin:
Type signature changed, had a test failing locally - but twas also failing on master so I think it's unrelated to these changes.

leebyron:
I'll look into the test. It seems to pass in newer versions of node which is suspect

PR #728, opened 2015-12-21
Removed Object.observe reference from readme.

maxehnert:
Removed the reference to `Object.observe` from the readme. This method is set to be removed [1][2] and will not make for a clear reference. 

gh-pages branch should also be updated to reflect the same change to the `index.html` file, however I am unsure how pull requests work for the gh-pages branch. 

[1] https://esdiscuss.org/topic/an-update-on-object-observe
[2] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/observe

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,7 @@ Much of what makes application development difficult is tracking mutation and
 maintaining state. Developing with immutable data encourages you to think
 differently about how data flows through your application.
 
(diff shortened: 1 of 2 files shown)


leebyron:
Agreed, thanks!

PR #743, opened 2016-01-06
Ensure tests script fails if any tasks fail

glenjamin:
Without this, failing tests would not produce a non-zero exit code.


leebyron:
Most excellent

PR #753, opened 2016-01-10
Cursor: adding function to be able to easily identify an object as a cursor

richardanaya:
Exposing function for easily identifying if an object is a cursor. This was necessary for me to identify props that were cursors passed down through react elements to determine if I should update based off differences of internal state.

### DIFF
--- a/contrib/cursor/index.d.ts
+++ b/contrib/cursor/index.d.ts
@@ -55,6 +55,9 @@ declare module 'immutable/contrib/cursor' {
     key: any,
     onChange?: (newValue: any, oldValue?: any, keyPath?: Array<any>) => any
   ): Cursor;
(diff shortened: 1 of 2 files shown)


richardanaya:
Signed CLA

lacker:
This seems pretty reasonable to me. Would you mind adding some tests and also adding the Flow types in addition to the typescript ones?

lacker:
Actually - chatting with @leebyron it sounds like the ideal outcome would be to move Cursor out of this library entirely. So perhaps a better plan would be to stop trying to add functionality there and first move it out. So, I am going to close this pull request, but if you are interested in maintaining Cursor separately from the core immutable library then I am interested to chat!

PR #755, opened 2016-01-12
Minor grammar improvement: ‘…regardless of if…’ --> ‘…regardless of whether…’

gasi:
More appropriate according to http://languagelog.ldc.upenn.edu/nll/?p=4358.

Alternatively: ‘…regardless whether…’ :wink:


leebyron:
Thanks!

PR #759, opened 2016-01-13
Fixed Range.toString step

langpavel:
current behavior
```js
Immutable.Range(0, -10, 2).toString() // "Range [ 0...-10 ]"
Immutable.Range(0, 1, 0.2).toString() // "Range [ 0...1 ]"
```
fixed behavior
```js
Immutable.Range(0, -10, 2).toString() // "Range [ 0...-10 by -2 ]"
Immutable.Range(0, 1, 0.2).toString() // "Range [ 0...1 by 0.2 ]"
```


leebyron:
Great catch, thanks

Issue #432, opened 2015-04-08
Iterable#toObject shall use Object.defineProperty


XeCycle:
This does something weird:
`Map().set("__proto__", null).toJS()`

leebyron:
Yes, any Map key that means something specific in the context of an Object property will be dangerous to convert to a JS object.

facebook-github-bot-4:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

PR #763, opened 2016-01-18
Use Object.defineProperty in .toObject

XeCycle:
For #432.


leebyron:
Do you have any profiling that helps understand the performance implication of a change like this?

It seems like we also have the choice of saying just not to convert to plain ol JS objects if your map keys contain strings that overlap with JS object special keys.

This ultimately boils down to a tradeoff, so we need some information to make a choice

XeCycle:
Will work on that this weekend.

I think we also have a third option of providing a `.toObjectSafe()`; changing the current `.toObject` is, technically speaking, a breaking change, so another method may as well be preferred if a smooth transition is desired.

lacker:
It seems like this pull request has been abandoned so I am going to close it. Please feel free to reopen if you'd like to continue working on this!

PR #765, opened 2016-01-18
Make cursor typescript definition accessible from other ambient external modules

polkovnikov-ph:
`cursor/index.d.ts` definitions are inaccessible from other ambient external modules. This is needed, for example, to write definitions for `immstruct`.


polkovnikov-ph:
Any movement?

leebyron:
Great, thanks!

PR #769, opened 2016-01-24
Documentation: fix reference to evenSquares in Seq

ablakey:
evenSquares was referenced in an example but should be oddSquares.


leebyron:
Thank you!

PR #779, opened 2016-02-03
Add typings to package.json. Fixes #734

### LINKED ISSUE #734: Typescript compatible
add ```"typings":"./dist/immutable.d.ts"``` to `package.json` would allow `Typescript` to automatically import the definitions without implicitly reference it in each file uses it

jmalonzo:


blakeembrey:
@jmalonzo This would not work. The `.d.ts` file needs to be in external module definition format, which this is not.

Issue #742, opened 2016-01-05
Modernification Master Task


leebyron:
Immutable.js's tools are behind the times and it is limiting progress on some of the top line goals.

This is a master task that details all the steps required to modernize the repository.
- [x] move the source for gh-pages into master branch (#741)
- [x] automatically deploy gh-pages when master pages src changes (a71d9b60f0944535c2b05c1dafb926833132473d)
- [x] use latest jest for tests
- [x] use eslint for linting (#1101)
- [x] improve strictness of lint rules (#1101)
- [x] run tests against latest versions of node (a91b5e7)
- [x] update to the latest version of typescript (#745)
- [x] use buble for ES6 -> ES3 compilation (#1102)
- [ ] remove usage of gulp in favor of npm scripts? (1b4166952995a7ceb898c254dfdc3b219210b597)
- [x] remove usage of grunt in favor of npm scripts (1b4166952995a7ceb898c254dfdc3b219210b597) (#1102)
- [x] include flow type definition file
- [ ] no longer check in distributed files (breaking change)
- [ ] move contrib to separate repository (breaking change)
- [x] setup yarn (#1100)
- [x] setup prettier (#1101)
- <s>move documentation generator source of truth to flow type + babel.</s> - Note, updated to latest TypeScript instead.

umidbekkarimov:
I also suggest to add https://github.com/prettier/prettier and https://github.com/yarnpkg/yarn to the list

umidbekkarimov:
As I said in https://github.com/facebook/immutable-js/pull/1096#issuecomment-284227352

> And first problem.
> 
> According to the spec - es6 classes can not be called as a functions.
> 
> Few solutions:
> 
> Make a breaking change and disalow var map = Map() syntax in favor of var map = new Map().
> Refactor all es6 classes to Function.prototype style.
> Keep using declassify (but first - need to fix it for latest rollup).

leebyron:
Using `new` with collections is actually an anti pattern. They're designed to be used like the JS scalar types (Number, String) as just a callable function. 

That does mean the class syntax either needs to only ever be consumed by an off-spec transformer or it needs to be converted to use MapPrototype.method = function(){} or rely on Object.create's extra properties param or something like that

leebyron:
I think it's safe to call this task done. There's always more modernization to be done, but this hits the major objectives.

Thanks so much for your help, @umidbekkarimov!

umidbekkarimov:
[...]

PR #784, opened 2016-02-06
Add eslintrc file

thomasboyt:
This adds an eslintrc file, extended from the fbjs-scripts base eslintrc (which React and Jest also use). I disabled rules that were currently failing, and the `src/` directory lints with no errors.

It doesn't add any automated linting yet, but does allow editors configured for linting to lint individual files.

Related to #742.


leebyron:
Thanks Thomas!

This is a great start.

PR #787, opened 2016-02-10
fix broken link for immutable.min.js

raunakkathuria:
as of now link for immutable.min.js is broken on
http://facebook.github.io/immutable-js/ but README has correct link
so this replaces broken with correct one

**Overview**
```
<a [-href=\"./dist/immutable.min.js\">dist/immutable.min.js</a>\nor-]{+href=\"https://github.com/facebook/immutable-js/blob/master/dist/immutable.min.js\">dist/immutable.min.js</a>\nor+}
```

fix_immutable_link


leebyron:
This file is automatically generated by our website's build script, it pulls directly from our README.md, so unfortunately if I were to merge this, it would be quickly overwritten (in fact, this file should probably be in the .gitignore)

leebyron:
Seems like what was actually going on was that the readme.json file *wasnt* being autogenerated anymore :(. I got that all fixed up in aca20705ee96bb00b038b24a531d9f92014ffdf2 which happened to solve this issue. Thanks also for bringing this to my attention, I added these auto-gen files to the .gitignore while I was at it.

raunakkathuria:
@leebyron No problem. Thanks for fixing.

Issue #456, opened 2015-04-22
mergeDeepWith providing keyPath instead of key.


rickyvetter:
The current mergeDeepWith passes the two values and the key that they share - would it make more sense to provide the entire keyPath? With just the key, it's pretty difficult to understand where the object exists within the structure.

leebyron:
Good idea!

facebook-github-bot-4:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

gasi:
:+1:

mattkrick:
:+1:

PR #794, opened 2016-02-21
add keyPath to merger function in mergeDeepWith

rickyvetter:
This addresses #456.

It maintains key as the third argument in the merger to avoid breaking code.

This has added some messy overloading on mergeDeepWith to be able to pass the keyPath recursively and still be able to pass in n iterables to merge. I'd love recommendations to make it better.


src/Map.js
```diff
@@ -743,12 +747,12 @@ export function deepMerger(existing, value, key) {
     is(existing, value) ? existing : value;
 }
 
-export function deepMergerWith(merger) {
+export function deepMergerWith(merger, keyPath = []) {
   return (existing, value, key) => {
     if (existing && existing.mergeDeepWith && isIterable(value)) {
-      return existing.mergeDeepWith(merger, value);
+      return existing.mergeDeepWith([...keyPath, key], merger, value);
```
leebyron: `keyPath.concat([ key ])` would result in a much cleaner built JS file, but the same behavior you're looking for.

src/List.js
```diff
@@ -140,8 +140,12 @@ export class List extends IndexedCollection {
     return mergeIntoListWith(this, deepMerger, arguments);
   }
 
-  mergeDeepWith(merger, ...iters) {
-    return mergeIntoListWith(this, deepMergerWith(merger), iters);
+  mergeDeepWith(...args) {
+    var keyPath, merger, iters;
+    typeof args[0] === 'function' ?
+      [merger, ...iters] = args :
+      [keyPath, merger, ...iters] = args;
```
leebyron: I think providing a keyPath as a first argument is confusing. When would someone write code that uses it? I fear that it's more likely that someone accidentally hits this codepath by forgetting to provide a `merger` function.

src/Map.js
```diff
@@ -743,16 +743,20 @@ export function deepMerger(existing, value, key) {
...
     }
-    var nextValue = merger(existing, value, key);
+    var nextValue = merger(existing, value, key, keyPath.concat(key));
     return is(existing, nextValue) ? existing : nextValue;
   };
 }
 
+export function mergeDeepWithKeyPath(existing, merger, keyPath, ...iters) {
+  return existing.mergeWith(deepMergerWith(merger, keyPath), ...iters);
+}
```
rickyvetter: This avoids the keyPath getting passed directly to mergeDeepWith. It also avoids adding another method to map and list, while still allowing a call to the custom mergeIntoMapWith/ListWith functions by using mergeWith. Not sure if this is the best solution, but I definitely like it more than before.

ldabiralai:
This is exactly what I'm looking for, will this be merged at some point?

It might be an idea to add this as a new function, rather than modify the existing one otherwise it'll break other implementations of this.

kozlitinaelja:
Hi there. I am starting handling backlog. I know it is been a while since you've submitted PR, sorry for such delay. The are some conflicts. Could you, please, rebase? 😅

rickyvetter:
Hey @kozlitinaelja I tried to do a rebase and it ended up changing way more than the original PR. I haven't kept up with the changes since thie PR was originally put up. I put up the new change set here if you'd like to take a look: https://github.com/rickyvetter/immutable-js/tree/keypath2

If you know what's making the rebase do all of the extra changes and it's easy to fix, let me know! This was more of an exploration on how to do this in the first place. I know a couple other people were looking for this feature, but I'm not sure this was implemented in the best way. Happy to drop this PR as well if it doesn't look like it works.

leebyron:
I'd prefer to see an approach that didn't require the nested concatenation loop of the key-path. That's liable to make mergeDeep much slower for everyone regardless of if they rely on this value.

rickyvetter:
Definitely makes sense. Thanks for the feedback! I won't be able to work on this in the short term, but I hope someone else might be able to work with this idea :)

PR #801, opened 2016-03-04
Fix links

dijonkitchen:
Why:
Website was linking to 404 pages.

How:
Change links to absolute links rather than relative links since the website has a different root than the Readme file.


leebyron:
awesome. thanks so much for your help

PR #807, opened 2016-03-07
Support typings in package.json by autogenerating immutable-nonambien…

danielearwicker:
As noted in #734 it is now possible to tell TypeScript where to find type defs in `package.json` by setting the `typings` property.

However the `immutable.d.ts` file is an ambient module (exports a module path name as a string). This is why pull request #779 didn't work. What is needed is a plain module that directly exports the `Immutable` namespace.

There doesn't appear to be a way to define a module in one file and reuse it in both ambient and non-ambient contexts. So I've added a trivial bit of regex find/replace to the existing gulp task `'typedefs'` to generate `dist/immutable-nonambient.d.ts`, and amended `package.json` to expose via the `typings` property.

This means `immutable.d.ts` can continue to be maintained as it is, as long as the two significant bits of wrapping are left as they are. Maybe they should have comments around them in `immutable.d.ts` to warn people against changing them?

Also I haven't removed the `typescript` property in `package.json` in case it is important for backward compatibility. It appears to be harmless.

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -1,0 +1,2540 @@
+/**
+ *  Copyright (c) 2014-2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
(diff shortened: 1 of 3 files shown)


danielearwicker:
Oh damn, this isn't right. It's far worse than I thought...

Issue #676, opened 2015-10-24
Iterable.findEntry and Iterable.findLastEntry do not respect notSetValue argument


swac:
The [documentation for `Iterable.findEntry` and `Iterable.findLastEntry`](https://facebook.github.io/immutable-js/docs/#/Iterable/findEntry) mentions an optional `notSetValue` argument, but this argument does not appear in the [code](https://github.com/facebook/immutable-js/blob/master/src/IterableImpl.js#L176-L189). Instead it just returns `undefined` when the predicate isn't satisfied. I would send a pull request to fix this, but I'm not sure if it's wise to have it return something like `[-1, notSetValue]` because that wouldn't be falsy and could break call sites. It might be better to just update the documentation to not show this argument instead.

kabalage:
I think they should return [-1, notSetValue]. Now destructuring assignment is broken when none of the elements satisfy the predicate:
```const [i, elem] = list.findEntry(predicate)```

**Comment by quatrano** [comment]
I also stumbled upon this code/doc inconsistency.  I would prefer that the code were implemented to support the docs. I want to use this feature as follows:
```
var entry = list.findEntry(predicate, null, [list.length, new Immutable.Map()]);
entry[1] = entry[1].set('foo', 'bar');
list = list.set(entry[0], entry[1]);
```
I'm new to this library though, so maybe I'm misusing the paradigm.

leebyron:
Fixed!

PR #822, opened 2016-04-02
Iterable.findEntry and Iterable.findLastEntry now respect notSetValue argument #676

quatrano:


leebyron:
Nice simple implementation, thanks!

PR #838, opened 2016-04-16
doc fix for find and findEntry

nfcampos:
To make them more consistent with the docs for `findLast` and `findLastEntry`, added the word `first` to make explicit that `find` and `findEntry` return the first element for which the predicate returns `true`.


leebyron:
Thanks

PR #840, opened 2016-04-16
Update flow Libdefs

marudor:
I've took a quick look and changed some stuff.
In general very solid.
I've removed the booleany = any type as mixed is favored over any for this. It's used for checks that require a truthy or falsy value. Flow doesn't have them, but mixed allows the return values to be any but won't loose any types. Any would loose them.


leebyron:
Excellent work, thanks!

PR #841, opened 2016-04-16
Relax Record class for now

marudor:
As discussed here: https://github.com/facebook/immutable-js/pull/805#issuecomment-210793138 the new Record libdef broke old code.
I've relaxed it again to any. We will loose comfort but it will not error.
Ideas to improve exist, can't do them yet though.

Also fixed errors I've introduced with the last PR. (minor)

### DIFF
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -29,7 +29,9 @@
  */
 type ESIterable<T> = $Iterable<T,void,void>;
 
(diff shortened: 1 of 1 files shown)


leebyron:
cc @bestander This should probably also fix https://circleci.com/gh/facebook/react-native/5344

bestander:
I can check on Monday if you release 3.8.1

MSch:
Thank you so much for the fix!

PR #842, opened 2016-04-17
Modernification: using Mocha for tests

thomastuts:
Related issue: https://github.com/facebook/immutable-js/issues/742
### Introduction

I started implementing the tests using Mocha instead of Jest. I figured it'd be a good idea to send a preliminary PR and do things in the open. This way, I'd like to get as much feedback on it as possible before we convert too many tests. 
### Changes so far
- Added `npm` script (`npm run testmocha`) to run Mocha tests. The Mocha tests are also written in TS like the current tests. The code is compiled using [`ts-node`](https://github.com/TypeStrong/ts-node).
- Added watcher to re-run tests on file changes (`npm run testmocha:watch`). This is to make my life a little easier when converting the tests, and we can remove this when the tests are fully converted (although having a watcher is always nice, so we could leave it in).
- Tests are all in the `test/` directory, adhering to Mocha's convention. This also makes it easy to run the Mocha and Jest tests side-by-side, and to port them over incrementally.
- [`mocha-testcheck`](https://github.com/leebyron/mocha-check) is used for the generative testing, replacing `jasmine-check`.
- Some code styling was applied: making use of whitespace where applicable and using `const` and `let` where appropriate (not sure if you're okay with this @leebyron, if you'd rather convert the tests verbatim, let me know).
### Current issues
- The `Map` test has two failing test cases (`accepts flattened pairs via of()` and `does not accept mismatched flattened pairs via of()`) where `Map.of()` seems to be undefined, while it is defined in the Jest test. Any idea why that might be? I'm importing the exact same `immutable` as the Jest tests are, yet it still seems to be different for some reason.

---

I'll update this PR with more converted tests and any questions I might have regarding failing tests.


.gitignore
```diff
@@ -12,3 +12,4 @@ readme.json
 /pages/out
 /pages/resources/immutable.d.json
 /pages/resources/readme.json
+typings/
```
leebyron: Preceding slash as well

leebyron:
This is looking awesome, thanks for your help! I agree with this approach: set up a test infra and then let's do one test migration per PR to keep things easy to follow. 

I'm not sure what's up with the failing tests. I can help investigate tomorrow

package.json
```diff
@@ -23,6 +23,8 @@
     "build": "grunt default && gulp default",
     "lint": "eslint src/ && grunt lint && gulp lint",
     "testonly": "./resources/node_test.sh",
+    "testmocha": "mocha --require ts-node/register --recursive 'test/*.ts'",
```
leebyron: I prefer using a *-test.ts notation for mocha tests, so if we ever need any helper files they can live in the directory but won't be run as tests

thomastuts: Sounds good. Would you settle for something like `*.test.ts`? Feels just a little bit cleaner to me. Otherwise, I'd get the feeling that 'test' is part of the filename instead of telling you what the file is, if that makes sense.

thomastuts:
Huh, that's strange. Running the tests that use `Map.of()` on another machine works just fine, the `Map.of()` isn't undefined there. I'll investigate when I get home, but it looks like we're good to go for now. All tests for `Map` have been ported, and pass!

leebyron:
Closing in favor of #1018

PR #844, opened 2016-04-18
Rebuild and reload typedefs in dev server on change

thomastuts:
This PR adds a watcher for any file in the `/type-definitions/` folder and rebuilds the docs on changes. This makes it easier to work on docs and seeing the results without constantly having to rebuild manually.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -237,8 +237,8 @@ gulp.task('dev', ['default'], function() {
   gulp.watch('./app/docs/src/**/*.js', ['rebuild-js-docs']);
   gulp.watch('./app/**/*.html', ['pre-render', 'pre-render-docs']);
   gulp.watch('./app/static/**/*', ['statics', 'statics-docs']);
-  gulp.watch('./type-definitions/*', ['typedefs', 'rebuild-js-docs'], function () {
-    browserSync.reload();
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -237,6 +237,9 @@ gulp.task('dev', ['default'], function() {
   gulp.watch('./app/docs/src/**/*.js', ['rebuild-js-docs']);
   gulp.watch('./app/**/*.html', ['pre-render', 'pre-render-docs']);
   gulp.watch('./app/static/**/*', ['statics', 'statics-docs']);
+  gulp.watch('./type-definitions/*', ['typedefs', 'rebuild-js-docs'], function () {
```
leebyron: rebuild-js-docs already calls `browserSync.reload()` so I don't think you need the callback. However you do probably want to ensure `typedefs` are rebuild first, use `sequence` to have them done sequentially rather than in parallel

thomastuts: You're absolutely right. Fixed!

lacker:
This looks good to me but Travis is failing for some apparently-flawed reason. Let me try rerunning Travis and if that's good then let's accept this.

kozlitinaelja:
@lacker, is this PR good to be merged?

leebyron:
Most excellent.

PR #849, opened 2016-04-19
Update Immutable example console version to 3.8.1

Bhullnatik:
I wanted to try the changes with the new version quickly, but saw that in the console it was still running the 3.7.6 version.

Not sure if there is something else to change, and everything was working fine locally.


lacker:
Thanks for sending in this pull request! It looks like someone independently made this change, so this PR is unneeded and I am going to close it. But thanks for pointing this out!

PR #850, opened 2016-04-19
Fix toOrderedMap, toOrderedSet Flow annotations

rvikmanis:


marudor:
Actually this will introduce some problems.
Thats why I left them as Map/Set for now.
But the main signatures for Maps/Set is the same for Ordered.
So it shouldn't be a problem to leave them.

mitermayer:
I also hit on this problem

@marudor  on the project https://github.com/facebook/draft-js with immutable '3.7.4' the flow build works just fine.

But since the dependency on the package.json is set as '^0.7.4' the version that gets installed via npm install is ''3.8.1" which is incompatible and fails

lacker:

howardjing:
I fixed the merge conflicts in this PR / added tests / updated annotations for OrderedSet and OrderedMap in this PR: https://github.com/facebook/immutable-js/pull/1027

kozlitinaelja:
Hi there. I am starting handling backlog. I know it is been a while since you've submitted PR, sorry for such delay. The are some conflicts. Could you, please,  rebase? 😅

howardjing:
@kozlitinaelja #1027 is merged into master, which is a superset of this PR. I think this PR can be closed.

PR #855, opened 2016-04-21
Include Default Export

marudor:
Apparently it is fairly common for people to use Immutable with a default import 
```javascript
import Immutable from 'immutable';
const foo = Immutable.List();
```
currently flow complains due to missing default export.
I've added one.

### DIFF
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -220,8 +220,8 @@ declare class KeyedIterable<K,V> extends Iterable<K,V> {
   flatten(shallow?: boolean): /*this*/KeyedIterable<any,any>;
 }
 
-declare class IndexedIterable<T> extends Iterable<number,T> {
-  static <T>(iter?: ESIterable<T>): IndexedIterable<T>;
+declare class IndexedIterable<A> extends Iterable<number,T> {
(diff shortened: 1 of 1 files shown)


ooflorent:
@marudor The correct way to import Flow using ES6 should be:

```js
import * as Immutable from "immutable"
import {List} from "immutable
```

Using the _default_ export is an abuse permitted by Babel (and perhaps other transpilers).

marudor:
I know it should.
Sadly thats how its often used.

If it is a practice often used it should be supported. (I had 4 ppl asking it in the flowtype irc already)

zertosh:
Immutable actually only exports a default https://github.com/facebook/immutable-js/blob/e96d73f/src/Immutable.js#L26-L46. So either the type def is _incorrect_, or `src/Immutable.js` should be updated to:

```
export { Seq } from './Seq'
export { Collection } from './Collection'
export { OrderedMap } from './OrderedMap'
export { List } from './List'
export { Map } from './Map'
export { Stack } from './Stack'
export { OrderedSet } from './OrderedSet'
export { Set } from './Set'
export { Record } from './Record'
export { Range } from './Range'
export { Repeat } from './Repeat'
export { is } from './is'
export { fromJS } from './fromJS'
export { Iterable } from './IterableImpl'
```

PR #856, opened 2016-04-22
Deprecate the use of map/fiter/reduce (see #505)

AugustinLF:
As said in #505 some functions of Record don't behave as expected. They plan to be removed, but for now, it's confusing for anyone not aware of the problem (you can easily lose a couple of hours because of that), so warning the user should be reasonable way to let them know.

You mention "other API methods", but not knowing which ones you're talking about, I'll add them later if needed :)


src/Record.js
```diff
@@ -145,6 +145,19 @@ RecordPrototype.withMutations = MapPrototype.withMutations;
 RecordPrototype.asMutable = MapPrototype.asMutable;
 RecordPrototype.asImmutable = MapPrototype.asImmutable;
 
+// Some functions from Collection should not be called on a Record, since their behavior is not the one expected
+// https://github.com/facebook/immutable-js/issues/505
+function deprecateFunction(RecordPrototype, functionName) {
+  RecordPrototype[functionName] = function(...params) {
+    console.warn(`Record${functionName} has been deprecated. See https://github.com/facebook/immutable-js/issues/505 for more details`);
+    return KeyedCollection.prototype[functionName].call(this, ...params);
```
AugustinLF: I would have used super to access KeyedCollection prototype, but the transpiler refuses it.

caabernathy:
Bot check

facebook-github-bot-9:

src/Record.js
```diff
@@ -145,6 +145,19 @@ RecordPrototype.withMutations = MapPrototype.withMutations;
 RecordPrototype.asMutable = MapPrototype.asMutable;
 RecordPrototype.asImmutable = MapPrototype.asImmutable;
 
+// Some functions from Collection should not be called on a Record, since their behavior is not the one expected
+// https://github.com/facebook/immutable-js/issues/505
+function deprecateFunction(RecordPrototype, functionName) {
+  RecordPrototype[functionName] = function(...params) {
+    console.warn(`Record${functionName} has been deprecated. See https://github.com/facebook/immutable-js/issues/505 for more details`);
```
philipp-spiess: Seems like you're missing a separator between `Record` and the `${functionName}`, otherwise it will be parsed into: `Recordfilter` 😊

lacker:
It looks like this PR has been abandoned since that last comment never got any response. I am going to close it under that theory but if you are still interested in working on this please feel free to reopen!

AugustinLF:
Hey, sorry for not updating before, since I didn't get any feedback on the PR, I dropped it! I amended my [commit](https://github.com/AugustinLF/immutable-js/blob/feature/deprecateRecordMap/src/Record.js#L150), but I think that since you closed the PR, it doesn't show up anymore.

I can rebase my branch on top of master if you want!

lacker:
OK let's see what we should do here. I guess one question is how we should handle deprecations. I don't think the approach of doing a `console.warn` when you use it is quite the right one - that might cause trouble in some environments because you might have `console.warn` do something very interrupty and not even realize what's happening. I think we would be better off just commenting that methods are deprecated, and then removing them in the next major version bump. I will ping @leebyron about what the right timing is here.

PR #860, opened 2016-04-23
Use latest version of Immutable in docs

javiercejudo:
I got a bit confused seeing `Immutable.Map.of()` was not available in the console until I realised the docs had a previous version of Immutable.


lacker:
Hey, thanks for sending in this pull request! I don't think this is the right thing to do, because then any breaking change to Immutable would also be a breaking change to the website. So I am going to close this pull request, but thanks again for putting in the effort to improve things.

PR #865, opened 2016-04-28
added an example for the groupBy method

winkerVSbecks:
I've noticed that people new to functional style programming are usually not familiar with the API of `groupBy`. Therefore, added an example to demonstrate it. 

I was hoping to add a few more examples to the docs. Would it be better to submit them as one PR or should each be a separate PR?


lacker:
Either one big PR or separate PRs is ok. Maybe separate PRs are nicer because it's easier to just merge more things in then.

lacker:
Oh and thanks for improving this documentation!

PR #876, opened 2016-05-15
Uncomment types for Record, fix for .equals

nmn:
The Record type refs work as of Flow version 0.24.2.

The .equals method on `_Iterable` was causing Set.equals to not accept another Set. This seems to fix it.

### DIFF
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -620,9 +620,12 @@ declare function Range(start?: number, end?: number, step?: number): IndexedSeq<
 declare function Repeat<T>(value: T, times?: number): IndexedSeq<T>;
 
 declare class Record<T: Object> {
-  static <T: Object>(spec: T, name?: string): T & Record<T>;
+  static <T: Object>(spec: T, name?: string): Class<T & Record<T>>;
+  constructor(initialObject?: $Shape<T>): T & Record<T>;
   get<A>(key: $Keys<T>): A;
+  getIn(key: Array<string | number>): any;
   set<A>(key: $Keys<T>, value: A): T & Record<T>;
+  setIn(key: Array<string | number>, value: any): T & Record<T>;
   remove(key: $Keys<T>): T & Record<T>;
 }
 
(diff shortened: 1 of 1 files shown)


type-definitions/immutable.js.flow
```diff
@@ -619,13 +619,11 @@ declare class Stack<T> extends IndexedCollection<T> {
 declare function Range(start?: number, end?: number, step?: number): IndexedSeq<number>;
 declare function Repeat<T>(value: T, times?: number): IndexedSeq<T>;
 
-//TODO: Once flow can extend normal Objects we can change this back to actually reflect Record behavior.
-// For now fallback to any to not break existing Code
 declare class Record<T: Object> {
-  static <T: Object>(spec: T, name?: string): /*T & Record<T>*/any;
+  static <T: Object>(spec: T, name?: string): T & Record<T>;
```
tallixun: Hi, are you sure that this line is correct? Isn't it supposed to be (I have checked it only with flow 0.26, so maybe it is working with 0.24):

```
static <T: Object>(spec: T, name?: string): Class<T & Record<T>>;
```

nmn: Yes, you're right about this. I checked the other cases, and so I just uncommented the things I saw. I'll fix this.

nmn:
Hold off on merging for now. Thanks @tallixun

nmn:
Fixed outstanding issues. Added the `getIn` and `setIn` methods for Records.
If you have any suggestions for improving the type definitions let me know.

@tallixun would love another review

philipp-spiess:
Hey @nmn. I would love to see your PR merged. However, there are a few method definitions missing. So far I found:

- `Record#updateIn()`
- `Record#toJS()`
- `Record#toObject()`
- `Record#update()
- `Record#withMutations()`
- `Record#delete()`
- `Record#merge()`
 
I think it would make sense to extend/mix in `KeyedCollection`, since this is also the way it is implemented: https://github.com/facebook/immutable-js/blob/master/src/Record.js#L18

lacker:

kozlitinaelja:
Hi there. I am starting handling backlog. I know it is been a while since you've submitted PR, sorry for such delay. The are some conflicts. Could you, please, rebase? 😅

PR #878, opened 2016-05-18
Flow type declarations test

wokalski:
I'm using `Flow` with `Immutable` but I found numerous false positives when I was working with it. 
By _false positives_ I mean cases where types were ignored. Most of the time it resulted in types dropping their generic definition (from `T<K,V> -> T`).

This is `WIP` because:
- tests for `Iterable` and `Seq` are not implemented yet
- `merge` on maps ignores types when given a `ESIterable`. It only infers types from plain objects.
- after initialization without type (for instance by calling `Map()`) and setting a value on empty object it doesn't correctly infer types

```
var map: Map<number, string> = Map().set(1, '')
```

This might be a flaw of Flow so feel free to point it here.

---

In f31e5c2 I copied the definition of `$Iterable` to `immutable-js`. I did it for two reasons:
1. It didn't work
2. `$Iterable` is private

---

Tests for `Iterable` and `Seq` are not implemented yet. I had some problems implementing tests for `Iterable`. Ideally I would like to cast some concrete type to it, but I'm not sure how to do it.

Is the type of `Map<K,V>` `Iterable<K,V,void,void,void>`?

---

**Next steps:**
- If tests receive a 👍🏻 I'm going to implement tests for other types.
- ~~I believe I found a bug in Flow. Namely argument of union type is not correctly resolved. I am not sure how to phrase it in order to file a bug in flow. An instance of this problem can be seen [here](https://github.com/facebook/immutable-js/blob/ea507393e1ff306025938341d07d8153b2bf95ce/type-definitions/immutable.js.flow#L503). `ESIterable` is casted to `{ [key: string]: T }`. How do I name this, _`casting`_?~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
-node_js: stable
+node_js:
+    - '6'
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ env:
     - CXX=g++-4.9
     - secure: "VDsxy30sE9ivdqoXkaKXo0czbS4brNpwKEIblu7f1gVLx7OD9pjTc78cdwrVbZDBYroSiYVYuUrLDjpVjH88lL/LxRrru3V0CIlAqqa+ssXcqycCaT/6ds+ZymuTTGRh+Mf12pIKO+yc8jTov2M7AzPJdpS+ORP5dImYyE3ex9s="
 
-script: ./test_type_defs.sh
-
 addons:
   apt:
     sources:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ env:
     - CXX=g++-4.9
     - secure: "VDsxy30sE9ivdqoXkaKXo0czbS4brNpwKEIblu7f1gVLx7OD9pjTc78cdwrVbZDBYroSiYVYuUrLDjpVjH88lL/LxRrru3V0CIlAqqa+ssXcqycCaT/6ds+ZymuTTGRh+Mf12pIKO+yc8jTov2M7AzPJdpS+ORP5dImYyE3ex9s="
 
+script: ./test_type_defs.sh
+
 addons:
   apt:
     sources:
--- a/test_type_defs.sh
+++ b/test_type_defs.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -o errexit
+
+npm install
+cd type-definitions/tests
+../../node_modules/flow-bin/cli.js
+
--- a/test_type_defs.sh
+++ b/test_type_defs.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-set -o errexit
-
-npm install
-cd type-definitions/tests
-../../node_modules/flow-bin/cli.js
-
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -27,7 +27,28 @@
  *
  * Note that Immutable values implement the `ESIterable` interface.
  */
-type ESIterable<T> = $Iterable<T,void,void>;
+
+type IteratorResult<Yield,Return> = {
+  done: true,
+  value?: Return,
+} | {
+  done: false,
+  value: Yield,
+};
+
+
+interface _Iterator<Yield,Return,Next> {
+    @@iterator(): _Iterator<Yield,Return,Next>;
+    next(value?: Next): IteratorResult<Yield,Return>;
+}
+
+
+interface _ESIterable<Yield,Return,Next> {
+    @@iterator(): _Iterator<Yield,Return,Next>;
+}
+
+type Iterator<T> = _Iterator<T, void, void>
+type ESIterable<T> = _ESIterable<T,void,void>;
 
 declare class Iterable<K, V> extends _Iterable<K, V, typeof KeyedIterable, typeof IndexedIterable, typeof SetIterable> {}
 
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -639,8 +639,8 @@ declare module 'immutable' {
   declare function Range(start?: number, end?: number, step?: number): IndexedSeq<number>;
   declare function Repeat<T>(value: T, times?: number): IndexedSeq<T>;
 
-  //TODO: Once flow can extend normal Objects we can change this back to actually reflect Record behavior.
-  // For now fallback to any to not break existing Code
+  // TODO: Once flow can extend normal Objects we can change this back to actually reflect Record behavior.
+  // For now fallback to any to not break existing code
   declare class Record<T: Object> {
     static <T: Object>(spec: T, name?: string): /*T & Record<T>*/any;
     get<A>(key: $Keys<T>): A;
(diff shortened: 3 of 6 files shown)


wokalski:
pinging @marudor as a major contributor to this

type-definitions/immutable.js.flow
```diff
@@ -464,7 +485,7 @@ declare class List<T> extends IndexedCollection<T> {
 }
 
 declare class Map<K,V> extends KeyedCollection<K,V> {
-  static <K, V>(): Map<K, V>;
+  static <K,V>(_: void): Map<K,V>;
```
marudor: Whats the purpose of this?

type-definitions/immutable.js.flow
```diff
@@ -27,7 +27,28 @@
...
+type IteratorResult<Yield,Return> = {
+  done: true,
+  value?: Return,
+} | {
+  done: false,
+  value: Yield,
+};
+
+
+interface _Iterator<Yield,Return,Next> {
```
marudor: No need to redefine Iterator. There is $Iterator (https://github.com/facebook/flow/blob/master/lib/core.js#L401)

type-definitions/immutable.js.flow
```diff
@@ -27,7 +27,28 @@
...
+    @@iterator(): _Iterator<Yield,Return,Next>;
+    next(value?: Next): IteratorResult<Yield,Return>;
+}
+
+
+interface _ESIterable<Yield,Return,Next> {
```
marudor: We can use the $Iterable here (https://github.com/facebook/flow/blob/master/lib/core.js#L407)

marudor:
Added some comments - otherwise looks nice.

type-definitions/immutable.js.flow
wokalski: I thought so. However, it doesn't work (at least for me). I noticed that in my project and it didn't work here either. If I checkout the first commit here; the one which adds tests, I get whopping 45 errors, most of them connected with generics being simply ignored.

type-definitions/immutable.js.flow
wokalski: Type inference didn't work on objects without it. It looked like `flow` somehow ignores the object type, falling back to the constructor without named parameters. Since parameters in both constructors are optional, I thought that simply removing the empty constructor would solve the problem but it didn't. I added this one, and honestly I copied it from [here](https://github.com/facebook/flow/blob/master/lib/core.js#L424). It's hard for me to give further insight since I'm new to both `js` and `flow`.

I encourage to run the test without the patches first (to check out at d7ee48c) and then check out subsequent commits.

type-definitions/immutable.js.flow
wokalski: [see this](https://github.com/facebook/immutable-js/pull/878#discussion_r64115951)

wokalski:
Thanks for review. I would like to note that the test for `Map.merge` doesn't pass because types are ignored. I would love to get some help fixing this.

wokalski:
ping @marudor

wokalski:
Closing. Work on this will be continued [here](https://github.com/flowtype/flow-typed/pull/401)

wokalski:
I'm on a pretty epic bug. I wanted to fix imports according to [this]() comment. The problem is that when I wrap everything in `declare module` (which I should do) everything is exported correctly, except for `List`. The funny thing is that if I rename `List` to `Lista` or really anything, it works.

.travis.yml
```diff
@@ -1,12 +1,15 @@
...
-node_js: stable
+node_js:
+    - '6'
 
 env:
   global:
     - CXX=g++-4.9
     - secure: "VDsxy30sE9ivdqoXkaKXo0czbS4brNpwKEIblu7f1gVLx7OD9pjTc78cdwrVbZDBYroSiYVYuUrLDjpVjH88lL/LxRrru3V0CIlAqqa+ssXcqycCaT/6ds+ZymuTTGRh+Mf12pIKO+yc8jTov2M7AzPJdpS+ORP5dImYyE3ex9s="
 
+script: ./test_type_defs.sh
```
leebyron: this would over-write the existing default script `npm test` - can we instead add the flow cli run to package.json scripts?

lacker: I'm probably missing something about Travis here, but what is this replacing? Like what was the test script entry point before `script` was set here? Just want to make sure we aren't omitting something now.

test_type_defs.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -o errexit
+
+npm install
+cd type-definitions/tests
+../../node_modules/flow-bin/cli.js
```
leebyron: also - when called from the context of npm scripts, node modules are added to PATH, so you could just write `flow` there.

type-definitions/tests/.flowconfig
```diff
@@ -0,0 +1,12 @@
+[libs]
+../immutable.js.flow
+
+[ignore]
+.*/node_modules/.*
+
+[options]
+suppress_comment=^\\( \\|\n\\)*\\$ExpectError\\(([^)]*)\\)?$
```
lacker: What is this line doing? Is there a way to comment the .flowconfig to explain this?

type-definitions/tests/immutable-flow.js
```diff
@@ -0,0 +1,431 @@
+/*
+ * @flow
+ */
+
+// Some tests look like they are repeated but it's not actually the case; It is very likely to have a false positivie because Flow might not complain about an instance of (what it thinks is) T to be assigned to T<K, V>
```
lacker: "positive"

could you break this line up to keep to 80 char max or so

type-definitions/immutable.js.flow
```diff
@@ -21,643 +21,634 @@
...
+
+    flatten(depth?: number): /*this*/Stack<any>;
+    flatten(shallow?: boolean): /*this*/Stack<any>;
+  }
+
+  declare function Range(start?: number, end?: number, step?: number): IndexedSeq<number>;
+  declare function Repeat<T>(value: T, times?: number): IndexedSeq<T>;
+
+  //TODO: Once flow can extend normal Objects we can change this back to actually reflect Record behavior.
+  // For now fallback to any to not break existing Code
```
lacker: Code -> code

yea i know nitpicky

lacker:
I just wanted to say that overall having flow type tests in this repo will make it a lot better, because it'll be a lot easier for us to accept a lot of the type-bug-fixes that have been submitted, so you are a hero for putting this PR together @wokalski

.travis.yml
wokalski: It's the first time I use `node` on travis and I felt that something was wrong, too. Thankfully @leebyron pointed out that `npm install` and then `npm test` is automatically ran by travis. There's no need to specify a script.

test_type_defs.sh
wokalski: TIL - thanks

wokalski:
thanks for the review - I broke CI and didn't know about it. The little things you mentioned should be fixed now.

lacker:
Does `$ExpectError` actually fail if there is no error there or does it just allow an error on the subsequent line?

wokalski:
Good point! It does fail.

lacker:
YOLO

wokalski:
hahaha

PR #892, opened 2016-06-15
Nit: Fix License and Patent links for gh-pages

jeffreylin:
These links at the bottom of https://facebook.github.io/immutable-js/ were broken since they're relative and https://facebook.github.io/immutable-js/LICENSE and https://facebook.github.io/immutable-js/PATENTS don't exist.

This fixes the issue by using the full link instead of the relative link. I poked around regenerating the hg-pages for you guys, but couldn't find an obvious way to do it so heads up someone will have to regenerate them. =]

### DIFF
--- a/README.md
+++ b/README.md
@@ -471,4 +471,4 @@ name. If you're looking for his unsupported package, see [this repository](https
 License
 -------
 
-`Immutable` is [BSD-licensed](./LICENSE). We also provide an additional [patent grant](./PATENTS).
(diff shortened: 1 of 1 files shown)


facebook-github-bot-6:
Thank you for your pull request.  As you may know, we require contributors to sign our Contributor License Agreement, and we don't seem to have you on file and listed as active anymore.  In order for us to review and merge your code, please email cla@fb.com with your details so we can update your status.

lacker:
Thanks!

Issue #825, opened 2016-04-06
Map::deleteAll for removing multiple keys from a map


xixixao:
Could we add this to the library? What is the policy around these helpers combining multiple collections?

``` javascript
const base = Immutable.Map({x: 1, y: 2, z: 3});
const shrinked = base.deleteAll(['x', 'z']); // or `subtract`?
// => shrinked == Immutable.Map({y: 2});
```

(Java collections usually have these)

levidobson:
deleteAllIn could also be useful.

rawpixel-vincent:
For those who find this, it's how I did it

```javascript
const base = Immutable.Map({x: 1, y: 2, z: 3});

let myNewBase = base;
['y', 'z'].forEach((idToDelete) => { myNewBase = myNewBase.delete(idToDelete); });

```

umidbekkarimov:
I guess this https://github.com/facebook/immutable-js/commit/ba535442737d4045cbd412cae3350b4413f9ee9b should be close this issue.

Issue #228, opened 2014-12-05
.remove() mutations on list returns empty immutable 


varunarora:
Just like in the example on the doc about .withMutations(), the immutable returned that end of the multiple .push() mutations on the list works great. However, .remove() mutations diminish the mutable copy inside the function down to an empty immutable object, and not a mutable copy after being mutated.

Do you know why this might be happening?

facebook-github-bot-6:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

carlsverre:
@leebyron Why does `delete(2)` result in all the elements being removed while `delete(0)` works as expected?  (see code example above)

callin2:
I have same issue.  List.delete() work only on first and last index. inside withMutation().

PR #895, opened 2016-06-15
deliver #825 add Map::deleteAll for removing multiple keys from a map

vinnymac:
#### Summary

After some thought and benchmarking, `withMutations` turned out to be a performant way of implementing `deleteAll`. I tested against some other implementations such as mapping the keys to an object and merging, creating a new map for each key (obviously slow), and filtering the map by keys that did not exist in the provided iterable. They all resulted in slower benchmarks.
#### Questions
1. I see `iter`, `k`, and `v` being used throughout the codebase, while the contributions state to not use abbreviations, I avoided them wherever possible, is this a rule I should be following?
2. Should I be using `forceIterator` on the `keys` parameter or is my usage of `IndexedIterable` applicable here?
3. In `withMutations` I use `remove`, is it okay even though it is an [undocumented behavior](https://github.com/facebook/immutable-js/issues/228#issuecomment-67554984)? It works correctly for `Map` in this case.
4. I have never worked with `TypeScript` or `Flow` before, did I identify the supported types properly?
5. Should we keep the behavior of returning the current `Map` when no `keys` are passed in or assert that the `keys` parameter is defined and throw if it is not defined? i.e. `Map().deleteAll() // The required parameter keys was not provided.`
6. Do we want to also provide the alias `subtract` for this method on `Map`?
#### ToDo
- [x] Add tests
- [x] Add documentation with an example


qrohlf:
This PR looks good to me, and would be helpful for me in my usage of the library. Maintainers, anything holding it up from being merged (other than the merge conflicts)? 

I'm happy to adopt the PR and fix the merge conflicts if there are no other outstanding issues with it.

vinnymac:
@qrohlf I've resolved the merge conflicts.

src/Map.js
```diff
@@ -78,6 +78,18 @@ export class Map extends KeyedCollection {
...
 
+  deleteAll(keys) {
+    var iterable = IndexedIterable(keys);
+
+    if (iterable.size === 0) {
+      return this;
+    }
+
+    return this.withMutations(map => {
+      iterable.forEach((key) => map.remove(key, NOT_SET));
```
leebyron: `.remove()` only accepts one argument

leebyron: please remove the wrapping `()` for consistency with the previous arrow function argument style

type-definitions/Immutable.d.ts
```diff
@@ -601,6 +601,16 @@ declare module Immutable {
...
 
     /**
+     * Returns a new Map which excludes the provided `keys`.
+     *
+     *     var names = Immutable.Map({ a: "Aaron", b: "Barry", c: "Connor" });
+     *     names.deleteAll(['a']); // { b: "Barry", c: "Connor" }
+     *
+     */
+    deleteAll(keys: Array<any>): Map<K, V>;
+    deleteAll(keys: Iterable.Indexed<any>): Map<K, V>;
```
leebyron: `keys: Iterable<any, K>`

src/Map.js
```diff
@@ -78,6 +78,18 @@ export class Map extends KeyedCollection {
     return this.updateIn(keyPath, () => NOT_SET);
   }
 
+  deleteAll(keys) {
```
leebyron: The other delete methods have a `remove` alias. Could you add one here as well

src/Map.js
```diff
@@ -78,6 +78,18 @@ export class Map extends KeyedCollection {
...
+    var iterable = IndexedIterable(keys);
```
leebyron: You should only need `Iterable` - it shouldn't matter that it's indexed

type-definitions/Immutable.d.ts
```diff
@@ -601,6 +601,16 @@ declare module Immutable {
     remove(key: K): Map<K, V>;
 
     /**
+     * Returns a new Map which excludes the provided `keys`.
+     *
+     *     var names = Immutable.Map({ a: "Aaron", b: "Barry", c: "Connor" });
+     *     names.deleteAll(['a']); // { b: "Barry", c: "Connor" }
+     *
+     */
+    deleteAll(keys: Array<any>): Map<K, V>;
```
leebyron: `Array<K>`

vinnymac:
@leebyron I made the requested changes other than implementing `deleteAll` for List.

I attempted to add support for `deleteAll` but `deleteAll([0, -1])` is returning a different result than `.remove(0).remove(-1)` due to my usage of `withMutations`. I am pretty sure #228 is the issue I am seeing.

I can implement this some other ways, but I would like to hear what you would like to do next in this situation before I move forward. Thank you for the advice and your time.

leebyron:
That definitely makes sense thanks for investigating. Of course deletion at indices will change the shape of the List itself. Negative indices cause issue there. Let's start with Map since we know that works. I'll help get a clean merge

PR #909, opened 2016-07-02
Use polymorphic 'this' return types for Map methods

Merott:
## The Problem

As a simplified example, consider the following `ABClass`, which extends from `ABRecord` (itself instantiated with `Record({ a: 1, b: 2 })`):

``` ts
const ABRecord = Record({ a: 1, b: 2 });

class ABClass extends ABRecord {
  a: number;
  b: number;

  setA(a: number) {
    return this.set('a', a);
  }

  setB(b: number) {
    return this.set('b', b);
  }
}
```

Because the return type of `Map.set` is specified explicitly as `Map<K, V>`, the following attempt at chaining the method calls will alert the TS compiler:

``` ts
const ab = new ABClass({ b: 3 });
const ab2 = ab.setA(5).setB(6);
console.log(ab2.a, ab2.b);
```

As a result, we must cast the return types of `ABClass.setA` and `ABClass.setB` methods explicitly to `ABClass` in order to satisfy the TS compiler.

---

The situation is more complicated when calling `withMutations`:

``` ts
const ab3 = ab.withMutations(
  mutable => mutable.set('a', mutable.a + 1).set('b', mutable.b + 1));
```

The above example fails to compile because `mutable` is typed to `Map<string, any>`, which does not have the properties `a` and `b`, so we have to either access them with `get`, or cast mutable to `ABClass`, as well as having to cast the return type that will be assigned to `ab3`.
## The Solution

This PR addresses those problems, by using polymorphic `this` return types for `Map.*` methods that result in a new `Map` of the same type.

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -471,7 +471,7 @@
      * Returns a new Map also containing the new key, value pair. If an equivalent
      * key already exists in this Map, it will be replaced.
      */
-    set(key: K, value: V): Map<K, V>;
+    set(key: K, value: V): this;
 
     /**
      * Returns a new Map which excludes this `key`.
@@ -480,13 +480,13 @@
      * the ES6 collection API.
      * @alias remove
      */
-    delete(key: K): Map<K, V>;
-    remove(key: K): Map<K, V>;
+    delete(key: K): this;
+    remove(key: K): this;
 
     /**
      * Returns a new Map containing no keys or values.
      */
-    clear(): Map<K, V>;
+    clear(): this;
 
     /**
      * Returns a new Map having updated the value at this `key` with the return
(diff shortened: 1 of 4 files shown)


facebook-github-bot-2:

facebook-github-bot-5:

lacker:
What's up with the `DocVisitor` changes?

This is really touching a zillion different things, huh. How did you test to make sure none of these are in error?

Merott:
As it currently stands `DocVisitor#isAliased` assumes that `this.aliases` isn't empty, and runs a check based on that assumption. I can't see how verifying that assumption before doing a check could lead to adverse behaviour.

Without this check, the `typedefs` task fails with the following error:

```
[23:44:11] TypeError: Cannot read property 'OrderedMap' of undefined
    at DocVisitor.isAliased (/Users/Merott/Workspace/Projects/immutable-js/pages/lib/DocVisitor.js:45:38)
    at DocVisitor.visitFunctionDeclaration (/Users/Merott/Workspace/Projects/immutable-js/pages/lib/DocVisitor.js:87:41)
```

I assumed that any breaking changes would be picked up by existing tests.

Could this change cause problems? Let me know and I'll investigate.

kozlitinaelja:
@lacker, is this PR ready to be merged?

leebyron:
Hey @Merott sorry for the huge delay but I'd love to readdress this.

Two primary concerns:

1) `this` is usually used to describe returning the same instance. Not the case for Map.
2) The problem described is relative to Record, but being applied everywhere. Could you re-scope the change to only apply to Record?

PR #918, opened 2016-07-13
Make the documentation sidebar position fixed.

bradens:
The sidebar in the documentation is always positioned at the top of the document, making navigating the docs extremely difficult.  

I recently used Stylish (a chrome plugin that lets you change styles on pages) to make this change myself, and thought that other people might enjoy it as well.

### DIFF
--- a/pages/src/docs/src/style.less
+++ b/pages/src/docs/src/style.less
@@ -126,6 +126,7 @@
   height: 100%;
   overflow: auto;
   padding: 66px 0 100px;
+  position: fixed;
   width: 240px;
 }
 


leebyron:
:/ The documentation sidebar used to be fixed and people complained loudly that it was extremely difficult to navigate and so it was changed to remain on the page.

I'm happy to entertain this change again, but a main concern is an ability to scroll that navigation when it's longer than the page.

bradens:
Oh, I see.  Well i'll be 1 vote for a fixed sidebar.  I think having the sidebar unreachable while in 99% of the "Pages" of the docs is worse than having to scroll 2 separate containers in the case when the sidebar is larger than the rest of the page.

PR #919, opened 2016-07-13
Update Immutable.d.ts to support TypeScript 2.0

pavadeli:
TypeScript 2.0 provides a new `--strictNullCheck` option that adds non-nullable types. In order to use this option with Immutable.js, we need to fix the parameter-types of closure-parameters like mappers, predicates and groupers in the Immutable.d.ts type-definitions. 

The parameters of these closures are now marked as optional in the type definitions. Because of that, TS thinks that those parameters might not be passed to the closure, so it always adds an explicit `| undefined` to those types. These parameters are always filled though. Parameters to closure-parameters should not be explicitly marked optional.

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -251,11 +251,11 @@
      * @see `Map#mergeWith`
      */
     mergeWith(
-      merger: (previous?: T, next?: T, key?: number) => T,
+      merger: (previous: T, next: T, key: number) => T,
       ...iterables: Iterable.Indexed<T>[]
     ): List<T>;
     mergeWith(
-      merger: (previous?: T, next?: T, key?: number) => T,
+      merger: (previous: T, next: T, key: number) => T,
       ...iterables: Array<T>[]
     ): List<T>;
 
@@ -269,11 +269,11 @@
      * @see `Map#mergeDeepWith`
      */
     mergeDeepWith(
-      merger: (previous?: T, next?: T, key?: number) => T,
+      merger: (previous: T, next: T, key: number) => T,
       ...iterables: Iterable.Indexed<T>[]
     ): List<T>;
     mergeDeepWith(
-      merger: (previous?: T, next?: T, key?: number) => T,
+      merger: (previous: T, next: T, key: number) => T,
       ...iterables: Array<T>[]
     ): List<T>;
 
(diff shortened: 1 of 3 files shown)


facebook-github-bot-8:

facebook-github-bot-0:

pavadeli:
Full non-nullable support needs more work though.

The definitions are currently "waiting" on the closed issue: Microsoft/TypeScript#285. Can we fix the definitions using the special `this` type or do we need to support older TypeScript versions?

pavadeli:
BTW: The current pull request is still backwards compatible. Real non-nullable types support is not backwards compatible unfortunately 😞, so we at least have to wait until TS 2.0 is stable.

Strate:
I think that this change sould not be described as "to support TS 2.0". It should be described as "keep typungs closer to real world", and it could be merged before stabilization of ts 2.0

pavadeli:
You are right. Although I created this PR to fix the most common problem I had with the typings when using TS2.0 with  `--strictNullCheck`, it does make sense without TS2.0.

OliverJAsh:
Amazing, thank you for doing this. I really hope we can get this merged!

ooesili:
TypeScript 2 is out, and `strictNullChecks` has officially landed! Immutable data and strong non-nullable types are an amazing combination, but the current definitions, as mentioned, are causing some pain.

Here is some code of mine that no longer compiles when `strictNullChecks` are turned on:

```typescript
items
  .filter((item) => item.get('type') === filterKey)
  .sortBy((item) => item.get('name'))
```

I get 2 `Object is possibly 'undefined'.` errors refering to the `item.get` calls. This is the best workaround that I could come up with, which is not very good:

```typescript
items
  .filter((item) => (item && item.get('type')) === filterKey)
  .sortBy((item) => item && item.get('name'))
```

Considering how common operations like `.map` and `.filter` are, immutable-js and `strictNullChecks` feel, to me, unusable together at this point. I'm going to give the fork a try for a while and see if I can find some valuable feedback to give.

OliverJAsh:
We'd also gain increased type safety from updating the type definitions:

```
const result = Map<number, number>([[ 1, 2 ]]).get(1)
result // is number, with upgraded type definitions this would be number | undefined
```

Any ideas what is blocking progress on this? How can we help? I would really like to use immutable JS together with the increased type safety brought by `strictNullChecks`.

lacker:
Let's do it

OliverJAsh:
Any updates on when this will be released and ready to use?

Strate:
I think that this library should stop to provide types itself and move types to DefinitelyTyped repository

PR #921, opened 2016-07-14
Better typings for constructors of empty containers.

### LINKED ISSUE #886: Better typings for empty containers
``` typescript
interface Obj {
  prop: Immutable.List<string>
}

let b: Obj = {
  prop: Immutable.List()
}
```

fails with compiler error:

```
error TS2322: Type '{ prop: List<{}>; }' is not assignable to type 'Obj'.
  Types of property 'prop' are incompatible.
    Type 'List<{}>' is not assignable to type 'List<string>'.
      Type '{}' is not assignable to type 'string'.
```

this happens because empty `List` constructor defined as:

``` typescript
export function List<T>(): List<T>;
```

actually, empty list is list of any, and definition should looks like that:

``` typescript
export function List(): List<any>;
```

Adding that case of overloading does the job - compiler no more fails.

What so you think about this? Is it possible to add that case of overloads to each empty collection constructor?

_N.B._ we can not replace `export function List<T>(): List<T>` without breaking BC, we should add a new overload case.

Strate:
Actually, result of consturctor of empty collection could be assigned to collection of any type.
 fixes #886

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -140,6 +140,7 @@
    * Create a new immutable List containing the values of the provided
    * iterable-like.
    */
+  export function List(): List<any>;
   export function List<T>(): List<T>;
   export function List<T>(iter: Iterable.Indexed<T>): List<T>;
   export function List<T>(iter: Iterable.Set<T>): List<T>;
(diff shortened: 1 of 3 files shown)


facebook-github-bot-1:

facebook-github-bot-6:

lacker:
This doesn't seem like something we would want to commit to - what's the problem that this fixes besides sort of an aesthetic completeness? I could be wrong so feel free to object but for now I am closing this.

Strate:
Reason described at #886

Strate:
@lacker this is about reduce code boilerplate/repeat. From type system point of view, empty collection could be assignable to collection fo any type. As empty array could be assignable to array of any type. This PR just fix it. Also:

```ts
interface State {
  list: Collection.List<string>
}

const state: State = {
  list: Collection.List() // fails to compile, compiles only with Collection.List<string>()
}
```

PR #925, opened 2016-07-15
Typescript typings. Add overloads for `map` method to all classes.

Strate:
This is implements overloading for `map` methods for all concrete collection classes,
with proper return type.

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -813,7 +813,7 @@
      *
      */
     map<M>(
-      mapper: (value: V, key: K, iter: /*this*/OrderedMap<k, V>) => M,
+      mapper: (value: V, key: K, iter: /*this*/OrderedMap<K, V>) => M,
       context?: any
     ): OrderedMap<K, M>;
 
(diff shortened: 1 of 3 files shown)


lacker:
What problem is this fixing, exactly? Sorry for asking dumb questions.

Strate:
This make things closer to real world. map method actually returns same classes, with all methods, not base class

lacker:
Is there a `this` thing that works better for this? I feel like I saw a different PR that fixed this differently.

Strate:
No, `this` is not fit to this case. I will show you later

Strate:
consider the example:
```ts
interface BaseCollection<Key, Value> {
  map<U>(mapper: (v: Value) => U): this;
}
interface Map<K, V> extends BaseCollection<K, V> {
}
interface List<V> extends BaseCollection<number, V>{
}

const map = new Map<string, number>;
const mapped = map.map(v => String(v));
```
after this, `mapped` *actually* has a type `Map<string, string>`, but definition says that it has a `Map<string, number>`, because of `this` return type.
You could say that we could just return `BaseCollection<K, U>` from `map` method, and this is also wrong, because returned type actually is a `Map<K, U>`, with all extra methods of `Map` interface.

leebyron:
Updated before committing: some of the overrides were still returning C<T> instead of C<M>

PR #943, opened 2016-08-04
Make `isXXX` functions guard functions.

Strate:


Strate:
Reason to not to merge?

lacker:
No description so I just closed it. Sorry! Are you still interested in this pull request? If so can you describe what's going on and why this is good.

Strate:
Typescript has a `user defined guard functions`, which could be used to narrow type definitions. For example:
```ts
function handleCollection(c: List<any> | Map<string, any>) {
  if (isList(c)) {
    // c shluld be List<any> here
  } else {
    // c shluld be Map<string, any> here
  }
}
```
this is possible only if `isList` function defined as `isList(arg: any): arg is List<any>`. Return type `arg is List<any>` actually boolean, with extra semantic - it says that "if it is true, then arg has a type List<any>"

leebyron:
Sorry for the huge delay! Just kicked Travis to get a passing build and this looks great. Love this new TS feature.

Issue #52, opened 2014-08-10
Produce a diff from immutable objects?


steida:
I would like to use immutable-js as diff for app state in the same manner as React diffs DOM state. Is this even posible? Can you direct me to the proper/better approach? I would like to store app state diff into localStorage and sync with server. Thank you for the awesome library.

slorber:
@steida hi

Have you been successful with json patch?
I'm looking at how json state transitions can be recorded.
This is related to a video I just published: https://www.youtube.com/watch?v=zxN8FYYBcrI

pedroteixeira:
I've using: https://github.com/intelie/immutable-js-diff e https://github.com/intelie/immutable-js-patch

slorber:
thanks @pedroteixeira I didn't know about these projects

Lendar:
Another lib for immutable.js diffs: https://github.com/elierotenberg/remutable — it's more polished, but currently supports only Map.

ActiveObject:
@slorber, you can try [transit-js](https://github.com/cognitect/transit-js). I used it to serialize the history of application state for next transfering over the wire. You can add a cache on top of this library to serialize  immutable data structures efficiently. David Nolan wrote a post about this feature of transit - [Transit-js Caching](http://swannodette.github.io/2015/02/19/transit-js-caching/).

granmoe:
It's been over a year. Any new resources for this? I have a similar need.

jareware:
@granmoe FWIW, I'm using https://github.com/intelie/immutable-js-diff and it works really nicely.

Issue #541, opened 2015-07-08
Make the List "internal nodes" accessible through an API


slorber:
Hi,

If I understand correcly, ImmutableJS use a list that looks like Scala/Clojure vector with internal nodes of size 32.
I think exposing this internal structure may leverage better performances when rendering big immutable list. For example it an item has changed in an immutable list of 100000 elements, one does not have to iterate the 100000 elements and call `shouldComponentUpdate` 100000 times, but rather traverse the ImmutableJs List tree and call `shouldComponentUpdate` for the nodes, permitting to short-circuit the rendering faster and not have to call `shouldComponentUpdate` on all the 100000 items.

I mean if an item is modified in a big array of 100000 elements, the internal node holds an array of 32 references and the item "depth" in the tree is 3, instead of calling `shouldComponentUpdate` 100000 times, you can just call `shouldComponentUpdate` something like 4*32 times with PureRenderMixin.

I don't know how to achieve this in ImmutableJS nor if it's a good idea to use its internal structure if it's not part of the public API, but this could definitively boost performances.

See also related SO question: http://stackoverflow.com/questions/30976722/react-performance-rendering-big-list-with-purerendermixin

slorber:
This React feature seems required also to avoid creating intermediate dom elements when we simply want to optimize the rendering of a flat list: https://github.com/facebook/react/issues/2127

leebyron:
Rather than exposing the internal structure of Immutable.js collections directly, we would probably want some method that is semantically the kind of operation you're proposing and bakes the performance improvements described into the implementation of such a method.

But I'm also not totally sure how what you're proposing would work or how you see it connecting to React.

Can you maybe type some slightly-imaginary future code that describes what you're trying to achieve from the React-side point of view, and maybe that would help me understand what Immutable.js would need to implement to fulfill it?

slorber:
hi @leebyron 

Here's an exemple implementation of what I want to achieve. As you can see, the number of shouldComponentUpdate calls becomes linear to the depth of the immutable structure, instead of being linear to the number of items of the list.


```javascript

/////////////////////////////////////////////////////////////////////
// Here is the "lib" code

var ImmutableListRenderer = React.createClass({
    // Simple implementation, but should use something like PureRenderMixin instead
    shouldComponentUpdate: function(nextProps) {
[...]

PR #953, opened 2016-08-16
Efficient trie diff

errendir:
If there are two `Map` objects deriving from a common `Map` object ancestor, their underlying hash trie can be expected to be very similar. For example `const m1 = Map({...largeObject}); const m2 = m1.set('b', 11); const m3 = m1.set('a',12)` will produce maps `m1`, `m2` and `m3` that share most of the trie.

Entries with the same hashes can be expected to be on the "same" paths in both tries. In Immutable.js each parent-child node relation in a trie is associated with some 5-bit hash fragment. If two nodes in two different tries can be reached from the root of their respective tries by the same 5-bit hash fragments (in the same order), then any entry present in both tries will either be a child of both of those nodes or neither of them. (Property *)

The above two notions can be leveraged to efficiently produce shallow diffs between any trie based datastructures. Such a diff is a collection of added, removed and updated entries representing the changes needed to transform one provided datastructure into the other.

The diff is produced by synchronously traversing both tries. During such a traversal pairs of nodes (one from each trie) are marked as equivalent. The first pair of equivalent nodes are the roots of both tries (since they satisfy Property *). Then for each remaining equivalent pair, the child nodes are extracted together with the hash fragments associated with their parent-child connection. Any two child nodes (from different tries) with the same hash fragments are then marked as equivalent. For each child node with a hash fragment that doesn't exist for any relation in the other trie, all the entries "below" that node are known to be missing in the other trie.

I am aware that the concept of diffing has already bee brought up: https://github.com/facebook/immutable-js/issues/52. As far as I am aware the focus of https://github.com/intelie/immutable-js-diff is not on the efficient shallow diffs. The implementation of this diff algorithm is only possible with an insight into the internal structure of the tries that immutable.js uses. This is why I believe this code belongs here.

I believe this functionality is useful in declaratively built apps (approach encouraged by so many frameworks now). Recomputing data derived from an immutable.js object (for example through grouping objects from a `Map` by a specific attribute) can be computationally expensive. Thanks to the immutability memoization can be used to stop the computation if the result is already known. However if the input object is replaced with a new immutable.js object deriving from the previous one, the result must be recomputed in its entirety. Availability of an efficient diff can enable incremental updates of those derived objects. 

I hope you like my code. I added the `diffFrom` method to the `Map`. It can be easily adapted to the other datastructures. There is a optimization that can be made: The `added` and `removed` `Map`s don't have to be created from scratch. We could reuse the existing trie nodes to build those maps, thus reducing memory used.

Please tell me what you think
Thanks!

### DIFF
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -353,9 +353,9 @@ describe('Map', () => {
     expect(m2.diffFrom(m1).added.toObject()).toEqual({'b': 12})
   })
 
-  it('diffs against the older version of the map providing a map of deleted entries', () => {
+  it('diffs against the older version of the map providing a map of removed entries', () => {
     var m1 = Map({'a': 11, 'b': 12})
-    var m2 = m1.delete('a')
+    var m2 = m1.remove('a')
 
     expect(m2.diffFrom(m1).removed.toObject()).toEqual({'a': 11})
   })
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -346,4 +346,56 @@ describe('Map', () => {
     expect(is(m1, m2)).toBe(true);
   });
 
+  it('diffs against the older version of the map providing a map of added entries', () => {
+    var m1 = Map({'a' : 11})
+    var m2 = m1.set('b', 12)
+
+    expect(m2.diffFrom(m1).added.toObject()).toEqual({'b': 12})
+  })
+
+  it('diffs against the older version of the map providing a map of deleted entries', () => {
+    var m1 = Map({'a': 11, 'b': 12})
+    var m2 = m1.delete('a')
+
+    expect(m2.diffFrom(m1).removed.toObject()).toEqual({'a': 11})
+  })
+
+  it('diffs against the older version of the map providing a map of updated entries', () => {
+    var m1 = Map({'a': 11, 'b': 12})
+    var m2 = m1.set('a', 13)
+
+    expect(m2.diffFrom(m1).updated.toObject()).toEqual({'a': {prev: 11, next: 13}})
+  })
+
+  it('diffs against the older version of the large map', () => {
+    var largeObject = {}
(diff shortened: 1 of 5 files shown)


facebook-github-bot-6:

facebook-github-bot-1:

lacker:
It's not clear that this is worth the added code. Maybe we need some sort of "immutable.js improvement proposal" IIP type process, because I feel bad just saying "meh". But, that said, we have to draw the line somewhere. I am going to close this but if you object then let's discuss because maybe there is something more intelligent to do here.

butchler:
This article explains a contrived example of a use case that I think could benefit from something like this: https://hackernoon.com/an-artificial-example-where-mobx-really-shines-and-redux-is-not-really-suited-for-it-1a58313c0c70

The author of the article uses React components to turn the list of pixels into a hierarchy, but since Immutable already splits of lists into hierarchies, it seems like it would make sense to re-use that logic to find out which components need to be updated. I'm imagining some kind of alternate version of `redux-react` that lets you subscribe to a specific path within the state rather than subscribing to the state as a whole, and uses trie-based diffing like this to efficiently diff the current and next state and find out which subscribers need to be notified.

I'm getting way ahead of myself, but I wonder if something like this could be used to implement a version of MobX that uses immutable data structures, allowing time-travel and such, but is still almost as efficient as using mutable data.

I understand not wanting to put something like this in the core library since most people won't need it, but maybe it could be in a sub-module (e.g. `immutable/trie-diff`).

butchler:
I found an open issue that I think this PR might be a good solution for: https://github.com/facebook/immutable-js/issues/541

Someone requested access to the True internals so they could more efficiently update large lists of React components, but that is obviously fragile because the internals could easily change. I think exposing an efficient diff implementation could be a high-level API that might allow people to do what they wanted to do by inspecting the internals.

Issue #854, opened 2016-04-21
Default import from `immutable`. This module has no default export.


Anahkiasen:
### What happened

I was using Flow with Immutable without problem but since definitions were added I'm getting this error everywhere:

```
src/Services/SchemaConverter.js:2
  2: import Immutable from 'immutable';
            ^^^^^^^^^ Default import from `immutable`. This module has no default export.
```
### How to reproduce

Simply import Immutable and enable Flow on the file

jedwards1211:
I just remembered, I think they wrote the type defs this way because Flow unfortunately doesn't allow a class to be a static property of another class right now.  There are some notes in [immutable.js.flow](https://github.com/facebook/immutable-js/blob/master/type-definitions/immutable.js.flow) that will hopefully wind up in the docs soon.

jedwards1211:
Is there any simple workaround for dealing with the large number of Flow errors about this in preexisting code?  If not, I really don't think the flow defs should be included in the official release until they don't break (cause invalid Flow errors) on preexisting code that has followed the long standing import conventions in the docs.

gf3:
as annoying as this might be for those using the default export, please don't remove the flow type annotations—they're incredibly valuable. i've already been able to fix a couple edge-case bugs in my app because of them!

jedwards1211:
Yes, but the annotations could easily be moved to a well-publicized branch until they're a bit more mature, so that people can `npm install facebook/immutable-js#flowtype` for example

ccorcos:
I created a PR for this fix: https://github.com/facebook/immutable-js/pull/961

leebyron:
Fixed in master will be release soon

PR #961, opened 2016-08-29
Add Default Export to Flow Type definitions

ccorcos:
This PR solves this issue:

https://github.com/facebook/immutable-js/issues/854

In the [Flow documentation](https://flowtype.org/docs/third-party.html#example) the show an example where you define library definitions in the `libs` option. So you may be inclined to use these type declarations by having the following in your flowconfig:

```
[libs]
./node_modules/immutable-js/dist/immutable.js.flow
```

The problem with this approach is that those type definitions are now global. Anytime you define a Map type, its going to defer to the Immutable Map and ever the window.Map type definition.

In a [separate part of the documentation](https://flowtype.org/docs/declarations.html#declaration-files) the talk about delcaration files using the `.js.flow` approach. Now instead of adding the files as a global declaration in libs, you can simply `import immutable from 'immutable'` and everything just works :)

### DIFF
--- a/dist/immutable.js.flow
+++ b/dist/immutable.js.flow
@@ -661,3 +661,22 @@ export {
   fromJS,
   is,
 }
+
+export default {
+  Iterable,
+  Collection,
+  Seq,
+
+  List,
+  Map,
+  OrderedMap,
+  OrderedSet,
+  Range,
+  Repeat,
+  Record,
+  Set,
+  Stack,
+
+  fromJS,
+  is,
+}
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -661,3 +661,22 @@ export {
   fromJS,
   is,
 }
+
+export default {
+  Iterable,
+  Collection,
+  Seq,
+
+  List,
+  Map,
+  OrderedMap,
+  OrderedSet,
+  Range,
+  Repeat,
+  Record,
+  Set,
+  Stack,
+
+  fromJS,
+  is,
+}


facebook-github-bot-8:

ccorcos:
Just signed the contributors license agreement

facebook-github-bot-4:

ccorcos:
I also want to mention that it would be great if we could somehow get some strong typing behind Map and List literals. For example, If I have something like `Immutable.toJS({a: 1, b: 'hello'})` where I could define this as a strict schema that can be validated by the type checker rather than everything be nullable generic Maps.

lacker:
Does this PR still make sense now that we merged https://github.com/facebook/immutable-js/pull/878 ?

kozlitinaelja:
@ccorcos, is the issue addressed in this PR still exists? If no, please, close. 😅

ccorcos:
@kozlitinaelja yes -- the definitions are still declared globally. Therefore, you can only have one Map which refers to the immutable Map and not window.Map.

https://flowtype.org/docs/declarations.html#declaration-files

leebyron:
@ccorcos could you recommend a way to test this change that would otherwise currently fail? I just added a flow type test that ensures the native es6 collections can be used alongside the Immutable.js ones.

ccorcos:
Are you doing `import Immutable from 'Immutable'`?

sghall:
In case anyone else lands here.  Doing...

`import * as Immutable from 'immutable`

is a workaround until there's a release that resolves the issue.  That got rid of the flow errors for me.

Issue #814, opened 2016-03-25
Don't rely on .equals()


georeith:
I have come across an Issue where a library I am using implements `.equals()` on its objects. These are internally crucial to the functioning of the library, however they should not be used by ImmutableJS, just because two items are what the library considers equal doesn't mean I want them to be considered the same by ImmutableJS.

I want these to only be equal if they reference the same object. Is there any way currently to do that? If not this should really be optional as a lot of people using external modules do not have full control over the object prototypes and `.equals()` is going to be fairly prevalent that I imagine a fair few libraries will have side effects with ImmutableJS.

uniphil:
As a counterpoint, I implemented `.equals()` (and friends) in [`results`](https://github.com/uniphil/results) specifically for compatibility with Immutable. It's nice to be able to do that without introducing a dependency on Immutable, but I suppose it could do bad things if you didn't expect it.

arcanis:
It could use a shared symbol rather than a fixed string.

leebyron:
Unfortunately many rely on this behavior today.

In a future version we can consider requiring a shared symbol instead.

georeith:
@leebyron I don't see why it couldn't be optionally disabled. A setting on Immutable, or the ability to specify your own function name for this.

This doesn't have to be a breaking change. But as it stands it makes ImmutableJS unusable with certain other libraries.

elliotbonneville:
Any updates on this? @georeith's solution seems reasonable.

PR #990, opened 2016-10-11
Custom comparator

georeith:
Expose the `compareWith` function that allows you to set a custom function to use for `is()` comparison when comparing non Immutable objects as opposed to `equals()` or setting it to `null` to force strict object equality.

This solves this: https://github.com/facebook/immutable-js/issues/814

And any other situation where you don't want `equals` to be used on your objects. It does not change the behaviour of Immutable collection comparison nor does it cause any breaking changes to existing code. As the tests show.


lacker:
`compareWith` seems like a problematic thing to expose because it works indirectly, it's like it just operates with side effects. At least if I understand it properly. Also maybe this just timed out at this point. I am going to close this pull request but let me know if you think there is still something we should do here.

georeith:
@lacker It is not pure to speak but then again neither is the default of using `.equals()` if it exists on the item. Which is the whole reason this needs to exist because sometimes you are using items that you don't control the `.equals()` function for. You can't override it because X library expects a certain behaviour from it.

This is still not pure with or without `compareWith` its just `compareWith` lets you work around the side effects if you need to. I am happy for another solution but nothing else is being suggested.

In short this is just a fix around the already un pure behaviour which has causes side effects.

elliotbonneville:
I'm with @georeith on this one. `compareWith()`'s current behavior is causing loads of problems for me as well, in terms of working with other libraries which also expose `#equals()` functions.

lacker:
It seems like the right way to do this would be to create a collection with a custom comparator - that's how most languages handle this sort of data structure. Then it's per-collection rather than using `compareWith` impurely.

PR #991, opened 2016-10-13
Add named exports

bspaulding:
Would love to do this:

``` javascript
import { Map, Record } from 'immutable';
```

but at the moment, that's not spec compliant, since immutable only exports a default object.

Thoughts? Pardon if this has been discussed before and I just missed it. Thanks!


lacker:
Looks like this PR has been failing tests for the past year so I am closing it.

PR #1021, opened 2016-12-08
Make header nav links dark gray; increase contrast for usability

flarnie:
I noticed that the contrast between the navigation links and the
background on the Immutable.js landing page is very subtle - by
using higher contrast we can increase usability and accessibility.

screen shot 2016-12-07 at 1 56 57 pm

There is a handy tool for testing contrast for accessibility;
http://webaim.org/resources/contrastchecker/

screen shot 2016-12-07 at 4 06 09 pm

On the landing page the links are white on a light gray background.
We initially set the link color to match the main font color on the
page, which is #626466. However, this was not quite enough contrast.
Bumping the font color a step darker gave us #525455, which should be
very close but have a high enough contrast to be easily readable.

screen shot 2016-12-07 at 4 10 10 pm

screen shot 2016-12-07 at 4 09 04 pm

The dark gray (#525455) on the light blue background was also not high
enough contrast, so we lightened the light blue background.

screen shot 2016-12-07 at 4 26 21 pm

screen shot 2016-12-07 at 4 27 01 pm

screen shot 2016-12-07 at 4 27 12 pm

We considered leaving the links white on the blue background, and only
changing them on the landing page where the background is light gray.
However, to get enough contrast with a white font the light blue
background would have to be darkened beyond recognition.

screen shot 2016-12-07 at 4 52 22 pm

I'm open to adjusting this as needed, and thanks for your review!

More ideas:
- We could make the main copy font color match the nav links
- We could use a dark blue instead of dark gray for the nav links

PS - I don't know how to view the website locally, so I was just changing colors in the dev tools to test the effect. lmk if I can spin up the docs site locally to manually test this change.

### DIFF
--- a/pages/src/src/style.less
+++ b/pages/src/src/style.less
@@ -27,7 +27,7 @@
 }
 
 .miniHeader {
-  background: #6dbcdb;
+  background: #9ed2e7;
   position: fixed;
   width: 100%;
   z-index: 1;
@@ -63,7 +63,7 @@
 }
 
 .miniHeaderContents a {
-  color: #fff;
+  color: #525455;
   font-weight: bold;
   margin-right: 1em;
   text-decoration: none;


lacker:
It would be neat if you could run the docs locally, and then take a before & after screen shot so it's clear what happened. I think https://github.com/facebook/immutable-js/blob/master/gulpfile.js has the stuff to run the docs, maybe just running `gulp dev` works to see that.

leebyron:
Hey @flarnie thanks for digging into this!

In order to run locally, make sure you have `gulp` installed then run `gulp dev`.

I noticed in your screenshots that the text shadow is still quite dark, which makes the text fuzzy and harder to read.

I originally put the shadow on the white text to increase contrast against the "immutable blue" header color, which gives an appropriate 3:1 contrast ratio for the typical case.

Perhaps we could do the opposite for boosting contrast? Instead of making the text gray and lightening the blue, we could instead make the splash screen background a slightly darker gray or give a slightly darker shadow on the white link text. Ideally we could make that shadow darker specifically for the gray wash case without much disruption to the blue background case

leebyron:
I just added a commit which boosts contrast by deepening the shadows around those links and darkening the background

PR #1024, opened 2016-12-19
Add named exports

bspaulding:
Would love to do this:

```javascript
import { Map, Record } from 'immutable';
```

but at the moment, that's not spec compliant, since immutable only exports a default object.

Thoughts? Pardon if this has been discussed before and I just missed it. Thanks!

-----

Originally opened as #991, which got closed without discussion. Tests should be passing now.

I've checked in the rebuilt `dist` changes, as the contributing guide says to do, but this is creating a merge conflict. Please advise if I should revert.


wokalski:
I can't see why it would create a git conflict. Could  you try the following:
1. Add `https://github.com/facebook/immutable-js.git` as your upstream.
2. Rebase your changes on top of master branch in the upstream
3. Push it here

It fails because your changes are applied on top of an old commit.
zrzut ekranu 2016-12-20 o 21 13 37

bspaulding:
Ah, duh. :) Updated..

wokalski:
Looks good to me.

@lacker verify and merge.

Issue #1015, opened 2016-11-25
OrderedMap flow type inheritance problem


rt2zz:
### What happened
OrderedMap inherits its type from Map, which means the return value of all OrderedMap methods are of type Map.

One solution is to not inherit OrderedMap from Map and instead redefine all methods. I am hoping there is a more elegant solution using a `this` type, although a cursory attempt failed because this cannot be parameterized.

### How to reproduce
```js
[1, 2, 3]
  .reduce((acc: OrderedMap<number, number>, num: number) => {
    return acc.set(num, num)
  }, OrderedMap())
```
flow error:
```
Error:
  186:   }, OrderedMap())
            ^^^^^^^^^^^^ Map. This type is incompatible with
  184:   .reduce((acc: OrderedMap<number, number>, num: number) => {
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^ OrderedMap
```

wokalski:
I *think* it's impossible to express with `flow` now. I added an [issue](https://github.com/facebook/flow/issues/2959).

howardjing:
I ran into the same issue with `OrderedMap` / `OrderedSet` / `toOrderedMap` / `toOrderedSet`. The easiest thing I can think of to fix this is to copy paste the typings of `Map` and `Set`, and adjust return values appropriately. Will have a PR up shortly.

PR #1027, opened 2016-12-20
Fix flowtypes for OrderedSet, OrderedMap, toOrderedSet, and toOrderedMap

howardjing:
Unfortunately extending Set and Map caused issues with OrderedSet
and OrderedMap flow annotations. For example,

```
const orderedSet: OrderedSet<number> = OrderedSet.of(1,2,3);
```

was not valid, because `of` was incorrectly assumed to return a
`Set<number>` rather than an `OrderedSet<number>`.

The fix was to copy and paste the Set and Map flow annotations,
adjusting for OrderedSet and OrderedMap as appropriate.

See #1015 for more details.

----

Add negative flow test cases for OrderedSet and OrderedMap

One thing that came up as a result of this is that it seems
like the current flow definitions are laxer than typescript
definitions. For example, typescript's intersect looks like

```
intersect(...iterables: Array<T>[]): Set<T>
```

whereas flow's intersect looks like

```
intersect<U>(...iterables: ESIterable<U>[]): Set<T&U>
```

Also removed unnecessary OrderedSet overrides such as

  * delete
  * remove
  * clear

since we were already extending Set.

Also removed unused variables in immutable-flow tests.

### DIFF
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -561,9 +561,78 @@ declare module 'immutable' {
     flatten(shallow?: boolean): /*this*/Map<any,any>;
   }
 
-  // OrderedMaps have nothing that Maps do not have. We do not need to override constructor & other statics
-  declare class OrderedMap<K,V> extends Map<K,V> {
+  declare class OrderedMap<K,V> extends KeyedCollection<K,V> {
+    static <K, V>(obj?: {[key: K]: V}): OrderedMap<K, V>;
+    static <K, V>(iterable: ESIterable<[K,V]>): OrderedMap<K, V>;
     static isOrderedMap(maybeOrderedMap: any): bool;
+
+    set<K_, V_>(key: K_, value: V_): OrderedMap<K|K_, V|V_>;
+    delete(key: K): this;
+    remove(key: K): this;
+    clear(): this;
+
+    update<K_,V_>(updater: (value: this) => OrderedMap<K_,V_>): OrderedMap<K_,V_>;
+    update<V_>(key: K, updater: (value: V) => V_): OrderedMap<K,V|V_>;
+    update<V_>(key: K, notSetValue: V_, updater: (value: V) => V_): OrderedMap<K,V|V_>;
+
+    merge<K_,V_>(
+      ...iterables: (ESIterable<[K_,V_]> | { [key: K_]: V_ })[]
+    ): OrderedMap<K|K_,V|V_>;
+
+    mergeWith<K_,W,X>(
+      merger: (previous: V, next: W, key: number) => X,
+      ...iterables: ESIterable<W>[]
+    ): OrderedMap<K,V|W|X>;
+
+    mergeDeep<K_,V_>(
+      ...iterables: (ESIterable<[K_,V_]> | { [key: K_]: V_ })[]
+    ): OrderedMap<K|K_,V|V_>;
+
+    mergeDeepWith<K_,W,X>(
+      merger: (previous: V, next: W, key: number) => X,
+      ...iterables: ESIterable<W>[]
+    ): OrderedMap<K,V|W|X>;
+
+    setIn(keyPath: ESIterable<any>, value: any): OrderedMap<K,V>;
+    deleteIn(keyPath: ESIterable<any>, value: any): this;
+    removeIn(keyPath: ESIterable<any>, value: any): this;
+
+    updateIn(keyPath: ESIterable<any>, notSetValue: any, value: any): OrderedMap<K,V>;
+    updateIn(keyPath: ESIterable<any>, value: any): OrderedMap<K,V>;
+
+    mergeIn(
+      keyPath: ESIterable<K>,
+      ...iterables: (ESIterable<K> | { [key: string]: any })[]
+    ): OrderedMap<K,any>;
+    mergeDeepIn(
+      keyPath: ESIterable<K>,
+      ...iterables: (ESIterable<K> | { [key: string]: any })[]
+    ): OrderedMap<K,any>;
+
+    withMutations(mutator: (mutable: this) => any): this;
+    asMutable(): this;
+    asImmutable(): this;
+
+    // Overrides that specialize return types
+    map<V_>(
+      mapper: (value: V, key: K, iter: this) => V_,
+      context?: any
+    ): OrderedMap<K,V_>;
+
+    flatMap<K_,V_>(
+      mapper: (value: V, key: K, iter: this) => ESIterable<[K_,V_]>,
+      context?: any
+    ): OrderedMap<K_,V_>;
+
+    flip(): OrderedMap<V,K>;
+
+    mapKeys<K_>(
+      mapper: (key: K, value: V, iter: this) => K_,
+      context?: any
+    ): OrderedMap<K_,V>;
+
+    flatten(depth?: number): /*this*/OrderedMap<any,any>;
+    flatten(shallow?: boolean): /*this*/OrderedMap<any,any>
   }
 
   declare class Set<T> extends SetCollection<T> {
@@ -602,9 +671,40 @@ declare module 'immutable' {
     flatten(shallow?: boolean): /*this*/Set<any>;
   }
 
-  // OrderedSets have nothing that Sets do not have. We do not need to override constructor & other statics
   declare class OrderedSet<T> extends Set<T> {
+    static <T>(iterable: ESIterable<T>): OrderedSet<T>;
+    static of<T>(...values: T[]): OrderedSet<T>;
+    static fromKeys<T>(iter: ESIterable<[T,any]>): OrderedSet<T>;
+    static fromKeys<K, V>(object: { [key: K]: V }): OrderedSet<K>;
+    static (_: void): OrderedSet<any>;
     static isOrderedSet(maybeOrderedSet: any): bool;
+
+    add<U>(value: U): OrderedSet<T|U>;
+    delete(value: T): this;
+    remove(value: T): this;
+    clear(): this;
+    union<U>(...iterables: ESIterable<U>[]): OrderedSet<T|U>;
+    merge<U>(...iterables: ESIterable<U>[]): OrderedSet<T|U>;
+    intersect<U>(...iterables: ESIterable<U>[]): OrderedSet<T&U>;
+    subtract<U>(...iterables: ESIterable<U>[]): OrderedSet<T>;
+
+    withMutations(mutator: (mutable: this) => any): this;
+    asMutable(): this;
+    asImmutable(): this;
+
+    // Overrides that specialize return types
+    map<M>(
+      mapper: (value: T, value: T, iter: this) => M,
+      context?: any
+    ): OrderedSet<M>;
+
+    flatMap<M>(
+      mapper: (value: T, value: T, iter: this) => ESIterable<M>,
+      context?: any
+    ): OrderedSet<M>;
+
+    flatten(depth?: number): /*this*/OrderedSet<any>;
+    flatten(shallow?: boolean): /*this*/OrderedSet<any>;
   }
 
   declare class Stack<T> extends IndexedCollection<T> {
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -81,9 +81,9 @@ declare module 'immutable' {
     toArray(): V[];
     toObject(): { [key: string]: V };
     toMap(): Map<K,V>;
-    toOrderedMap(): Map<K,V>;
+    toOrderedMap(): OrderedMap<K,V>;
     toSet(): Set<V>;
-    toOrderedSet(): Set<V>;
+    toOrderedSet(): OrderedSet<V>;
     toList(): List<V>;
     toStack(): Stack<V>;
     toSeq(): Seq<K,V>;
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -671,6 +671,7 @@ declare module 'immutable' {
     flatten(shallow?: boolean): /*this*/Set<any>;
   }
 
+  // Overrides except for `isOrderedSet` are for specialized return types
   declare class OrderedSet<T> extends Set<T> {
     static <T>(iterable: ESIterable<T>): OrderedSet<T>;
     static of<T>(...values: T[]): OrderedSet<T>;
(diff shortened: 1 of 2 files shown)


howardjing:
If there's a way to fix this without copy pasting flow annotations for Map / Set, I'm happy to update this PR as necessary. I made this PR because with immutable 3.8.1 I couldn't use `list.toOrderedSet` with flow.

type-definitions/immutable.js.flow
```diff
@@ -561,9 +561,78 @@ declare module 'immutable' {
     flatten(shallow?: boolean): /*this*/Map<any,any>;
   }
 
-  // OrderedMaps have nothing that Maps do not have. We do not need to override constructor & other statics
-  declare class OrderedMap<K,V> extends Map<K,V> {
+  declare class OrderedMap<K,V> extends KeyedCollection<K,V> {
```
wokalski: Why did you decide to extend `KeyedCollection`? `OrderedMap` [extends `Map`](https://github.com/facebook/immutable-js/blob/master/src/OrderedMap.js#L17).

type-definitions/tests/immutable-flow.js
```diff
@@ -268,6 +272,41 @@ stringToNumber = Map({'a': 1}).mapKeys((key, value, iter) => 1)
...
+orderedStringToNumber = OrderedMap({'a': 1}).withMutations(mutable => mutable.set('b', 2))
+orderedStringToNumber = OrderedMap({'a': 1}).asMutable();
+orderedStringToNumber = OrderedMap({'a': 1}).asImmutable();
+orderedStringToNumber = OrderedMap({'a': 1}).map((v) => v + 1)
+orderedStringToNumber = OrderedMap({'a': 1}).flatMap((v, k) => (OrderedMap({ [k]: v + 1 })))
+orderedStringToNumber = OrderedMap({'a': 1}).flip().flip()
+orderedStringToNumber = OrderedMap({'a': 1}).mapKeys(x => x)
+orderedStringToNumber = OrderedMap({'a': 1}).flatten()
+orderedStringToNumber = OrderedMap({'a': 1}).flatten(1)
+orderedStringToNumber = OrderedMap({'a': 1}).flatten(true)
```
wokalski: Thanks for the tests! It would be great if there were tests for expected errors, too. You can do it with `// $ExpectError`. See [here](https://github.com/facebook/immutable-js/pull/1027/files#diff-167148588528b5fa451dc535e9c29037R399`).

wokalski:
Good jobs. I think it's good to merge when there are tests for false positives (`$ExpectError`)

type-definitions/immutable.js.flow
howardjing: Ah, didn't realize that. I extendedKeyedCollection because `Immutable.Map.isMap` is a function whereas `Immutable.OrderedMap.isMap` doesn't seem to be a function:

image

I'll make the change for this and `OrderedSet`.

howardjing:
I'm currently working on false positives, will update this PR tonight or tomorrow. I've found a few edge cases that I'll add TODOs / FIXMEs with explanations.

type-definitions/immutable.js.flow
howardjing: Turns out I'm getting an issue overriding `update` by extending Map.

image

I'll leave it as extending `KeyedCollection` for the purposes of this PR. Having `OrderedSet` extend `Set` worked without issues (I didn't realize I was already doing that), and I'll update the PR to remove extraneous OrderedSet definitions.

howardjing:
Updated PR with negative test cases, and removed some unnecessary variables / overrides.

howardjing:
The flow definitions seems to be a little laxer than the typescript definitions, I'm not sure if this is by design, so I left it alone.

wokalski:
Out of curiosity - what do you exactly mean by laxer? 

They are verbose, it's true. I don't have any experience with TS so I cannot compare. `immutable` is a bit special case because of generics and inheritance hierarchy.

type-definitions/immutable.js.flow
wokalski: I think it's because of `this` in the `update` definition. Does it happen if you replace `this` on line 574 with `OrderedMap<K,V>` and with `Map<K,V> on line 580?

type-definitions/immutable.js.flow
```diff
@@ -680,19 +681,11 @@ declare module 'immutable' {
     static isOrderedSet(maybeOrderedSet: any): bool;
 
     add<U>(value: U): OrderedSet<T|U>;
-    delete(value: T): this;
```
wokalski: nice!

type-definitions/tests/immutable-flow.js
```diff
@@ -5,7 +5,7 @@
 // Some tests look like they are repeated in order to avoid false positives.
 // Flow might not complain about an instance of (what it thinks is) T to be assigned to T<K, V>
 
-import Immutable, {
```
wokalski: Why is it deleted?

type-definitions/tests/immutable-flow.js
```diff
@@ -173,6 +160,8 @@ stringToNumber = Map({'a': 'a'})
 stringToNumber = Map([['a', 1]])
 // $ExpectError
 stringToNumber = Map([['a', 'b']])
+// FIXME: this should trigger an error -- this is actually a Map<string, string>
+stringToNumber = Map(List.of(List(['a', 'b'])))
```
wokalski: Great find! Would you mind adding an issue for it? FIXMEs get lost pretty easily.

type-definitions/tests/immutable-flow.js
```diff
@@ -212,6 +201,14 @@ stringToNumber = Map({a: 1}).merge([[1, 'a']])
...
 
+/**
+ * FIXME: this should throw an error, the signature should be
+ *
+ * ```
+ * merger: (previous: V, next: V, key: K) => V,
+ *    ...iterables: Iterable<K, V>[]
+ * ```
+ */
 stringToNumber = Map({'a': 1}).mergeWith((previous, next, key) => 1, [1])
```
wokalski: I don't understand why it should throw an error :(.

type-definitions/tests/immutable-flow.js
```diff
@@ -275,37 +272,157 @@ anyMap = Map({'a': 1}).flatten()
...
+ * ```
+ * const x: OrderedMap<string, string> = OrderedMap({'a': 1})
+ *   .update(() => OrderedMap({'b': '1'}))
+ * ```
+ *
+ * In the above example, `update` is changing an OrderedMap<string, number> to an OrderedMap<string, string>
+ * This seems inconsistent with the typescript signature of
+ *
+ * ```
+ * update(updater: (value: Map<K, V>) => Map<K, V>): Map<K, V>
```
wokalski: Thanks, great find again. Would you mind adding an issue (again)?

type-definitions/immutable.js.flow
```diff
@@ -561,9 +561,78 @@ declare module 'immutable' {
...
+    update<K_,V_>(updater: (value: this) => OrderedMap<K_,V_>): OrderedMap<K_,V_>;
+    update<V_>(key: K, updater: (value: V) => V_): OrderedMap<K,V|V_>;
+    update<V_>(key: K, notSetValue: V_, updater: (value: V) => V_): OrderedMap<K,V|V_>;
+
+    merge<K_,V_>(
+      ...iterables: (ESIterable<[K_,V_]> | { [key: K_]: V_ })[]
+    ): OrderedMap<K|K_,V|V_>;
+
+    mergeWith<K_,W,X>(
+      merger: (previous: V, next: W, key: number) => X,
```
wokalski: Shouldn't key be of type `K` here? And `K_` in the output?

type-definitions/immutable.js.flow
howardjing: Unfortunately the updater function is still problematic. I can work around it by changing the update signature from

```
update<K_,V_>(updater: (value: Map<K,V>) => Map<K_,V_>): Map<K_,V_>;
```

to

```
update<K_,V_>(updater: (value: any) => any): Map<K_,V_>;
```

but I'm not sure how else to fix it. The error being generated is as follows:

```
inconsistent use of library definitions
571:     update<K_,V_>(updater: (value: OrderedMap<K,V>) => OrderedMap<K_,V_>): OrderedMap<K_,V_>;
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ intersection. This type is incompatible with
500:     update<K_,V_>(updater: (value: Map<K,V>) => Map<K_,V_>): Map<K_,V_>;
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ polymorphic type: function type
  Member 1:
  571:     update<K_,V_>(updater: (value: OrderedMap<K,V>) => OrderedMap<K_,V_>): OrderedMap<K_,V_>;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ polymorphic type: function type
  Error:
  inconsistent use of library definitions
  500:     update<K_,V_>(updater: (value: Map<K,V>) => Map<K_,V_>): Map<K_,V_>;
                                                       ^^^^^^^^^^ Map. This type is incompatible with
  571:     update<K_,V_>(updater: (value: OrderedMap<K,V>) => OrderedMap<K_,V_>): OrderedMap<K_,V_>;
                                                              ^^^^^^^^^^^^^^^^^ OrderedMap
  Member 2:
  572:     update<V_>(key: K, updater: (value: V) => V_): OrderedMap<K,V|V_>;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ polymorphic type: function type
  Error:
  inconsistent use of library definitions
  500:     update<K_,V_>(updater: (value: Map<K,V>) => Map<K_,V_>): Map<K_,V_>;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ undefined (too few arguments, expected default/rest parameters). This type is incompatible with
  572:     update<V_>(key: K, updater: (value: V) => V_): OrderedMap<K,V|V_>;
                                       ^^^^^^^^^^^^^^^^ function type
  Member 3:
  573:     update<V_>(key: K, notSetValue: V_, updater: (value: V) => V_): OrderedMap<K,V|V_>;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ polymorphic type: function type
  Error:
  inconsistent use of library definitions
  500:     update<K_,V_>(updater: (value: Map<K,V>) => Map<K_,V_>): Map<K_,V_>;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ undefined (too few arguments, expected default/rest parameters). This type is incompatible with
  573:     update<V_>(key: K, notSetValue: V_, updater: (value: V) => V_): OrderedMap<K,V|V_>;
                                                        ^^^^^^^^^^^^^^^^ function type
```

type-definitions/immutable.js.flow
howardjing: I think that makes sense to me -- this was copy pasted from `Map` so this issue exists there as well. The other thing to note is that `iterables` should probably be `ESIterable<K_,W>` rather than `ESIterable<W>[]`.

type-definitions/tests/immutable-flow.js
howardjing: `Immutable` wasn't actually being used, so I got rid of it.

type-definitions/tests/immutable-flow.js
howardjing: Yes, I'll create an issue with a list of flowtype wonkiness I've found.

howardjing:
To clarify on 

> The flow definitions seem to be a little laxer than the typescript definitions

I mean that functions that functions that make persistent changes to the data structure in flow are allowed to introduce new types, whereas they are not allowed to introduce new types in typescript. For example, this is the `Set#add` function in flow:

```
class Set<T> {
  add<U>(value: U): Set<T|U>;
}
```

and this is the `Set#add` function in typescript:

```
interface Set<T> {
  add(value: T): Set<T>;
}
```

So we can do the following in flow

```
Set([1]).add('a');
```

but not in typescript. This seems like an inconsistency that I thought I'd bring up, and probably belongs in a separate issue.

howardjing:
I made a gigantic issue here: https://github.com/facebook/immutable-js/issues/1029, let me know if you want me to split it up into mini issues.

type-definitions/tests/immutable-flow.js
wokalski: Ah! Gotcha.

type-definitions/immutable.js.flow
wokalski: Extending `KeyedCollection` is better IMO.

type-definitions/tests/immutable-flow.js
wokalski: It tested valid imports. I know it might be confusing though. I should've put a comment there.

wokalski:
Great job. The only remaining issue are the `mergeWith` methods. Do you want to fix them here or should I or you create an issue for it? Other than that - thanks for helping all of us with this contribution.

type-definitions/tests/immutable-flow.js
howardjing: Interesting -- can you elaborate on that? I'll switch back to 

```
import Immutable, {
  // ...
} from 'immutable';
```

but I'm not clear as to what importing`Immutable` is testing. Should I add back 

```
import * as Immutable2 from 'immutable'

const ImmutableList = Immutable.List
const ImmutableMap = Immutable.Map
const ImmutableStack = Immutable.Stack
const ImmutableSet = Immutable.Set
const ImmutableKeyedIterable = Immutable.KeyedIterable
const ImmutableRange = Immutable.Range
const ImmutableRepeat = Immutable.Repeat
const ImmutableIndexedSeq = Immutable.IndexedSeq

const Immutable2List = Immutable2.List
const Immutable2Map = Immutable2.Map
const Immutable2Stack = Immutable2.Stack
const Immutable2Set = Immutable2.Set
const Immutable2KeyedIterable = Immutable2.KeyedIterable
const Immutable2Range = Immutable2.Range
const Immutable2Repeat = Immutable2.Repeat
const Immutable2IndexedSeq = Immutable2.IndexedSeq
```

as well?

howardjing:
The `mergeWith` methods are included in #1029, but I'm working on that now and will update this PR with those changes.

type-definitions/tests/immutable-flow.js
wokalski: I made it - I know it's confusing. It tests if (1) the default import (2) the `import * as X from 'immutable'` (3) individual imports are supported.

howardjing: Cool, I'll add that back and include a comment as to what it's testing. Thanks for the clarification.

howardjing:
Updated with commits to add back the Import tests, and fix mergeWith and mergeDeepWith type signatures for Maps and OrderedMaps. I'm currently working on fixing `updateIn` for Maps / OrderedMaps in a separate commit.

howardjing:
Updated with fix to `updateIn`

wokalski:
@lacker I think it's good to merge.
@howardjing great job. Thanks for being so precise. Sorry for the nitpicking!

PR #1032, opened 2017-01-04
fix for typescript error

xaviergonz:
Fixes the following typescript error:
All declarations of merged declaration 'Indexed' must be exported or not exported.

Typescript 2.1.4


leebyron:
Thanks for the fix!

PR #1035, opened 2017-01-08
Add '| undefined' to TypeScript definitions, which can return undefined

astashov:
TypeScript supports non nullable types (optionally), and it's very
helpful to know when some function or method could return `undefined` as
well as a valid result.

This PR adds `| undefined` to the return types of the methods, which can
return `undefined` (like, `get`, `find`, etc).

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -141,6 +141,9 @@
      * ```js
      * List.of(1, 2, 3, 4).toJS();
      * // [ 1, 2, 3, 4 ]
+     *
+     * List.of({x:1}, 2, [3], 4).toJS();
+     * // [ { x: 1 }, 2, [ 3 ], 4 ]
      * ```
      */
     function of<T>(...values: T[]): List<T>;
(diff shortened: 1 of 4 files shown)


type-definitions/Immutable.d.ts
```diff
@@ -2733,7 +2733,7 @@ declare module Immutable {
    * @ignore
    */
   export interface Iterator<T> {
-    next(): { value: T; done: boolean; }
+    next(): { value: T; done: boolean; } | undefined
```
leebyron: Iterators will always return a value. Did you mean `value: T | undefined`?

astashov: Yes, sorry.

astashov: Fixed

leebyron:
The next issue this is exposing is with generating the documentation. I've worked in a solution and will merge manually.

astashov:
Hooray!!! Thanks!

dist/immutable-nonambient.d.ts
```diff
@@ -1928,7 +1940,7 @@
      * so if `notSetValue` is not provided and this method returns `undefined`,
      * that does not guarantee the key was not found.
      */
-    get(key: K, notSetValue?: V): V;
+    get(key: K, notSetValue?: V): V | undefined;
```
montanonic: I would like to suggest we change this to the following:

```
get(key: K): V | undefined;
get(key: K, notSetValue: V): V;
```

If you've provided a `notSetValue`, this should only ever return `undefined` if `V : undefined`. The current iteration therefore demands unnecessary null-handling, if I'm not mistaken.

leebyron: Great suggestion

PR #1037, opened 2017-01-10
Add missing characters to end of sentence in documentation

Carlsson87:
I noticed that the documentation for the `insert`-method was missing a few characters. So I fixed it.

PR #1038, opened 2017-01-10
Build with the changes from pull request 1027

Carlsson87:
I noticed this while making #1037. It seems like the changes that were merged in #1027 never got built.


leebyron:
Thanks :)

PR #1041, opened 2017-01-13
export hash function

UnsungHero97:
It would be useful to have access to the hash function in the case where one needs to ensure equality in custom objects. For example, if wanted to use `Immutable.Set()` to store a custom object, the set would contain duplicates:

```
class Principal {
  constructor(type, id) {
    this.type = type;
    this.id = id;
  }
}

const principals = [
  new Principal('type_0', 'id_0'),
  new Principal('type_0', 'id_0')
];

const principalsSet = Immutable.Set().withMutations((set) => {
  principals.forEach((principal) => {
    set.add(principal);
  });
});

principalsSet.size; // 2, but intend for it to be 1
```

In order for `Immutable.Set` to drop duplicate `Principal` objects, I would need to additionally implement `valueOf`, or both `equals` and `hashCode`, on the `Principal` class. Having access to `Immutable.hash()` would make this much easier.

I'm exposing the internal hash function along with the main `Immutable` API, however, it's worth considering exporting this as a standalone utility function.

### DIFF
--- a/src/Immutable.js
+++ b/src/Immutable.js
@@ -20,6 +20,7 @@ import { Range } from './Range'
 import { Repeat } from './Repeat'
 import { is } from './is'
 import { fromJS } from './fromJS'
+import { hash } from './Hash'
 import { Iterable } from './IterableImpl'
 
 
@@ -41,7 +42,8 @@ export default {
   Repeat: Repeat,
 
   is: is,
-  fromJS: fromJS
+  fromJS: fromJS,
+  hash: hash
 
 };
 
@@ -62,5 +64,6 @@ export {
   Repeat,
 
   is,
-  fromJS
+  fromJS,
+  hash
 }
--- a/src/Immutable.js
+++ b/src/Immutable.js
@@ -46,21 +46,21 @@ export default {
 };
 
 export {
-	Iterable,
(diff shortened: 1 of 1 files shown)


leebyron:
I agree with you about the utility of this. I'll merge this for now, but we certainly should consider a separate module in the future

UnsungHero97:
@leebyron I think that makes a lot of sense. Do you think this is should be extracted as another Facebook project, or would you consider relying on a 3rd party dependency, say if I were to create a new project with the current implementation of `Immutable.hash()` as the starting point?

leebyron:
I'd prefer not to rely on 3rd party dependencies - I'll likely consider setting up a multi-package distribution in the future that allows subsets of Immutable to be used in isolation

UnsungHero97:
@leebyron any idea when there will be another `immutable-js` release? `v3.8.1` was almost a year ago, and there have been a lot of changes since then.

leebyron:
I plan on getting to another release sometime before 2019. 

Just kidding, I'm cooking up a new release in the next few days

UnsungHero97:
A year later, the initial "sometime before 2019" estimate seems reasonable :)

PR #1047, opened 2017-01-27
Change http to https in docs

styfle:
I noticed a lot of URLs were still on the old http which is no longer cool.
So I updated to https since it's 2017 :tada:

I also fixed a couple dead links for TypeScript 💯


leebyron:
https all the things!

Issue #985, opened 2016-09-30
Immutable.Record keys not enforced consistently on initialization and set


jcarbo:
### What happened

Record keys seem to be enforced only on `set`, not during initialization (see steps below). I'd prefer it to not throw and/or you could configure whether or not it throws an error, but I found the inconsistency weird. Here's my use-case:

I'm using Records to store data returned from an API in a normalized store. If a record doesn't exists, I create one. If one does exist, I merge the data into the existing record. If the API has added keys that my client app is not aware of, on initialization everything works (just ignores the keys it's not aware of) but on the merge will throw.

As per [this dicussion (#573)](https://github.com/facebook/immutable-js/issues/573#issuecomment-131736533), the commenter says:

> Records are only meant to be used if you want to enforce a certain schema for your data while providing defaults when removing a key.

I'd disagree. For me, the enforcement of keys has actually caused more headache than benefit. I use Records so that I have the benefits of immutable data while still being able to use `.` access and destructure in a sane way throughout my app without needing to call `get` everywhere.

Thoughts?
### How to reproduce

``` node
$ node
> var Immutable = require('immutable')
undefined
> var MyRecord = Immutable.Record({ foo: undefined })
undefined
> var record = new MyRecord({ foo: 5, bar: 10 })
undefined
> record.toJS()
{ foo: 5 }
> record.set('bar', 10)
Error: Cannot set unknown key "bar" on Record
    at Record.set (/Users/jeffcarbonella/Documents/contactually/frontend/node_modules/immutable/dist/immutable.js:3690:15)
    at repl:1:8
    at sigintHandlersWrap (vm.js:22:35)
    at sigintHandlersWrap (vm.js:96:12)
    at ContextifyScript.Script.runInThisContext (vm.js:21:12)
    at REPLServer.defaultEval (repl.js:313:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:513:10)
    at emitOne (events.js:101:20)
```

volkanunsal:
I just declare all the keys I need upfront. It forces me to think hard about the schema and you can build a simple helper function very easily.

```js
> const makeSchema = (...rest: string[]) = rest.reduce((s, v) => Object.assign(s, {[v]: undefined}), {})
> var MyRecord = Immutable.Record(makeSchema('foo', 'bar'))
[...]

PR #1048, opened 2017-01-27
Record set not throw error anymore for consistency with constructor

Freddy03h:
Following the issue #985
I'm testing this behavior on my project to deal with `merge` and `Record` on a redux store.

The goal is to not throw error when `set`or `merge` unexisting attribut(s) on Record to keep consistency with constructor, but also because it's hard to deal with `merge` for `Record` object.

```js
const MyRecord = Immutable.Record({ foo: null })
const record = new MyRecord({ foo: 5, bar: 10 })

record.toJS() // { foo: 5 }
record.set('bar', 10).toJS() // Error: Cannot set unknown key "bar" on Record
record.merge({ foo: 7, bar: 11 }).toJS() // Error: Cannot set unknown key "bar" on Record
record.remove('bar').toJS() // { foo: 5 }
```

Without this change, it's not easy to set new attributs from a data object. And it's useless to transform the data to `Record` before merging, because merging two `Record` mean only keep the second, without anything from the first.

```js
const MyRecord = Immutable.Record({ foo: null, bar: null })
const record1 = new MyRecord({ foo: 5, bar: 10 })
const record2 = new MyRecord({ foo: 8 })

record1.merge(record2).toJS() // { foo: 8, bar: null }
record2.toJS() // { foo: 8, bar: null }
```

(ps: I don't know why there is so much change in `immutable.js.flow`)


leebyron:
I think the error with `set` still makes sense. It protects against misuse. The desire to have merge behave similarly to constructor makes sense. It would be great to limit this change to merge.

Freddy03h:
• Why protect against misuse on `set` and `merge` but not on `constructor` and `remove` ?

• Why have a different behavior between `merge` and `set`, when currently, `merge` uses the Record's `set` function because this is the behavior on `Map`. I think having a different behavior between `set` and `merge` has more impact than changing the Record's `set.`

leebyron:
>  Why protect against misuse on set and merge but not on constructor and remove ?

I think `set` and `remove` should have similar behavior and `merge` and `constructor` should have similar behavior.

Specifically, I think it's appropriate for errors to be thrown for misuse of `set` or `remove` because they're indicative of mistaken usage. `record.set('unknownkey', 'value')` would be a no-op and typos would be likely to blame for a difficult to debug issues. However I agree `merge` should work similarly to the constructor and allow for applying a merged set that may contain more keys than the record itself does.

leebyron:
Thinking about this more, I think making a breaking change to `remove` by adding throwing behavior could be inappropriate for the next immediate release and so perhaps dropping the error behavior is a reasonable backup plan - and ideally we can use type checking systems like typescript and flow to detect the sort of accidental misuse of set and remove.

PR #1053, opened 2017-02-07
Fix flowtypes for Iterable#map()

izumin5210:
I found that the definition of `Iterable#map` is missed in `immutable.js.flow`. But `Immutable.d.ts` has the definition for this method.

https://github.com/facebook/immutable-js/blob/29a5c55727c523b0d83c53c7ccd1032aaa579f49/type-definitions/Immutable.d.ts#L2155-L2158


izumin5210:
Sorry, I has been a misunderstanding :bow:

PR #1055, opened 2017-02-10
'npm build' to add flow updates to dist

mjmahone:


mjmahone:
We should just merge pull request 1038: @carlsson87 got to it before I did.

Issue #1050, opened 2017-02-04
Immutable.fromJS incorrectly handles object with "constructor" key


matthewfranglen:
When converting an object using `fromJS` the original object can be returned. This occurs when the original object has a `constructor` value within it, and that value doesn't equal `Object`.

The relevant code appears to be [this](https://github.com/facebook/immutable-js/blob/c65062599ecad54a1ad5bacd72f65e8f9ef7449b/src/fromJS.js#L38-L40). Presumably this is being done to prevent converting an Immutable data structure a second time.

How can I convert such an object?

### What happened

I converted an object created from downloaded json data. The conversion method, `fromJS`, returned the original object. My further attempts to use the object failed because the required methods were missing.

### How to reproduce

```javascript
let a = JSON.parse('{ "constructor": { "noun": [ "builder", "creator" ] } }');
// OR: let a = { constructor: { noun: [ "builder", "creator" ] } };
let b = Immutable.fromJS(a);

console.log(a === b); // true
console.log(a.constructor === Object); // false
b.get('constructor'); // TypeError: b.get is not a function
```

adamtal3:
Unfortunately for me, this is what I do in all my classes to overcome this:

```javascript
export default class MyClass {
  constructor() {
    // Support ImmutableJS conversion
    this.constructor = Object;
  }
}
```

matthewfranglen:
Thank you for commenting @adamtal3. I feel you have pointed out the other example of this situation - one where you have an object which is an instance of a class and you still want it to be converted.

In my case I resorted to removing the value from my object. It took me some time to determine the cause of this issue, and as such I opened a ticket about it. This could happen to some small part of a very deeply nested object which makes it quite an annoying bug to work around.

If I were to fix this then I would take one of two approaches:
 * Add a boolean flag to the `fromJS` method which will make the conversion unconditionally deeply convert objects when it is true (this would be backwards compatible as not providing it would make it's value undefined - a falsey value).
 * Add a function which decides if the value being considered should be converted. The current `isPlainObj` function would be the default implementation.

I am no expert in javascript though, so I am sure that there are better ways to approach this.

augusto-moura:
There is a way using `Object.getProtypeOf`, I will work on a PR for fix it

leebyron:
Related to #1039

egeriis:
Did anyone have a chance to look into this further, more recently?

PR #1056, opened 2017-02-12
`Immutable.fromJS` bug with no classes objects with custom constructor property

augusto-moura:
As said in #1050, the `Immutable.fromJS` current implementation to plain objects conversion, does not work as expected with objects with a custom `constructor` property. In reality the implementation just returns the same object passed as argument.

I fixed it changing the test for plains objects. Now it tests if the prototype of the object it's the same of `Object.prototype` (if the object has a different prototype it means that it is a class instance). Also, I added some tests cases for the solved problem.

I din't tested the performance impact, it will be more slow than the current implementation for sure.


augusto-moura:
JSPerf points that the new implementation is about 45% slower compared to the current implementation
Https://jsperf.com/plain-object-predicate

leebyron:
Thanks for investigating this. Ideally we could find a more performant implementation. ~1/2 as fast is a very high cost for this edge case

augusto-moura:
Even that the new implementation is slower than the current one, it is quite fast (in practical terms the difference is just some milliseconds).
The performance test is floating between browsers, but the most modern implementations are really optimized.
JSPerf test: https://jsperf.com/plain-object-check

simonkberg:
Couldn't you check `value.constructor === Object || value.constructor === undefined` first and return early, falling back to checking the prototype only if it returns false? Then you would avoid the additional performance cost for most cases.

leebyron:
@simonkberg yes however there are many cases where `isPlainObject()` is handling the uncommon case and it will be very common for non-plain-objects to fall through and hit the slow path.

@augusto-moura milliseconds matter in aggregate! This function is one called very frequently, so I don't think a 25x slowdown will be acceptable. We need to find another approach with a less punishing performance cost.

kozlitinaelja:
@leebyron , it looks like you fixed it by this commit recently:
https://github.com/facebook/immutable-js/commit/2bff0cf0c13e6ec9f2bbfe2c48aec3740d4f552e

So, now there is no need for this PR

leebyron:
I don't think that rev will fix this particular issue. Objects with custom or other-realm prototypes still are not converted

augusto-moura:
I tested adding a condition for only when the object has a own constructor pass by the proptotype check, so the slow path will only hit when necessary, still 10x slower than the current implementation. Imho, in terms of performance, the "constructor is equal to" it's unbeatable.

JSPerf: https://jsperf.com/plain-object-check/6 (for some reason jsperf messed up with the order of test cases)

PR #1059, opened 2017-02-20
Fixed display images in mobile version

VladBlow:
In mobile version big image destroyed width layout, I tried to fix this.

### DIFF
--- a/pages/src/src/base.less
+++ b/pages/src/src/base.less
@@ -58,6 +58,11 @@ blockquote > :last-child {
   margin-bottom: 0;
 }
 
+img {
+  width: auto;
+  max-width: 100%;
+  height: auto;
+}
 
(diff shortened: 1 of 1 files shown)


VladBlow:
@leebyron I think the problem is associated with the tests `es-6` and config compiler TypeScript. Just I don't see this config that fixed.

leebyron:
FYI there's no need to close and reopen the PR, other activity also is visible.

Let me take a closer look at this - just from first glance I see the formatting doesn't match the rest of the file. I'll patch locally to make sure this doesn't break any other aspect of the layout

leebyron:
I applied some edits before merging. Thanks for your help!

VladBlow:
@leebyron Thanks!

PR #1062, opened 2017-02-23
Fix `List.isList` return type in the docs

Stephie:
`List.isList` shows a return type of `maybeList` in the documentation, but the examples clearly show it returning a `boolean`.


leebyron:
This is actually a bug in the documentation generator, not a bug in the type definitions. This is a user defined type guard function:

https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

leebyron:
I just opened an issue for this at #1068 but would still appreciate help in fixing the doc generator!

Issue #705, opened 2015-11-23
Update typescript definitions to latest version of typescript.


Strate:
Typescript now support:
- tuples: `[int, string]` (usable for `entries(): Iterator<[K, V]>`)
- union types: `A | B` (usable for `concat(...valuesOrIterables: Array<Iterable<K,V> | V>`)

chadaustin:
Also polymorphic this, which helps a ton for things like concat.

Strate:
Related to #684

Strate:
```
filter(
      predicate: (value?: V, key?: K, iter?: /*this*/Iterable<K, V>) => boolean,
      context?: any
    ): /*this*/Iterable<K, V>;
```
could be changed to
```
filter(
      predicate: (value?: V, key?: K, iter?: this) => boolean,
      context?: any
    ): this
```

myitcv:
[Related discussion](https://github.com/typings/typings/issues/49)

leebyron:
Updated in master branch, release soon

PR #1064, opened 2017-02-25
make use of Typescript's tuple notation

jbrownson:
Fixes part of https://github.com/facebook/immutable-js/issues/705

The places I fixed were already documented w/ comments, I searched for "[K," and "(K," and filled in the appropriate stuff. The Iterable stuff wants two arguments to Iterable so I left them commented. I'm not sure what the fix for that is.

This should save you some typing anyway if you plan to do this.

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -192,7 +192,7 @@
   export function List<T>(): List<T>;
   export function List<T>(iter: Iterable.Indexed<T>): List<T>;
   export function List<T>(iter: Iterable.Set<T>): List<T>;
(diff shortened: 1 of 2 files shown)


leebyron:
Stuck on #1074

leebyron:
Thanks!

leebyron:
Unfortunately I had to revert this in aeb9aa4f3f9fc374a08ec8d9695ed0dfd2a72cba since it's still breaking tests. Also, this PR seems to have only altered the compiled files and not the original sources, so other tests weren't triggered.

jbrownson:
Thanks for looking at it. I'm not sure if I'm clear, did I do something wrong or just the infrastructure had a problem?

PR #1073, opened 2017-02-28
Update Record typedef (ts)

leebyron:
This uses keyof, Partial, and T[K] typescript advanced types to provide much better type correctness for use of Records

Fixes #909


leebyron:
This isn't safe to land until after #1074

PR #1099, opened 2017-03-07
Static Set.intersect() and Set.union()

leebyron:
These two functions warrant a variant for when the number of sets to be dealt with are unknown, with empty-set special cases.

Closes #415


kentor:
Will this be released in 3.x branch?

PR #1096, opened 2017-03-05
Modernification

umidbekkarimov:
- [x] yarn
- [x] remove jshint in favor of eslint
- [x] prettier
- [ ] transpile src with babel
- [ ] transpile docs with babel
- [ ] remove gulp in favor of npm scripts
- [ ] remove grunt in favor of npm scripts

See #742

PR #1101, opened 2017-03-07
eslint + prettier - jshint

umidbekkarimov:
See #742, #1096


umidbekkarimov:
I didn't modify code to not create merge conflicts.

Preferred eslint rules marked as `warn`.

Run `npm run format && npm run lint:fix` to see the beauty.

pages/src/.eslintrc
```diff
@@ -0,0 +1,9 @@
+{
+  "env": {
+    "browser": true
+  },
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "script"
+  }
+}
```
leebyron: won't this file override the parent eslintrc and remove the custom rules you set up? If the only difference is adding `env.browser` then I think it's safe to merge in

leebyron:
This is awesome, nice work

pages/src/.eslintrc
umidbekkarimov: Of course it's safe 👍 check http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy

umidbekkarimov:
Do not forget to `yarn format` this after merge 😊

pages/src/.eslintrc
leebyron: TIL!

leebyron:
I'd rather not include all the warnings - from past usage I've found warnings essentially get ignored. I think the eslintrc should either turn things off or make them errors.

Some examples of rules that are broken pervasively in this codebase and are safe to turn off:

- vars-on-top
- no-var
- consistent-return
- no-param-reassign

umidbekkarimov:
> Some examples of rules that are broken pervasively in this codebase and are safe to turn off:

Done.

> I think the eslintrc should either turn things off or make them errors

Agree, I kept "worthy" rules as a `warn`s for discussion.

umidbekkarimov:
Closing this from https://github.com/facebook/immutable-js/commit/bb2fbf1749a8dffc3efee3fc0033ba3d2457d252

PR #1104, opened 2017-03-07
Overhaul flow type definitions for Record

### LINKED ISSUE #992: Can we strengthen the return type of Record constructor?
I have whole chunks of code that seems to be uncovered by Flow because of this TODO item:

```
//TODO: Once flow can extend normal Objects we can change this back to actually reflect Record behavior.
// For now fallback to any to not break existing Code
declare class Record<T: Object> {
  static <T: Object>(spec: T, name?: string): /*T & Record<T>*/any;
```

I'm wondering if anyone knows when Flow will have a fix for this in place, so that we can start type checking Records.

leebyron:
Fixes #992
Closes #876

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -1537,7 +1537,7 @@
        *
        * @alias toJSON
        */
-      toJS(): any;
+      toJS(): Object;
 
       /**
        * Shallowly converts this Record to equivalent JS.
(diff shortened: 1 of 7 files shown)


rgbkrk:
Thank you! Do you think this is the limit of where these flow definitions can get with current flow primitives?

leebyron:
As far as I can tell yes, though if you're able to squeeze out any improvements that I missed, send a PR!

PR #1105, opened 2017-03-07
Enforces keyPath is Ordered/Array and not string + error clarity.

### LINKED ISSUE #451: Small improvement for deep methods
As for me it would be convenient  to pass not only array in 'getIn, updateIn, setIn' methods.
https://monosnap.com/image/3fIhtzRKIFP3um0vCF0eSmrzqBPc05

### LINKED ISSUE #635: invalid keyPath errors are very cryptic and hard to debug
For example when running this code:

``` javascript
Immutable.Map().set('foo', undefined).setIn(['foo', 'bar'], 1)
```

You get a generic error without much information:

```
Uncaught Error: invalid keyPath
```

It would be great if this helped you out a bit more and said something along the lines of:

```
Uncaught Error: Attempting to set the path "foo, bar" but "foo" has already been set.
```

Or whatever is most appropriate.

### LINKED ISSUE #940: getIn(string) should throw an error
``` js
Immutable.fromJS({a: {b: 4}, ab: 5}).getIn('ab') // returns 4
// whereas someone might expect 5 but really an error should be thrown
```

A correct way to do this would be:

``` js
Immutable.fromJS({a: {b: 4}, ab: 5}).getIn(['a','b'])
```

Strings are Iterable but often not considered collections because that leads to confusion.

leebyron:
This makes a few important changes:

* Existing errors for bad key paths have been improved to explain why and where the issue is.
* Previously strings were converted into an array of characters key path, a constant source of confusion. After this PR providing strings will throw an error.
* Previously anything iterable was allowed, but now only Arrays and Ordered Iterables, otherwise an error is thrown. This avoids a rare pitfall where Sets are used as a path, which have undefined iteration order.
* Adds a new thrown error for getIn()/hasIn() if a key path encounters a value which cannot be read with get(). Previously notSetValue was returned.

Fixes #635
Fixes #940
Fixes #451


kyldvs:
Awesome! Thank you for this :)

PR #1117, opened 2017-03-08
Fix: deleteIn can remove items from a nested Set

### LINKED ISSUE #489: `deleteIn` fails for Set within Map
I would expect this code to work:

```
new Map().set("decks", new Set([1, 12, 2])).deleteIn(["decks", 12])
```

instead it throws: `Uncaught Error: invalid keyPath`

leebyron:
Fixes #489


namuol:
Will this fix be available in a 3.x release?

PR #1127, opened 2017-03-09
Pull Request Test

rtemelkovski:
Commented Stack.js data structure with description of each of the methods including the parameter types and descriptions and descriptions for the return (if there is one)

### DIFF
--- a/src/Stack.js
+++ b/src/Stack.js
@@ -7,33 +7,54 @@
  *  of patent rights can be found in the PATENTS file in the same directory.
  */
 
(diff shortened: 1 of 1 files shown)


leebyron:
Thanks for digging into this. 

Our documentation lives in type-definitions/immutable.d.ts and is the source from which all other documentation is generated. I'd prefer not to create duplicate efforts for full documentation

leebyron:
I'm going to close this PR - but feel free to re-open or opening a new PR if you are interested in moving some of this additional documentation to the documentation source of truth. See #1144 for an example of editing those files.

PR #1130, opened 2017-03-09
Fix extra args type

nmn:
I think I said that `_: empty` works. Well, it no longer works. So I basically did a find and replace with `..._: []` which does work.

My Bad:(


leebyron:
Thanks for fixing :)

PR #1131, opened 2017-03-09
Update to prettier 0.22

vjeux:
This fixes many issues that were found in 0.21!


leebyron:
So good! Thanks for the rapid fire release

PR #1135, opened 2017-03-10
RFC: Refactor Record. No longer a Collection. Faster guts.

### LINKED ISSUE #286: Faster implementation for Record
Record is currently implemented with a Map. But because it has a fixed size, and because the keys are always strings, it's possible to use a much simpler and faster implementation.

Each Record would have two properties, `_keys` and `_values`. `_keys` is a mutable JavaScript object mapping string keys to indexes, and `_values` is a mutable JavaScript array that contains the actual values.

To lookup a key in the Record, you first look it up in `_keys` which gives you an index, and then you lookup that index in `_values`.

To set a key in a Record, you first look it up in `_keys`, which gives you an index. You then make a copy of the `_values` array, mutate the index to be the new value, then return a new Record with the new `_values` array.

There are a few reasons this is fast:
- You don't need to mutate or copy `_keys` at all: you can simply pass it as-is to the new Record, because it never changes.
- Copying an array is _super_ fast in JavaScript, because of various VM optimizations.
- Looking up a string key in a JavaScript object is faster than doing the equivalent lookup in a Map, and you get to take advantage of the various VM optimizations for objects.

I have implemented this technique in my library, and you can find the benchmarks here:

https://github.com/Pauan/Immutable/blob/javascript/benchmarks/Record/2015-01-16

Getting is always super fast. Setting is fast until you reach around 1,000 keys or so, and it's _extremely_ rare to have Records with 1,000+ keys in them, so that isn't a problem.

### LINKED ISSUE #505: Remove homogeneous collection API from records

leebyron:
This is a breaking change refactor of Record which no longer extends a collection type and therefore does not inherit sequence methods - which was a constant source of confusion and breakage. It also refactors the internals to no longer rely on a Map, but instead a fixed size List which should result in dramatically faster performance. This is also a breaking change as `delete()` (aka `remove()`) and `clear()` are no longer available - previously these methods reverted values to their default value, now that can be done with `set(k, undefined)`.

This adds a predicate function `isRecord()` and also minorly refactors related functionality such as `Seq()` constructors.

Fixes #505
Fixes #286


leebyron:
Also tested an implementation which used a native Array to host Record values rather than an Immutable List, but did not find a significant performance improvement vs the List, but did see rather punishing memory usage for very large Records.

PR #1136, opened 2017-03-10
Separate type for get() with notSetValue provided.

leebyron:
As suggested in https://github.com/facebook/immutable-js/pull/1035#discussion_r105333309


eirslett:
Maybe this patch can be added to the flow typings as well?

PR #1141, opened 2017-03-10
Run deploy docs script only on `master` branch.

umidbekkarimov:
See https://docs.travis-ci.com/user/deployment/script/


umidbekkarimov:
Also i suggest to add `tags: true` - so docs will be deployed only when new release published.

leebyron:
Thanks for this - I've been meaning to add this forever.

PR #1143, opened 2017-03-10
Add pre-commit and pre-push git hooks

umidbekkarimov:
I saw this https://github.com/facebook/immutable-js/commit/0c12cea7376ace34feb6bfeb41d33f941b630c33 commit and think that this checks should be done also on developer side.

Also this PR should be updated after merge of #1140.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,16 @@
   "typescript": {
     "definition": "dist/immutable.d.ts"
   },
+  "lint-staged": {
+    "{src,pages/src,pages/lib}/**/*.js": [
+      "prettier --write --single-quote",
+      "eslint",
+      "git add"
+    ]
+  },
   "scripts": {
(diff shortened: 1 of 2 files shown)


leebyron:
I think this might be too invasive. I find that git hooks can be more burdensome than helpful and what we really want is "pre merge" hooks where travis fills in.

leebyron:
I'm going to close this PR - feel free to open it again if you would like to continue discussion on it

umidbekkarimov:
> I think this might be too invasive. 

Agree. I thought it can be an overkill for a lib after I published PR.

We can keep only `pre-commit` hook and stale this PR until `fix lint` and `fix format` commits start to appear.

> what we really want is "pre merge" hooks where travis fills in.

I don't think that I understand what should go there

PR #1150, opened 2017-03-12
Update README.md

wisdomofgod:
I found that when using immutable.js v4, the TypeScript version needs to be higher than v.2.1.0, and need to use import instead of require


leebyron:
Thanks!

Issue #1146, opened 2017-03-11
Flow-friendly method for filtering nulls


omerzach:
It'd be great to have a method on `Collection`s that turns a `Collection<?T>` into a `Collection<T>` by filtering out `null` and `undefined` values.

You can trivially do this with `xs => xs.filter(x => !!x)` or something along those lines, but Flow isn't smart enough to infer the desired return type.

Currently we're doing this with a utility function, but it'd be nice to have it as a method so it can be called in a sequence of other chained methods.

leebyron:
I'd love to support this in a general way, but as far as I know Flow and Typescript don't support lifting a predicate function onto a collection like this. As soon as they add support, then your function as written would pick up that knowledge.

Until then I recommend a type cast or utility function as you're using today.

PR #1152, opened 2017-03-12
Existing TS typing for `Collection.filter` sometimes causes issues

astashov:
Right now in TypeScript typings `Collection.filter` returns `this`,
which means it's a subtype of the containing class. I.e. if we have a
type `Collection<number, string>`, `Collection.filter` would return
`Collection<number, string>`.

It causes some issues like those:

```typescript
function foobar(): List<string> {
  const a = List.of<string | number>("a", 1, "b");
  const b = a.filter(i => typeof i === "string").toList();
  return b;
}
```

causes a type error:

```
Error:(689, 3) TS2322:Type 'List<string | number>' is not assignable to type 'List<string>'.
  Type 'string | number' is not assignable to type 'string'.
    Type 'number' is not assignable to type 'string'.
```

Same thing happens if we work with subclasses, like:

```typescript
class A {
}

class B extends A {
  public foo(): string {
    return "B";
  }
}

class C extends A {
  public bar(): string {
    return "C";
  }
}

function foobar(): List<C> {
  const a = List.of<A>(new B(), new C());
  const b = a.filter(i => i instanceof C).toList();
  return b;
}
```

shows an error:

```
Error:(702, 3) TS2322:Type 'List<A>' is not assignable to type 'List<C>'.
  Type 'A' is not assignable to type 'C'.
    Property 'bar' is missing in type 'A'.
```

Not sure what's the right solution would be. Maybe just add another
typing for `filter`, along with already existing one, which accepts
a value type in the returned collection, like I did in this
Pull Request?

Then, if you have a case like I described above, you can just provide a
type for M, like `foo.filter<string>(a => typeof a === "string)`.

What do you think?

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -3283,6 +3283,10 @@
       predicate: (value: V, key: K, iter: this) => boolean,
       context?: any
     ): this;
+    filter<M>(
+      predicate: (value: V, key: K, iter: this) => boolean,
+      context?: any
+    ): Collection<K, M>;
 
     /**
      * Returns a new Collection of the same type with only the entries for which
--- a/dist/immutable.d.ts
+++ b/dist/immutable.d.ts
@@ -3283,6 +3283,10 @@ declare module Immutable {
       predicate: (value: V, key: K, iter: this) => boolean,
       context?: any
     ): this;
+    filter<M>(
+      predicate: (value: V, key: K, iter: this) => boolean,
+      context?: any
+    ): Collection<K, M>;
 
     /**
      * Returns a new Collection of the same type with only the entries for which
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -3283,6 +3283,10 @@ declare module Immutable {
       predicate: (value: V, key: K, iter: this) => boolean,
       context?: any
     ): this;
+    filter<M>(
+      predicate: (value: V, key: K, iter: this) => boolean,
+      context?: any
+    ): Collection<K, M>;
 
     /**
      * Returns a new Collection of the same type with only the entries for which


leebyron:
Wouldn't this PR also result in problems? I would expect an error like: "Collection<string, C> is not assignable to type List<C>" - Returning a different type would probably mean replicating this change across all types much like `map<M>`

Looking into TypeScript's own lib, it seems like https://github.com/Microsoft/TypeScript/blob/master/lib/lib.d.ts#L1097 might be leveragable.

leebyron:
#1146 Also mentions a desire for a sounder filter function - I was wrong to suggest TypeScript couldn't do that, but it seems Flow cannot yet.

astashov:
> Wouldn't this PR also result in problems? I would expect an error like: "Collection<string, C> is not assignable to type List" - Returning a different type would probably mean replicating this change across all types much like map<M>

Ah, right, it should be replicated across all types. Well, it seems like we could still keep `filter` method (without type arg) returning `this`, and replicate `filter<S>` (with type arg) returning some specific collection type for each collection type.

> Looking into TypeScript's own lib, it seems like https://github.com/Microsoft/TypeScript/blob/master/lib/lib.d.ts#L1097 might be leveragable.

You mean using guards in a return type of a predicate? Like:

```ts
filter(
  predicate: (value: V, key: K, iter: this) => boolean,
  context?: any
): this;
filter<S extends V>(
  predicate: (value: V, key: K, iter: this) => value is S,
  context?: any
): Collection<K, S>;
```

It seems like the problem with guard functions - you explicitly should specify that their return type is a guard. I.e. I can't use `filter` like:

```ts
const b = a.filter<C>((i: A) => i instanceof C);
```

it will complain like:

```
Error:(700, 25) TS2345:Argument of type '(i: A) => boolean' is not assignable to parameter of type '(value: A, key: number, iter: List<A>) => value is C'.
  Signature '(i: A): boolean' must have a type predicate.
```

I'd have to write it as:

```ts
const b = a.filter<C>(((i: A) => i instanceof C) as (i: A) => i is C);
```

requirement to always specify the type of the predicate could be annoying.
Thought since maybe it's sort of an edge case, and usually you'd use the `filter` without a type arg, maybe not that big of a deal :)

leebyron:
I just put up #1155 which includes a type-checked test that seems to be working just fine without requiring much specific annotation, the test before the changes to the type definitions failed with the same "Type 'List<A>' is not assignable to type 'List<C>'." error you were seeing before

leebyron:
I think the trick was to put the version using guards before the version that does not, since TypeScript attempts to apply each override in the order they were defined.

PR #1155, opened 2017-03-13
Improve filter typescript def to leverage predicate functions

leebyron:
This expands the type def of filter() to return a type-limited version of the original type.

Fixes #1152

### DIFF
--- a/__tests__/List.ts
+++ b/__tests__/List.ts
@@ -484,6 +484,21 @@ describe('List', () => {
     expect(r.toArray()).toEqual(['b', 'd', 'f']);
   });
 
+  it('filters values based on type', () => {
+    class A {}
+    class B extends A {
+      b(): void { return; }
+    }
+    class C extends A {
+      c(): void { return; }
+    }
+    let l1 = List<A>([ new B(), new C(), new B(), new C() ]);
+    // tslint:disable-next-line:arrow-parens
+    let l2: List<C> = l1.filter((v): v is C => v instanceof C);
+    expect(l2.size).toEqual(2);
+    expect(l2.every(v => v instanceof C)).toBe(true);
+  });
+
   it('reduces values', () => {
     let v = List.of(1, 10, 100);
     let r = v.reduce<number>((reduction, value) => reduction + value);
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -767,6 +767,22 @@
       mapper: (value: T, key: number, iter: this) => Iterable<M>,
       context?: any
     ): List<M>;
+
+    /**
+     * Returns a new List with only the values for which the `predicate`
+     * function returns true.
+     *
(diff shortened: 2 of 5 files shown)


__tests__/List.ts
```diff
@@ -484,6 +484,16 @@ describe('List', () => {
     expect(r.toArray()).toEqual(['b', 'd', 'f']);
   });
 
+  it('filters values based on type', () => {
+    class A {}
+    class B extends A {}
+    class C extends A {}
```
astashov: Please make sure to add some methods to classes `B` and `C`. Because of structural typing, TS will be fine with this case, but if you do something like:

```ts
class A {}
class B extends A {
  public foo(): string {
    return "B";
  }
}
class C extends A {
  public bar(): string {
    return "C";
  }
}
```

then `let l2` will have a type error.

leebyron: Thanks for pointing that out. I'll add this to the test

leebyron: Yeah you're right - this also demands a properly attributed predicate function

__tests__/List.ts
```diff
@@ -484,6 +484,24 @@ describe('List', () => {
...
+    }
+    class C extends A {
+      c(): void {}
+    }
+    let l1 = List<A>([ new B(), new C(), new B(), new C() ]);
+    // Note: TypeScript requires a defined type guard to reduce the type. Since
+    // arrow functions cannot define return types, we must use a normal function.
+    let l2: List<C> = l1.filter(function (v): v is C {
+      return v instanceof C;
+    });
```
leebyron: @astashov - this works well and isn't that awkward.

__tests__/List.ts
```diff
@@ -484,6 +484,24 @@ describe('List', () => {
...
+    class B extends A {
+      b(): void {}
+    }
+    class C extends A {
+      c(): void {}
+    }
+    let l1 = List<A>([ new B(), new C(), new B(), new C() ]);
+    // Note: TypeScript requires a defined type guard to reduce the type. Since
+    // arrow functions cannot define return types, we must use a normal function.
+    let l2: List<C> = l1.filter(function (v): v is C {
```
astashov: Oh, I didn't that you can workaround that `as` by just using a `function`, nice!

artursvonda: Not sure that comment is correct. Arrow functions do support return types. 

http://www.typescriptlang.org/play/#src=class%20A%20%7B%20%7D%0D%0A%5Bnew%20A()%5D.filter((v)%3A%20v%20is%20A%20%3D%3E%20v%20instanceof%20A)

Or is this about something else?

leebyron: @artursvonda oh, TIL!

astashov:
Woot! Thanks a lot for this! It was a major blocker for us during migration to 4.0.0.

__tests__/List.ts
artursvonda: Cheers :)

PR #1175, opened 2017-03-15
Add failing test for power-of-2 hashes

garrettm:
It seems there is an issue with hash codes that are powers of 2 above 2**32--they cause infinite recursion.

Associated issue: https://github.com/facebook/immutable-js/issues/1176

### DIFF
--- a/__tests__/Set.ts
+++ b/__tests__/Set.ts
@@ -299,7 +299,7 @@ describe('Set', () => {
     expect(set.count(x => true)).toEqual(5);
   });
 
-  it('does not stack overflow on certain hash collisions', () => {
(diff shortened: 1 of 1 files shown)


kozlitinaelja:
Thank you @garrettm for adding test! It can be merged when the fix will be ready  =)

leebyron:
Unfortunately this is just part of the contract of `hashCode()` that this test violates. `hashCode()` is expected to return a uint type, you could fix your test by instead using:

```
hashCode() {
  return (2 ** 32) | 0
}
```

leebyron:
I just added this note to the documentation, and also included a fix to better handle this case!

garrettm:
Great, thank you @leebyron

PR #1189, opened 2017-03-26
Add safe refs for no-op 'map' for List and Map

### LINKED ISSUE #679: Return safe ref for no-op `map`
This optimization should allow `list === list.map(x => x)` and `map === map.map(x => x)`

DanielMiddleton:
* Implement map() in List and Map

Fixes #679 

By delegating to set() at each step, we reuse the existing update logic that handles the structural sharing.

This arguably introduces new inconsistencies. For example, flatMap still returns different references. I think a "safe" version of flatten() would need to be implemented first.


leebyron:
I just tested this out, and unfortunately it's considerably slower.

If you'd like to rebase and test a version with better performance - that would be great to see, please reopen if you do so.

PR #1193, opened 2017-03-30
Flow: Record: Fix created class types

oreqizer:
Change definition for Record where constructor was specified with 'new'.

### DIFF
--- a/dist/immutable.js.flow
+++ b/dist/immutable.js.flow
@@ -1143,8 +1143,8 @@ declare class Record {
 }
 
 declare class RecordClass<T: Object> {
-  static <T>(values?: $Shape<T> | Iterable<[string, any]>): RecordClass<T> & T;
-  constructor<T>(values?: $Shape<T> | Iterable<[string, any]>): RecordClass<T> & T;
+  static (values?: $Shape<T> | Iterable<[string, any]>): RecordClass<T> & T;
+  constructor(values?: $Shape<T> | Iterable<[string, any]>): RecordClass<T> & T;
 
   size: number;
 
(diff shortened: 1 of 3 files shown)


oreqizer:
hmm no feedback, should I change something?

ianwcarlson:
@oreqizer This solution works a lot better for me.  The downside is it requires making a dummy instance of RecordClass for `typeof` to work. Example:
```
import { Record } from 'immutable';

const testRecord = Record({
  foo: '',
  bar: 0,
});

const dummyInstance = testRecord(); // can't use "new" here
const anotherInstance = testRecord();

// can't use typeof testRecord here
const testFunc = (inputRecord: typeof dummyInstance) => null;

testFunc(anotherInstance);
```
Semantically it makes sense that a record class is different than an instance of it, but it makes the code a little awkward.  Still, better than before where I had to create my own intersection type from RecordInstance to get this to work.

I've been trying to come up with a clean solution myself, but haven't so far.

gajus:
Whats stopping this from being released?

It looks like a reasonable approach.

kozlitinaelja:
@leebyron, are you agree with this approach? =)

leebyron:
`new` is generally discouraged across all of Immutable.js - these are factory functions (like `String` or `Number`) and not class constructors. We only use class in flow types where necessary to help flow understand type refinements it otherwise wouldn't understand.

One unfortunate issue with this approach is that while it technically allows you to use `new`, the resulting value is only a class instance and does not also have property types. 

That might be fine, but it's a limitation.

I'm going to amend this PR to also include a `RecordOf` higher-order type which allows you to better type the return of a record factory function call.

leebyron:
Specifically check out https://github.com/facebook/immutable-js/blob/master/type-definitions/tests/record.js#L66-L83 for how this ends up affecting the ability to type Records

PR #1194, opened 2017-03-30
expect setting Record property to throw

jokester:
Hello there, just realized the property setter is not tested so this PR :)


kozlitinaelja:
Thank you for improving tests @jokester =) Looks legit to me!

Issue #458, opened 2015-04-29
Zipping uneven iterable


dkingman:
Immutable currently will only zip iterables up to the shortest length of the set. So for something like this:

```
var a = Seq.of(1, 2, 3);
var b = Seq.of(4, 5;
var c = a.zip(b)
```

`c` will be `Seq [ [ 1, 4 ], [ 2, 5 ] ]`. Is this intended? This seems like the least useful case. Lodash for example will fill in, "missing" values with `undefined`.

mduvall:
I believe this is intended, consider a pathological case where `a` in your example is a stream and you know that `b` is a finite sequence. The `Seq` returned should close over the finite sequence.

Other language examples include:

The documentation for Scala's `List.zip` says it will return: `a new list containing pairs consisting of corresponding elements of this list and that. The length of the returned collection is the minimum of the lengths of this list and that.`

In both Scala and Haskell this is true:

```scala
scala> val a = List(1,2,3)
a: List[Int] = List(1, 2, 3)

scala> val b = List(1,2)
b: List[Int] = List(1, 2)

scala> a zip b
res0: List[(Int, Int)] = List((1,1), (2,2))
```

```haskell
Prelude> let a = [1,2,3]
Prelude> let b = [1,2]
Prelude> zip a b
[(1,1),(2,2)]
```

dkingman:
Fair enough. Wasn't thinking about the case of infinite sequences. Might be nice to have a function like Scala's `zipAll` which takes up to the longest list, filling in with a default as it goes. Not sure how well that would play with the existing interface in Immutable. 

Another option would be to pass in a length parameter and take up to that length, filling in with a default. This would avoid the infinite sequence issue.

leebyron:
This is a great suggestion. Other libs call this "zipLongest" but I kinda like "zipAll"

mduvall:
@leebyron Ah interesting, did not know that!

mduvall:
@leebyron @dkingman I gave it a swing over in https://github.com/facebook/immutable-js/pull/460, would love feedback!

dkingman:
Great work @mduvall! Thanks for taking this on. Scala and [Python](https://docs.python.org/2/library/itertools.html#itertools.izip_longest) both provide the option to fill with a default value. I know doing it that way would involve diving into the iterator so I'm not sure if it's worth it. Ruby and others don't allow specified defaults and just fill with nil/null/etc. I can see how allowing a specified default would be useful though. What do you all think?

[...]

PR #1195, opened 2017-03-30
Implement zipAll (Issue #458)

adityavkk:
Implements `zipAll`

Unlike `zip`, `zipAll` continues zipping until the longest collection is exhausted. Missing values from shorter collections are filled with `undefined`.

This PR builds on some earlier work done in a previous [PR](https://github.com/facebook/immutable-js/pull/460)


src/Operations.js
```diff
@@ -822,18 +825,22 @@ export function zipWithFactory(keyIter, zipper, iters) {
       let steps;
       if (!isDone) {
         steps = iterators.map(i => i.next());
-        isDone = steps.some(s => s.done);
+        if (zipAll) isDone = steps.every(s => s.done);
+        else isDone = steps.some(s => s.done);
       }
+
```
arjans: This is the critical change in logic, as @leebyron [pointed out](https://github.com/facebook/immutable-js/pull/460/files#r29642099) in an earlier pull request.

dist/immutable.js
```diff
@@ -1848,18 +1851,22 @@ function zipWithFactory(keyIter, zipper, iters) {
...
+
       if (isDone) {
         return iteratorDone();
       }
+
       return iteratorValue(
         type,
         iterations++,
-        zipper.apply(null, steps.map(function (s) { return s.value; }))
+        zipper.apply(null, steps.map(function (s) { return s.done ? undefined : s.value; }))
```
leebyron: When `s.done` is true, `s.value` is guaranteed to be `undefined`, so is this change necessary?

dist/immutable.js
```diff
@@ -1848,18 +1851,22 @@ function zipWithFactory(keyIter, zipper, iters) {
       var steps;
       if (!isDone) {
         steps = iterators.map(function (i) { return i.next(); });
-        isDone = steps.some(function (s) { return s.done; });
+        if (zipAll) { isDone = steps.every(function (s) { return s.done; }); }
+        else { isDone = steps.some(function (s) { return s.done; }); }
```
leebyron: Please use a ternary for this:

```
isDone = zipAll ? steps.every(...) : steps.some(...)
```

PR #1196, opened 2017-03-31
Make fromJS() doc more consistent with code

karland:


leebyron:
Thanks for improving the docs, and sorry for the wait :)

PR #1204, opened 2017-04-06
Bundle an es module

### LINKED ISSUE #1190: Support tree shaking
I avoid using immutable because it is a huge chunk of code relative to my apps, and for initial page load on mobile, every KB matters. I do get it via third party components, which I then load on-demand only.

However, I noticed that the size is always the same, 57KB minified (using the excellent [webpack-bundle-analyzer](https://github.com/th0r/webpack-bundle-analyzer)), which is also the size that it ships with in the npm package.

I doubt that all third party modules use all functionality, so that means that the minification cannot identify unused pieces of code. I build with webpack tree shaking, which works using ES2016 export syntax, first identifying unused exports, not exporting them, and then letting the minification remove any unused code.

So, I think that the immutable library could be smaller if it was also available in an `es` flavor (in the same package, via a package.json key) that exports everything separately.

However, that will only help if the code is "chunky", so that if you only use e.g. a `Map`, it will only touch part of the code.

Furthermore, the code should not have side effects. I noticed a lot of `createClass` calls, the minifier won't be able to optimize those away because it cannot assume it to have no side effects.

Another caveat is that lodash faces the same problem, and the `lodash-es` build [is not helping](https://github.com/webpack/webpack/issues/1750), probably due to the fact that the default export exports everything as an object. Lodash has to resort to a Babel plugin that rewrites `import {foo} from 'lodash'` to `import foo from 'lodash/foo'`, and as a consequence, both `lodash` and `lodash-es` get the same build size without tree shaking.

TL;DR: *If* Immutable is "chunky", it can lead to smaller build sizes by tree shaking *iif* it has an `es` build **and** it doesn't export everything on `default` **and** things like `createClass` are done at build time. I think/hope.

EnoahNetzach:
This PR is aimed to produce an es module build, targeted by bundlers (e.g. rollup & webpack).

This will enable tree shaking on their side, so will fix #1190.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
     "url": "https://github.com/facebook/immutable-js/issues"
   },
   "main": "dist/immutable.js",
-  "jsnext:main": "dist/immutable.es.js",
   "module": "dist/immutable.es.js",
   "typings": "dist/immutable-nonambient.d.ts",
   "typescript": {
--- a/package.json
+++ b/package.json
@@ -15,17 +15,20 @@
     "url": "https://github.com/facebook/immutable-js/issues"
   },
   "main": "dist/immutable.js",
+  "jsnext:main": "dist/immutable.es.js",
+  "module": "dist/immutable.es.js",
   "typings": "dist/immutable-nonambient.d.ts",
   "typescript": {
     "definition": "dist/immutable.d.ts"
   },
   "scripts": {
     "build": "run-s build:*",
-    "build:dist": "run-s clean:dist bundle:dist copy:dist stats:dist",
+    "build:dist": "run-s clean:dist bundle:dist bundle:es copy:dist stats:dist",
     "build:pages": "gulp --gulpfile gulpfile.js default",
     "stats:dist": "node ./resources/dist-stats.js",
     "clean:dist": "rimraf dist",
     "bundle:dist": "rollup -c ./resources/rollup-config.js",
+    "bundle:es": "rollup -c ./resources/rollup-config-es.js",
     "copy:dist": "node ./resources/copy-dist-typedefs.js",
     "lint": "run-s lint:*",
     "lint:ts": "tslint \"__tests__/**/*.ts\"",
--- a/dist/immutable.es.js
+++ b/dist/immutable.es.js
@@ -1,0 +1,5501 @@
+/**
+ *  Copyright (c) 2014-2015, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+// Used for setting prototype methods that IE8 chokes on.
+var DELETE = 'delete';
+
+// Constants describing the size of trie nodes.
+var SHIFT = 5; // Resulted in best performance after ______?
+var SIZE = 1 << SHIFT;
+var MASK = SIZE - 1;
+
+// A consistent shared value representing "not set" which equals nothing other
+// than itself, and nothing that could be provided externally.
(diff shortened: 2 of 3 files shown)


EnoahNetzach:
I acknowledge there should be a "module" like in the [`bower.json`](https://github.com/facebook/immutable-js/blob/36c9a98776642af40aa9af04c78dd20c8098248c/bower.json#L13), I just wasn't able to find out how to do it..

package.json
```diff
@@ -15,17 +15,20 @@
     "url": "https://github.com/facebook/immutable-js/issues"
   },
   "main": "dist/immutable.js",
+  "jsnext:main": "dist/immutable.es.js",
```
Shyam-Chen: This is no longer recommended.

EnoahNetzach: From [rollup docs](https://github.com/rollup/rollup/wiki/pkg.module):

>  Note: Tools such as rollup-plugin-node-resolve and Webpack 2 treat jsnext:main and module interchangeably. They're the same thing, except that module is more likely to become standardised.

Shyam-Chen:
I think no way tree-shaking, because it has been bundled into a pack.

brantphoto:
Is this something that is being considered? I believe it would make sense to have an es module included in the final package

EnoahNetzach:
I'd hope to. It's only little more than a POF on my part, but I'd like to receive some feedback, especially on adoptability.

deepsweet:
>I think no way tree-shaking, because it has been bundled into a pack.

It should work because Webpack will modify bundle exports so that unused parts of the code are not exported anymore, and then it will be cleaned up by "dead code elimination" after minification with Uglify/Babili (see https://webpack.js.org/guides/tree-shaking/ for details). That's why ES build should contain untranspiled imports/exports – it can be statically analyzed and modified.

I've been waiting for modular ImmutableJS so long ❤️ 🙏

Shyam-Chen:
@deepsweet Webpack can tree-shaking `three.js`?

StevenLangbroek:
@Shyam-Chen theoretically yes, it includes a "module" attribute in its package.json.

wmertens:
I'm really curious how well this will work! I noticed code like `mixin(...)` in Collection.js, any fuction calls that run at module load time will be considered as possible side effect code and can't be tree shaken.

I will take a look at the effect next week or so.

inakianduaga:
@leebyron Are there any instructions on how to use this new es-module? I'm not getting any benefits in build after the update. I haven't modified the way I import immutableJS in the codebase. It could also be something related to my Webpack config. Therefore some light on how to benefit from this would be nice

deepsweet:
@inakianduaga it should be noticeable only with minified builds because it relies on "dead code elimination". have you checked your final `.min.js` files?

deepsweet:
also you can take a closer look at ImmutableJS in your bundle, it should contain special "unused harmony export" comments made by Webpack for stuff you have never used. [example](https://github.com/webpack/webpack/tree/master/examples/harmony-unused).

inakianduaga:
@deepsweet Yeah, this was for minified files. I will check the webpack comments on the bundle. BTW, do you have a rough estimate of the reduction in size? I'm using list/map/records and nothing else from the library.

deepsweet:
I haven't checked it by myself yet, but it depends on Immutable.js internals – if list/map/records are using 90% of the entire library then you'll get that 90% in your bundle. And if so then it requires another PR to reduce amount of internal cross-dependencies when it's possible.

leebyron:
>  I'm using list/map/records and nothing else from the library.

That's about 95% of the library :) - You're not likely to see much reduction

DanielRamosAcosta:
I just installed `v4.0.0-rc.9`, but `dist/immutable.es.js` isn't in the `node_modules` directory, any idea why?

Issue #1202, opened 2017-04-05
v4.0.0-rc.2 Record.equals breaks when passed null or undefined


cmschuetz:
### What happened

When passing in `null` or `undefined` into a record instance's `.equals` method, I receive a `TypeError: Cannot read property '_keys' of null` exception instead of the expected `false`.

### How to reproduce

```
const ABRecord = Record({ a: 1, b: 2 });
const myRecord = new ABRecord();

console.log(myRecord.equals(null));
console.log(myRecord.equals(undefined));
console.log(myRecord.equals(0);
console.log(myRecord.equals('');
console.log(myRecord.equals(false);
```

Output

```
TypeError: Cannot read property '_keys' of null
TypeError: Cannot read property '_keys' of undefined
false
false
false
```

anilreddykatta:
I would like to take up this issue. Adding a check for null or undefined in equals should fix this?

PR #1208, opened 2017-04-11
Contains fix for error during equals check on Record with undefined or null

anilreddykatta:
1. Added fix for checking null or unfined during equality checking
2. Added tests for the check

Addresses #1202


kozlitinaelja:
Looks legit to me! Thank you @anilreddykatta for fixing it! @leebyron, is it ok if I merge it? =)

anilreddykatta:
@leebyron Thanks for the feedback. I made changes accordingly @kozlitinaelja Can you please take a look and let me know of any concerns?

leebyron:
Great fix! Sorry for the wait :)

PR #1222, opened 2017-05-05
Fix typo: change 'hiearchy' to 'hierarchy'

wdhorton:
Noticed this when I was reading over the flow type definitions.


kozlitinaelja:
Thank you @wdhorton for fixing it! =)

Issue #1220, opened 2017-05-03
Seq.rest() throws an exception when invoked on a single item sequence


olivr70:
I am using the version 4.0.0-rc.2
### What happened
An exception was thrown at Iterator.next() (immutable.js line 1496)

### How to reproduce
``` Javascript
var r = Seq(['a']).rest()
var x = [...r] // throws 'TypeError: iterator.next is not a function'
```

This also occurs with multiple items in the sequence, on the last iteration when invoking .rest() (or .slice(1)) several times

olivr70:
As a temporary quick and dirty solution, I patched line 1491 of immutable.js (method **sliceSeq.__iteratorUncached**)
``` javascript
var iterator = sliceSize !== 0 && collection.__iterator(type, reverse);
```
to
````Javascript
    var iterator = sliceSize !== 0 ? collection.__iterator(type, reverse)
      : emptySequence()[Symbol.iterator]()
```
There is certainly a better solution to return an empty sequence, that's why I did not PR

Issue #10, opened 2014-07-30
Newcap convention?


cowboy:
I know built-in `Math` is an exception to this, but it's a widely accepted convention to only capitalize constructor functions, and nothing else, in user JavaScript (see [the JSHint newcap option](http://www.jshint.com/docs/options/#newcap)).

Have you considered using `immutable` instead of `Immutable` in the [documentation and examples](https://github.com/facebook/immutable-js/blob/master/README.md)?

Eg. This:

``` js
var Immutable = require('immutable');
var map = Immutable.Map({a:1, b:2, c:3});
```

Would become this:

``` js
var immutable = require('immutable');
var map = immutable.Map({a:1, b:2, c:3});
```

(etc)

PR #1226, opened 2017-05-14
Fix to issue #1220 - exception thrown when iterating the result of re…

olivr70:
The fix to issue #1220 is just to handle the case of a 0 length slice as a special case, and return an empty iterator.

Note : although the problem was detected using Seq.rest(), is much more general and applies to Seq.slice() also. 

Also I did not test the same scenario for other collections.

Added a issues.ts test file, which can be kept for non regression of thrown away. It fails on 4.0.0-rc2.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "immutable",
-  "version": "4.0.0-rc.2",
+  "version": "4.0.0-rc.2+olivr70",
```
kozlitinaelja: Why this suffix was added? =)

olivr70: It should not have been shared in the PR. This is just to distinguish my own version, which I npm install using the github url. 

Semver spec #10 allows such build metadata, and also specifies that it should be ignored for determining version precedence.

PR #1229, opened 2017-05-18
Fix genTypeDef for MappedType to make Record documentation work

alexandrebodin:
While reading the type def I noticed how the module Collection exported module Keyed etc... and I figured That we could do the same to make the Record.Instance documentation generate correctly. 

This is certainly not a definitive solution but might still be better than no documentation.


alexandrebodin:
Any plans of merging ?

leebyron:
Thanks for fixing this!

PR #1232, opened 2017-05-20
Change the has method of Record to be a type guard

### LINKED ISSUE #1230: The 'has' method on a Record should be a Typescript type guard
The key thing is that right now you can do this:

```
if (recordInstance.has(arbitraryString)) {
  return recordInstance.get(arbitraryString)
}
```

...and still get a Typescript error. I'm sure there must be some way to make this act as a type guard for Typescript, but I'm not sure how to implement it.

icopp:
This allows use of the `has` method as a type guard, while still otherwise acting as a boolean. Resolves #1230.

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -2239,7 +2239,7 @@
 
       // Reading values
 
-      has(key: string): boolean;
+      has(key: string): key is keyof T;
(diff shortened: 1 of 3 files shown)


kozlitinaelja:
Looks legit to me. Thank you @icopp for fixing!

Issue #1161, opened 2017-03-13
4.0rc2 `getIn()` not handling nulls gracefully on Records


EvHaus:
### What happened

After upgrading from `3.x` to `4.x` the `getIn()` method is not behaving as before and not as documented. It often returns `Invalid keyPath: Value at ["someKey"] does not have a .get() method: null` instead of handling null keys gracefully.

### How to reproduce

```
const ABRecord = Immutable.Record({ a: 1})
const myRecord = new ABRecord()
console.log(myRecord.getIn(['a', 'b']));
```

Returns
```
TypeError: Invalid keyPath: Value at ["a"] does not have a .get() method: 1
```

But I expected it to return `null` instead.

leebyron:
This is an intentional breaking change to bring `getIn` and `setIn` into alignment. Bug's resulting from the silent return of `null` for this case were a common complaint from the previous major version.

The reason this fails is because there is a value at `"a"` however that value is not a container and doesn't allow deeper retrieval. This is usually indicative of a mistake, expecting the value at `"a"` to be a collection if it exists.

Note if you change the test code to:

```js
const ABRecord = Immutable.Record({ a: null })
const myRecord = new ABRecord()
console.log(myRecord.getIn(['a', 'b']));
```

That you'll see `null` logged.

EvNaverniouk:
@leebyron Thanks for the detailed explanation, however I believe there is still a bug as the code you posted:

```javascript
const ABRecord = Immutable.Record({a: null});
const myRecord = new ABRecord();
console.log(myRecord.getIn(['a', 'b']));
```

Also returns:

```
TypeError: Invalid keyPath: Value at ["a"] does not have a .get() method: null
```

Additionally, even if this is still intended behavior, I would highly recommend this breaking change get called out specifically on the 4.0 changelog as it was not clear from the current description and it's caused a lot of headaches for us during the upgrade process.

leebyron:
Ah interesting! I'll definitely make sure there's more tests and clear language around this

leebyron:
Reopening as a reminder to myself to dig in

loveholly:
It also happens on nested `Map` structure.

iamdustan:
@leebyron would it be possible to have this behavior have a deprecation cycle for one major version where instead of `throw`ing a warning is logged? For example on our codebase we have 1000+ callsites of `getIn` to migrate before we can update.

[...]

PR #1233, opened 2017-05-22
Deprecate getIn null throwing

iamdustan:
This changes the `getIn` null to a warning rather than an error with a deprecation notice to ease the migration path.

https://github.com/facebook/immutable-js/issues/1161#issuecomment-301813826


leebyron:
I think this is reasonable - smoothing the upgrade path is important. I'll make some small changes to your diff to better fit the original behavior

iamdustan:
Whoa thanks @leebyron. Hope you’ve had a great summer 😄

PR #1258, opened 2017-06-24
more type-safe types for zip for up to three lists.

emmanueltouzery:
more type-safe type TS definitions for zipping for up to three lists. I saw something similar is already done for `zipWith`. Anything is better than `any`, and the typescript compiler is smart enough to pick the most precise type definition.

I've confirmed the compiler does pick up the new type definitions in `__tests/zip.ts` line 28 the type of `zipped`, and by extracting variables in a couple other examples in that file, and all seems to work great.

### DIFF
--- a/dist/immutable-nonambient.d.ts
+++ b/dist/immutable-nonambient.d.ts
@@ -791,6 +791,32 @@
     ): this;
 
     /**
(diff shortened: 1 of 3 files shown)


leebyron:
Agreed, thanks for the help

PR #1261, opened 2017-07-06
Inline runnable examples with RunKit

tolmasky:
I'd like to propose using RunKit embeds for the examples in immutable's documentation. The way I've implemented it, the existing examples remain as is, but also have a "run it" button on the bottom right. When clicked, a RunKit embed is swapped in and the example becomes runnable and editable:

gif1

As you can see from the above gif, we've also taken the time to add native support for Immutable classes into RunKit, so that all the objects show up correctly (you can try this now [here](https://runkit.com/tolmasky/immutable-classes/branches/master/clone)).

Additionally, I made some custom assert.equals/etc so they show up nicely too:

screen shot 2017-07-06 at 11 57 51 am

Some technical details:

1. Its designed in such a way that if RunKit is down, the site behaves like before (so no absolute dependency).
2. I've modified the markdown parser so that if you add "\<!-- runkit:activate --\>" before a code snippet, it will be turned into a runnable snippet, and the others are left alone.
3. I've left some as not runnable, since they rely on the new unpublished 4.0.0 changes. Once 4.0.0 goes live, these could be made runnable too (or alternatively I could modify the examples to use the current API, but I decided to not make any content changes at all, just usability).
4. Some examples were not technically runnable on their own, as they do not contain the proper imports. RunKit has a solution for this called the preamble (where you can load code before the users code). I've kept these examples the same by using this feature and placing the imports in a not visible first cell. However, I am happy to simply add the imports to the examples themselves too.

The basic goal we're trying to accomplish is to make code immediately usable in docs, to shorten the time it takes to get familiar with a library. Lodash has been using RunKit on their documentation for about a year now, and Express recently incorporated RunKit into their Getting Started example as well. https://expressjs.com/en/starter/hello-world.html . Would love to see the same here, and I am happy to make any changes as necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,9 +157,9 @@ const { Map } = require('immutable')
 const map1 = Map( {a: 1, b: 2, c: 3 })
 const map2 = map1.set('b', 2)
 assert.equal(map1, map2) // uses map1.equals
-assert.identical(map1, map2) // uses ===
+assert.strictEqual(map1, map2) // uses ===
 const map3 = map1.set('b', 50)
-assert.not_equal(map1, map3) // uses map1.equals
+assert.notEqual(map1, map3) // uses map1.equals
 ```
 
 Note: As a performance optimization Immutable.js attempts to return the existing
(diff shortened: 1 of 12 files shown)


leebyron:
This is super cool, thanks @tolmasky!

PR #1263, opened 2017-07-07
docs: updates hash() docs to indicate that it's new in v4.0

nsfmc:
this tripped me up the other day when i was looking to use `hash()`. I couldn't offhand find another method that indicated which version it appeared in so i just italicized the string.

grnk


leebyron:
Works for me

PR #1267, opened 2017-07-12
Docs: Use commas consistently in JS-first API section (where giving mutating/array function examples)

joshburgess:
Just a punctation fix in the JavaScript-first API section.


leebyron:
Thanks!

PR #1271, opened 2017-07-15
Add identity discontinuity proof typechecking

jinalkathiara:
Fixing FromJS.isPlainObject for the cross window identity discontinuity. This fix will compare isPlainObject with deep prototype chain.


jinalkathiara:
I see @ihodes @leebyron @styfle are contributors, can you please review and merge

jinalkathiara:
@kozlitinaelja  @liubinyi - I see you contributed last changelist in this project, can you help us resolve this pull request ?

jinalkathiara:
Reason for having this code is Aura locker service. We are building a salesforce -> cisco spark integration where cisco spark library is going to be considered as a 3rd party library. Because of prodsec requirement we are trying to isolate that library from main DOM and we are using Aura Locker Service for that approach. Look at this code how we are isolating it.
https://github.com/forcedotcom/aura/blob/master/aura-resources/src/main/resources/aura/resources/lockerservice/safeEval.html

This way, we are evaluating 3rd party code under the iFrame and all the XHR and various other APIs are wrapped in a secure way. Given that some of the XHR response is of type ArrayBuffer, we might be in a situation where we will check instanceof ArrayBuffer cross window.

Look at this link for cross window instanceof problem. This code is just fixing that and checking deep prototype. This is not some buggy proxy implementation.
http://tobyho.com/2011/01/28/checking-types-in-javascript/

leebyron:
Could you explain more the issue this is solving? isPlainObject should be returning false for deep prototype chains, correct?

Also, could you include performance measurements for this change? This function is called quite a lot and is a potential performance bottleneck for the library

jinalkathiara:
isPlainObject should be returning false for the deep prototype chains,
however this PR is fixing one problem where you are comparing a plain
Object from a different type system. (for example an object defined under
an iFrame)

Look at this link for more information for this here.
http://tobyho.com/2011/01/28/checking-types-in-javascript/

I've also created a perf test for this change. Look at the test result
here. https://jsperf.com/check-performance-for-prototype-checking/1

-Jinal

leebyron:
Thanks, @jinalkathiara.

I updated your perf test here: https://jsperf.com/check-performance-for-prototype-checking/5 - unfortunately yours included console.log which is very slow and warped the test results.

As you can see, this change makes this function 100x slower, which is not going to be acceptable since it's one of the most often called functions within the library.

Could you talk more about the use case you're trying to enable? Perhaps there's a way to accomplish it without the library-wide performance penalty

jinalkathiara:
I updated the perf test one more time with revised version, however its slower as you pointed out. 
Check that here. https://jsperf.com/check-performance-for-prototype-checking/9

What we are trying to solve is addressing this use case handled by this library.

el = document.createElement("iframe")
document.body.append(el)
el.contentWindow.Array instanceof el.contentWindow.Object // true
checkDeepPrototype(el.contentWindow.Array, Array) // true

jinalkathiara:
We are going to change our architecture so that we won't hit this problem.

Issue #1240, opened 2017-06-04
Flow doesn't type check my Record properly


eirslett:
I've tried creating a new `Person` record, and added a couple of test cases for it that should fail the static type checking.

```javascript
// @flow
import { Record } from 'immutable';

type PersonFields = {
  name: string,
  age: number
};

// I also tried this, which didn't work; Use Class<PersonType> instead of Class<*>
// import type { RecordInstance } from 'immutable';
// type PersonType = RecordInstance<PersonFields> & PersonFields;

const Person: Class<*> = new Record(({
  name: '',
  age: 0
}: PersonFields));

const john = new Person({
  name: 'John',
  age: 'invalid' // should give a type error ('invalid' is not a number), but it doesn't
});

// Should give type error ('invalidProperty' cannot be set), but it doesn't
const updatedPerson = john.set('invalidProperty', 4);

// Should give type error ('invalid' doen't exist), but it doesn't
console.log(john.get('invalid'));
```

Is this a bug in the flow definitions for Immutable, or have I used the API wrong? Is there a good example somewhere of how to define a new `Record` with a corresponding flow type for it?

**UPDATE**
For the record, I was using immutable@4.0.0-rc.2 and flow 0.47.0.

I had this in my .flowconfig:

```
[libs]
node_modules/immutable/dist/immutable.js.flow
```

When I removed that line, suddenly it type checked as expected. So this might be a Flow configuration problem on my side. Are we supposed to never include flow typings from node_modules?

eirslett:
This still fails though:

```javascript
const john = Person({
  name: 'John',
  age: 20
});

function increaseAge(person: Person): Person {
  return person.update('age', i => i + 1);
}

console.log(increaseAge(john));
/* FLOW ERROR:
function increaseAge(person: Person): Person {
                                  ^^^^^^ RecordClass. Ineligible value used in/as type annotation (did you forget 'typeof'?)
*/
```

singhshashi:
```
const {Record} = require('immutable');

const PersonRecord = Record({
  name: undefined,
  age: undefined,
});

class Person extends PersonRecord {
  name: string;
  age: number;
  
  constructor(name: string, age: number){
    super({
     name: name,
     age: age
    });
  }
}

function increaseAge(person: Person) : Person {
   return person.update('age', i => i + 1);
}

const John = new Person('John', 20);
console.log(increaseAge(john));
```

eirslett:
[...]

Issue #1172, opened 2017-03-14
Unclear how to accept Record types


jkk:
It's unclear how to annotate Record instances with Flow types:

```js
// @flow
import {Record} from 'immutable';

const TestRec = Record({a: 0, b: 0});

// Flow gives an error for the `rec` argument's type:
//
//   RecordClass
//   Ineligible value used in/as type annotation (did you forget 'typeof'?)
//   TestRec
//
// Is there a type that can be used here? `typeof TestRec`
// does not work, nor do any other variations I've tried.
function handleRec(rec: TestRec) {
  console.log(rec.a);
}

const myRec = TestRec();
handleRec(myRec);
```

If it's possible, examples in the docs would be great. Or, if it's not possible, documentation about the limitations of Records + Flow types would also be great. Happy to help write docs if I can figure out what's going on.

leebyron:
The same issue affects TypeScript. I'm not sure how to improve this but would love some help.

EChikunov:
@jkk 
In case of TypeScript I'm using it like this:

```js
import * as Imm from "immutable";
interface ITestRec {
    field1: number;
    field2: string;
}

const TestRec = Imm.Record<ITestRec>({field1: 0, field2: "default text"});

function someMethod(rec: Imm.Record.Instance<ITestRec>) {
    console.log(rec.field1);
}
```

@leebyron 
Offtop: v4 is a lot better in terms of typescript integration. Thank you guys, it removed a lot of frustration, nice job!

jkk:
@EChikunov Unfortunately the equivalent in Flow doesn't seem to work.

I've tried a million variations, and also tried revising `immutable.js.flow`, but my static-typing skills are not strong enough.

thessem:
I'd absolutely love an example of how to do what @EChikunov did in flow

johnfn:
Hey guys,

I was able to get Records to work with TypeScript with no additional interface types.

Here's the implementation: https://gist.github.com/johnfn/12defda0d0548a4b165201f77618f36c (It's quite long because I copied a lot from other immutable.js definitions. If anyone knows how to clean it up, please feel free to do so.)

Here's a typical use case:

```
const record = MakeRecord({
    a: 1,
    b: foo,
});

console.log(record.a); //.a autocompletes, as does all methods on the Record class.
```

[...]

PR #1276, opened 2017-07-22
Upgrade flow; solidify Record and RecordInstance a bit

rsolomon:
This pull request aims to help out a bit with issues #1240 and #1172 . It does 3 things:

1) Exposes `RecordInstance` for import
2) Changes the return type of some chainable functions to return `this & T` instead of just `this`
3) Uses the new `$ElementType<T, K>` for some better type safety

Note that I cannot seem to get `$ElementType<T, K>` to check `set` properly, but it works fine for `get`. I really don't know why it wouldn't work the same, so any assistance would be appreciated.

### DIFF
--- a/dist/immutable.js.flow
+++ b/dist/immutable.js.flow
@@ -1151,35 +1151,35 @@ declare class RecordInstance<T: Object> {
   size: number;
 
   has(key: string): boolean;
-  get<K: $Keys<T>>(key: K): /*T[K]*/any;
+  get<K: $Keys<T>>(key: K): $ElementType<T, K>;
 
   equals(other: any): boolean;
   hashCode(): number;
 
-  set<K: $Keys<T>>(key: K, value: /*T[K]*/any): this;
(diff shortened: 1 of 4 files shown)


skeet70:
Can someone please merge this before releasing 4.0? Can confirm that even just exporting `RecordInstance` alone solves all our project's issues with code like:

```js
type StateT = Record<{
  reportData: Map<string, List<Object>>,
  loadingSummary: boolean,
  loadingReport: boolean
}>;
const State = Record({
  reportData: Map(),
  loadingSummary: false,
  loadingReport: false
});
```

when changed to:

```js
import type { RecordInstance } from 'immutable';

type StateT = RecordInstance<{
  reportData: Map<string, List<Object>>,
  loadingSummary: boolean,
  loadingReport: boolean
}>;
const State = Record({
  reportData: Map(),
  loadingSummary: false,
  loadingReport: false
});
```

nanot1m:
About `set` look here https://github.com/facebook/flow/issues/4804?_pjax=%23js-repo-pjax-container

leebyron:
Solid improvements, thanks for the help!

PR #1279, opened 2017-07-29
avoid variable name used in upper scope (hashed)

aretecode:


leebyron:
Nice, thanks for fixing this lint warning

errendir:
I think that this change unintentionally altered the behaviour of `hashString` and `hashJSObj` functions. In `hashString`:
```
hash = 31 * hash + string.charCodeAt(ii) | 0;
```
was changed to:
```
hashed = 31 * hash + string.charCodeAt(ii) | 0;
```
thus resulting in `31 * hash` being the multiplication of an integer with a function which evaluates to `NaN`. This makes the `hashString` function return 0 for all strings.

The `hashJSObj` function suffers from a similar issue, and it only returns `undefined`.

Here is a PR that fixes both of those issues: https://github.com/facebook/immutable-js/pull/1325

PR #1284, opened 2017-08-08
Fix get and set types using the new $ElementType

nmn:
Flow finally added $ElementType<T, K> which works exactly like Typescript's `T[K]`.
Updated the types of record to use it.


oriSomething:
I think it's a duplication of #1276

leebyron:
Looks like this got handled in #1276, but thanks for your help @nmn!!

Issue #1283, opened 2017-08-08
flatMap typescript type for Keyed, [Ordered]Map are not substitutable for Collection.flatMap


sandersn:
I'm getting immutable.d.ts to work with Typescript 2.4 right now. The 2.4 compiler finds several new bugs in the types. The worst bug is that `Keyed.flatMap` (and its descendants' `flatMap`s) are not substitutable for the type of `Collection.flatMap`. This type error is reflected at runtime:

```
> var i = require('immutable')
> var s = i.Set([1,2,3])
> var l = i.List([1,2,3])
> var m = i.Map([['a', 1], ['m', 2]])
> var cols = [s, m, l]
> cols.map(col => col.flatMap((v, k) => [v, v, v, v, v, v]))
TypeError: Expected [K, V] tuple: 1
    at validateEntry (/home/nathansa/src/test/node_modules/immutable/dist/immutable.js:3577:13)
    at /home/nathansa/src/test/node_modules/immutable/dist/immutable.js:2924:11
    at ArraySeq.__iterate (/home/nathansa/src/test/node_modules/immutable/dist/immutable.js:375:13)
    at FromEntriesSequence.__iterate
// ... snip some error output ...
> cols.map(col => col.flatMap((v, k) => [[k, v], [k, v]]))
[ Set { 1,1, 1,1, 2,2, 2,2, 3,3, 3,3 },
  Map { "a": 1, "m": 2 },
  List [ 0,1, 0,1, 1,2, 1,2, 2,3, 2,3 ] ]
``` 

As you can see, `Map.flatMap` expects its mapper to return a list of pairs. The other flatMaps expect their mappers to return a list of items. Mixing the two collection types in a list produces either a TypeError or unsatisfactory results.

This mismatch is reflected in the typings in immutable.d.ts:

```ts
export interface Keyed<K, V> extends Collection<K, V> {
  // ...
  flatMap<KM, VM>(mapper: (value: V, key: K, iter: this) => Iterable<[KM, VM]>, context?: any): Collection.Keyed<KM, VM>;
  // ...
}
export interface Collection<K, V> extends ValueObject {
  // ...
  flatMap<M>(mapper: (value: V, key: K, iter: this) => Iterable<M>, context?: any): Collection<K, M>;
  // ...
}
```

The Keyed.flatMap says that `mapper` is allowed to change the key type and the value type, but Collection.keyMap says that `mapper` is only allowed to change the value type.

## Fixes ##

[...]

Issue #879, opened 2016-05-20
Subtyping relationship of Set types makes Iterable#map and friends unsound


samwgoldman:
`SetIterable` is a subtype of `Iterable` by setting `T=K` and `T=V`. There is a similar relationship between `SetCollection`/`Collection` and `SetSeq`/`Seq`.

This is actually problematic and prevents us from soundly defining methods on `Iterable` which change either `K` or `V`, like `map` and `flatMap`.

Here's a subset of type definitions that I can use to show the problem:

``` js
interface Iterable<K,V> {
  map<M>(
    mapper: (value: V, key: K, iter: this) => M,
    context?: any,
  ): Iterable<K,M>;
}

interface Set<T> extends Iterable<T,T> {
  map<M>(
    mapper: (value: T, key: T, iter: this) => M,
    context?: any,
  ): Set<M>;
}
```

I should be able to define a function that operates on arbitrary `Iterable`s, but if that function maps over the iterable, we end up with unsound behavior.

``` js
function f(iter: Iterable<string, string>): Iterable<string, number> {
  // what does this return? if we pass a set, this function's type is wrong
  // if we pass any other iterable, this function's type is OK
  return iter.map(s => s.length);
}
```

I don't have a solution to this, but one idea is to override `Set`'s `map` function to return an `Iterable` and not a `Set`. This would make `map` also obey a commonly assumed law: `x.map(f).size = x.size`.

`Set` could also provide an alternate `mapSet` function, or similar, that does what `Set.prototype.map` does today.

svieira:
Since a `Set` is "keyed" by number when iterated over, wouldn't it be possible to preserve the soundness of `.map` by simply providing the specific type of `K`:

```js
interface Set<T> extends Iterable<number,T> {
  map<M>(
    mapper: (value: T, key: number, iter: this) => M,
    context?: any,
  ): Set<M>;
}
```

leebyron:
Seems to be fixed in master by moving the definitions of `map` to their subtypes

PR #1285, opened 2017-08-09
Fix typescript types for typescript 2.4

sandersn:
Typescript does a lot more analysis than previous versions. This (1) exposed some bugs in the types and (2) caused the compiler to run out of memory and crash. The crash, and many of the new checks, can be avoided by passing `--noStrictGenericChecks` but users of immutable shouldn't be required to do that.

## Improved Checks
1. The return type of `Stack.flatMap`'s `mapper` was incorrectly `M` but should be `Iterable<M>`.
2. `Set` and related types should extend `Collection<never, T>` not `Collection<T, T>`. That's because `Collection.map` (and flatMap and filter) assert that the returned collection has the same key type and a different value type. But `Set.map` returns `Set<T>`, which, if it extends `Collection<T, T>` changes both the key type and value type. Changing to `Set<T> extends Collection<never, T>` means that the key type never changes.

## Workarounds
1. `Map` and related types have a `flatMap` method that is not a valid `Collection.flatMap` implementation because it changes the key type and value type, much like `Set` (see #1283). However, here the implementation *and* the type are wrong, so I worked around the problem by letting `Map.flatMap` return `Map<any, any>` instead of `Map<KM, VM>` (the previous type) or `Map<K, M>` (a valid subtype of Collection.flatMap).
2. (Fixed in Typescript 2.5.3) In Typescript 2.4, the compiler spends too much time and memory checking the `extends` relationships between `Collection` and its subtypes. For example, `OrderedMap extends Map extends Collection.Keyed extends Collection`. The worst offender is `toSeq` since it refers to a subtype, Seq, which is then extended multiple times by further subtypes. This makes checking goes around in circles multiple times for each subtype. I chose to rely on upcoming optimisations in Typescript 2.5.3 that will short-circuit most of these cycles.

Both workarounds have multiple possible fixes; for example, workaround (1) could instead declare two overloads of `flatMap`, one with the Collection-compatible type and one with the current type. Workaround (2) could delete `toSeq` entirely from `Collection`, or change the return type of toSeq to `any` in `Collection`, or change the return type of `toSeq` in subtypes of Collection.


samwgoldman:
We noticed the issue with Set and map, flatMap as well. (See #879)

sandersn:
The failing tests points out an additional bug with the definition of `Set.has`, which for other collections checks the key but for sets checks for the value.

I'm still working on running tests locally, but I don't know how to copy type-definitions/Immutable.d.ts to dist/immutable-nonambient.d.ts, which is what the tests refer to. I poked around the gulpfile and didn't find anything? Is this manual?

sandersn:
Never mind, I eventually got it by running through the tasks in package.json

emmanueltouzery:
this is could be totally the wrong place to report this (or maybe not?) but I tried to port my immutables&#8209;using typescript app to `typescript@2.6.0-dev.20170826` and the immutables code from this pull request, and I get an unexpected behaviour.

See my commit:
https://github.com/emmanueltouzery/ng-typeview/commit/86ab9e6a7cdb5a920ed778d45be4e52b203964c6

I would have thought the extra cast I added to make it build wouldn't be needed? Without it, it appears to infer type `{}` for `Map<string,string>.get(string)`, instead of `string|undefined` and I have no idea why.

That cast was not needed with tsc 2.3 and the previous immutables TS bindings.

sandersn:
@emmanueltouzery the problem is that `controllerNameToFilename: Map<{}, {}>`. If you manually annotate it as `Map<string, string>`, the compiler error goes away. I'll see why the call to the map constructor doesn't infer the right type arguments.

sandersn:
I tracked the problem back to the preceding declaration:

```ts
    const controllerNameToFilename =
        Map(
            viewInfos
                .filter(vi => vi.controllerName.isSome())
			          // JS files are not going to have a scope interface
			          // definition so they're not helpful. Also, we can
			          // get twice the same file: original TS & compiled JS.
			          // => keep only the original TS in that case.
			          .filter(vi => vi.fileName.toLowerCase().endsWith(".ts"))
                .map(vi => [vi.controllerName.some(), vi.fileName] as [string, string]));
```

Notice that I added **as [string, string]** there at the end. Without it, the return type is inferred as `string[]`, and the wrong overload of `map` gets chosen, and `controllerNameToFilename` gets the type `Map<{}, {}>` instead of `Map<string, string>`. I'm not sure why this worked in previous versions of immutable, though.

leebyron:
Happy to see work here, but some feedback:

* I think all of the changes with `key: never` aren't correct where previously they were correct.
* Ideally we can avoid returning `any` types. What we did for better flow support was just add explicit overrides of methods on the final concrete collection types.

sandersn:
Here's an example of where code goes wrong without `key: never`:
```ts
let nset = Immutable.Set.of(1,2,3,4)
let ncol: Immutable.Collection<number, number> = nset;
nset.map(v => v.toString()).map((v, k) => k.length); 
ncol.map(v => v.toString()).map((v, k) => k.length); 
```

The last two lines produces `Set { 1 }` at runtime, but the last line has a type error because `map(v => v.toString())` produces a `Collection<number, string>` *not* `Collection<string, string>`, so the types say that `k: number` in `map((v, k) => k.length)`

With `key: never` you are prevented from using the key with sets. You have to use the value instead. Both lines have a type error and have to change the second `map` to `map(v => v.length)`.

I understand this is a break for Typescript 2.3 code that uses the keys of a Set today, so if you have guidance I would appreciate it. I'm still thinking about the second case; there might be a way to avoid breaks there.

leebyron:
@sandersn thanks for the example, I think perhaps this would be better solved by a more specific type for `map()` that's aware of a new key type as well on `Set` (and `SeqSet`) rather than blocking the use of keys

sandersn:
OK, I restored the original behaviour for subclasses. I had to do this by adding overloads in the base classes (Seq, Collection, etc) to account for the inconsistent types coming from Set and Dictionary. Instead of a single more specific type, this allows overrides to override one of two different types.

Most of these overloads are reasonable, except for `Collection.map`, because `Set.map` has identical parameter types to `Collection.map`, just a different return type. So the compiler can't distinguish them and always chooses the first one. There I used a hack suggested by @DanielRosenwasser, which is a variadic function `(...args: never[]): any`. Anything is assignable to this, but it won't be called by mistake because values with type never are so rare.

sandersn:
Also, overloads are not inherited, so only uses of the base Collection interface will see the bogus overload. Most people will only ever see the methods that go with the collection type they're using.

leebyron:
Thanks for your help on this. I amended a few changes in the process of merging this, please let me know if any follow ups are necessary

sandersn:
Nope, looks good. I forgot about the test changes for Set that no longer needed to be changed.

PR #1286, opened 2017-08-09
Update Immutable.d.ts

malcolmredheron:
The grammar in the comment about Record was off.


leebyron:
Thanks! Looks like this change was already merged in a previous PR

PR #1303, opened 2017-09-09
Fix type of argument for merge methods in Map

ShyykoSerhiy:
Fixed type of ...collections for merge... methods in Map


leebyron:
I think the correct type for these is probably more like `Iterable<[K, V]> | { [key: K]: V }`

Issue #1308, opened 2017-09-21
Error in the new version of flow 0.55


szagi3891:
Library in version : immutable@4.0.0-rc.2

To get this error just run flow:

```
Error: node_modules/immutable/dist/immutable.js.flow:519
519:   static of<T>(...values: T[]): SetSeq<T>;
                                     ^^^^^^^^^ SetSeq. This type is incompatible with
383:   static of<T>(...values: T[]): IndexedSeq<T>;
                                     ^^^^^^^^^^^^^ IndexedSeq
```

samwgoldman:
@leebyron Oops, sorry I didn't upstream a fix for this, but internally I resolved this in D5708450. On that diff, I wrote:

The Immutable API for sets has a fatal flaw, because Set<T> is defined to be a
subtype of Collection<T,T>. This makes it impossible to define methods on the
base class which modify either K or V in a type-safe way.

We previously ran into this exact issue with `Immutable.Iterable.map`, which
takes an `Iterable<K,V>` and a function `V => M` and produces an
`Iterable<K,M>`. This creates a problem, because `Immutable.Set<T>` is declared
to be a subtype of `Immutable.Iterable<T,T>`. Here's an example of the problem:

```
function f(iter: Iterable<string, string>): Iterable<string, number> {
  // what does this return? if we pass a set, this function's type is wrong
  // if we pass any other iterable, this function's type is OK
  return iter.map(s => s.length);
}
```

We fixed this by removing the declaration of `Iterable.map` and instead
declaring the signatures on every subclass.

This issue follows the same pattern. If `B` is a subtype of `A`, then
`Class<B>` is a subtype of `Class<A>`.

```
function f<K,V>(S: Class<Seq<K,V>>, ...values: V[]): Seq.Indexed<V> {
  return S.of(...values);
}

// We think this returns Seq.Indexed<string> (Seq<number,string>) but it
// actually returns Seq.Set<string> (Seq<string,string>).
f(Seq.Set, "foo", "bar");
```

The fix takes the same form as well: remove the declaration. Since it's an
alias for `Seq.Indexed.of`, I've codemodded all callers to call that method
instead.

szagi3891:
@samwgoldman 
So can we expect this error to be corrected in the version "immutable@4.0.0-rc.3" ? :)

samwgoldman:
[...]

PR #1310, opened 2017-09-27
Remove of<T> for Seq

rgbkrk:
Per https://github.com/facebook/immutable-js/issues/1308#issuecomment-331061620

Hope we see this in the next release candidate!


leebyron:
Thanks!

rgbkrk:
Thank you!!!

PR #1312, opened 2017-09-29
Update to latest version of flow

leebyron:


leebyron:
Investigating the incompatibilities

PR #1315, opened 2017-09-29
fixed #1313 by fixing two borked reducers

### LINKED ISSUE #1313: Broken documentation links.
The links on http://facebook.github.io/immutable-js/docs/ currently seem to be broken. Nothing happens when clicking the links such as `Map`, `List` etc. in the right-menu.

remhume:
Not sure if this is the right way to fix this because i'm unsure about the initial values in the reducers but it definitely fixes #1313 for me and looks fine locally.


leebyron:
Great catch, I like this fix

PR #1317, opened 2017-09-29
fixed #1290 by removing relative position from star count

### LINKED ISSUE #1290: GitHub button clickable on home page
### What happened
The stargazers link on the docs home page has static position and is clickable even when page is scrolled which is both annoying and unexpected.

I played around with `z-index` but was unable to find a working combination. When I remove `position: relative` from `.gh-count`, everything works, too. Tested on Chrome on Mac.

### How to reproduce
out

remhume:
Weird bug, seems to be a browser bug which doesn't occur with ```position: relative```.

### DIFF
--- a/pages/src/src/StarBtn.js
+++ b/pages/src/src/StarBtn.js
@@ -29,6 +29,7 @@ var StarBtn = React.createClass({
           <span className="gh-ico" />
           <span className="gh-text">Star</span>
         </a>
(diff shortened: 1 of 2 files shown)


remhume:
also not sure why i don't get the cla-signed badges, since i did that about 2 hours ago and the website says i'm already registered when i try to do it again 🙄

leebyron:
Looks great to me, thanks for your help investigating this!

PR #1318, opened 2017-09-29
Upgrade prettier and jest

philipp-spiess:
I figured that now after @leebyron worked hard on getting most of the outstanding PRs merged, It would be a good point to update the `prettier` dependency.

The version jump from `0.22.0` to `1.7.2` touches quite a few files so we should probably merge this PR fast.

I also upgraded `jest` while at it since it doesn't require any changes.


leebyron:
Trying to get to PR-zero, saw the number go back up and went:



Then saw what's in here:

PR #1323, opened 2017-09-30
Remove dist/ from repo

leebyron:
This adds dist/ to the gitignore, removing the need to build before submitting a PR and making merge conflicts easier to resolve.

This comes at the cost of bower support, since bower relied on these checked in resources. Existing tagged commits still exist with the bower.json file intact and the #npm branch can still be used as a latest bower.json point of access.


philipp-spiess:
There's also a note in the contribution guide that mentions that a contributor should check-in the resulting artifacts. This needs to be removed as well I guess 😊 

(https://github.com/facebook/immutable-js/blob/master/.github/CONTRIBUTING.md#pull-requests Point 6.)

leebyron:
Good catch :)

leebyron:
This is "step 2" in https://bower.io/blog/2017/how-to-drop-bower-support/

Existing tags in the repo will continue to work, but anyone who was depending on `"*"` or `"master"` will need to fix to a specific tag.

PR #1324, opened 2017-09-30
Stop the iteration of an exhausted, unknown-sized sequence when slicing

### LINKED ISSUE #1256: Seq iterator has an undefined size, causing it to access elements that don't exist
### What happened
```
/Users/neel/Encircle/encircle/client/node_modules/immutable/dist/immutable.js:3238
          return iteratorValue(type, iterations - 1, step.value[1], step);
                                                               ^

TypeError: Cannot read property '1' of undefined
    at Iterator.next (/Users/neel/Encircle/encircle/client/node_modules/immutable/dist/immutable.js:3238:64)
    at Iterator.next (/Users/neel/Encircle/encircle/client/node_modules/immutable/dist/immutable.js:3025:29)
    at Object.<anonymous> (/Users/neel/Encircle/encircle/client/test.js:16:18)
    at Module._compile (module.js:573:32)
    at Object.Module._extensions..js (module.js:582:10)
    at Module.load (module.js:490:32)
    at tryModuleLoad (module.js:449:12)
    at Function.Module._load (module.js:441:3)
    at Module.runMain (module.js:607:10)
    at run (bootstrap_node.js:382:7)
```

I was able to minimise this issue down to the following case:
```
Immutable = require('immutable')

var seq = Immutable.Seq([1, 2, 3, 4]);
seq = seq.flatMap(function(elem) {
    return [elem, elem];
}).slice(0, 20).map(function(elem) {
    return true;
});

// At this point seq is an `IndexedIterator` with an undefined size

var iter = seq.values();
var done = false;

while (!done) {
    var next = iter.next();
    done = next.done;
}
```

After a bit of debugging the issue seemed to be that when a `Seq` is sliced it creates an `IndexedIterator` and in the current configuration (a `flatMap` followed by a `slice` followed by a `map`) it causes this `iterator` to not have a size value. Due to this reason, creating an `iterator` on this causes issues as the internals of `Immutable` try to access the value of the next element which doesn't exist.

philipp-spiess:
Fixes #1256

This PR fixes a bug with the slice operator that caused non-value
iterators of an unknown sized sequence to continue although the source
sequence was exhausted, eventually accessing the value of a key/value
pair that does not exist.

The issue did not happen with fixed sized sequences since we can
calculate resulting size in advance (as `sliceSize`) and bail out early.
For an unknown sized sequence, however, we can not know this size in
advance (see this comment: [1]).

To solve this, we now bail out when we encounter a completed iterator
before proceeding to process it further.

[1]: https://github.com/facebook/immutable-js/blob/27b231c8952366ddb96ab2a20d1eb351b3c8719e/src/Operations.js#L427-L428


leebyron:
Great, thanks for adding a test!

PR #1325, opened 2017-10-01
Fix a bug in the hashString function resulting in all string hashes being 0

errendir:
I believe when variables were being renamed in https://github.com/facebook/immutable-js/pull/1279 one identifier name was not changed, thus altering the behaviour of the `hashString` function. The current implementation returns 0 for all strings making the hashing process pointless.

This PR restores the intended algorithm of the `hashString` function.

I found this issue while testing the performance of `Immutable.Map` between versions 4.0.0-rc.3 and 3.8.1. When using integer keys there was no noticeable difference, but using string keys resulted in a huge performance loss in 4.0.0-rc.3.

An analogous issue is present in the `hashJSObj` function. This PR fixes it as well.


leebyron:
You're absolutely right! Wow it's bad this was not caught by a test. I'll get this out asap

PR #1339, opened 2017-10-04
BREAKING: no longer deeply coerce argument to merge()

### LINKED ISSUE #1293: Record#merge() converts argument to Immutable structure
Calling `Record#merge()` will convert the argument passed in into an Immutable structure, yielding the same output as calling `fromJS()` with it. Since instantiating a `Record` directly does not do this, it seems to be a bug.

Example:

```
const State = Immutable.Record({ foo: 'bar', biz: 'baz' });
const deepObject = { qux: 'quux' };

const firstState = new State({ foo: deepObject });
const secondState = new State().merge({ foo: deepObject });

console.log(firstState.toString());
console.log(secondState.toString());
console.log(firstState.equals(secondState));
```

Output:

```
> "n { 'foo': [object Object], 'biz': 'baz' }"
> "n { 'foo': Map { 'qux': 'quux' }, 'biz': 'baz' }"
> false
```

Expected:

```
> "n { 'foo': [object Object], 'biz': 'baz' }"
> "n { 'foo': [object Object], 'biz': 'baz' }" 
> true
```

leebyron:
The methods `merge()` and `mergeDeep()` were the last places internally to call `fromJS()`, a method that is expensive, and limits the ability to use plain object and array values within immutable collections.

This PR removes fromJS, not doing any deep coercion at all for `merge()` and simply passing object (incl array) values into the `merge()` function recursively for `mergeDeep()` operations, so only coercing those values that expect to be - leaving others alone.

This is breaking because calls to `merge()` previously were implicitly `fromJS()`'d to get deeply immutable collections, however that is no longer the case.

To update code which relied on this behavior, convert `collection.merge(x)` to `collection.merge(fromJS(x))`.

Fixes #1293

### DIFF
--- a/__tests__/issues.ts
+++ b/__tests__/issues.ts
@@ -59,3 +59,15 @@ describe('Issue #1247', () => {
     expect(r.wasAltered()).toBe(false);
   });
 });
+
+describe('Issue #1293', () => {
+  it('merge() should not deeply coerce values', () => {
+    const State = Record({ foo: 'bar' as any, biz: 'baz' });
+    const deepObject = { qux: 'quux' };
+
+    const firstState = State({ foo: deepObject });
+    const secondState = State().merge({ foo: deepObject });
+
+    expect(secondState).toEqual(firstState);
+  });
+});
--- a/__tests__/merge.ts
+++ b/__tests__/merge.ts
@@ -105,7 +105,7 @@ describe('merge', () => {
   it('can overwrite existing maps', () => {
     expect(
       fromJS({ a: { x: 1, y: 1 }, b: { x: 2, y: 2 } })
-        .merge({ a: null, b: { x: 10 } })
+        .merge({ a: null, b: Map({ x: 10 }) })
         .toJS(),
     ).toEqual(
       { a: null, b: { x: 10 } },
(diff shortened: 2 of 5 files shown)


renanbandeira:
Hi, I've been using Immutable for years and I have about 500 merge calls in my project that I used to prevent transformin plain objects in immutable before an update or set. Is there any way I could pre process those merge and mergeDeep, mergeDeepIn, mergeIn, mergeDeepWith converting the argument in immutable without doing by hand?

Thanks

leebyron:
Yes you can call fromJS() on any deep object tree, this is what merge was doing implicitly in v3, but v4 allows the use of plain JS structures with functional immutable methods

renanbandeira:
Ok, but is there any script to make this change automatically or I should change every one of the 500 merge usages in my code?

leebyron:
I’m not aware of a script, but if you’d like to contribute one, that would be a great positive impact for the community! Jscodeshift could be a good place to start

renanbandeira:
That sounds great! I will take a look at Jscodeshift in order to contribute. Thanks!

renanbandeira:
Hi @leebyron, I just created with one:  https://github.com/renanbandeira/jscodeshift-upgrade-immutable

Since this might be useful for projects that haven't upgraded to v4 yet for this reason, I will be working on the other merge calls support.

Thanks for the tip.

PR #1340, opened 2017-10-04
BREAKING: KeyedCollection.toArray() returns array of tuples.

### LINKED ISSUE #333: toArray behavior on Maps
[This ticket](https://github.com/moreartyjs/moreartyjs/issues/58#issuecomment-74389910) was raised over in Morearty.js regarding an incorrect example of using transit with OrderedMaps.

I was looking through the tickets and didn't see any questions raised so I figured I'd ask, is the API of `toArray` on Map / OrderedMap not preserving the map entries as k/v tuples intentional?

leebyron:
This fixes a long standing API quirk where `x.toArray()` had different behavior from `[...x]` or `Array.from(x)`.

This is breaking since previously Map, and other keyed collections would provide an array of only values from `.toArray()`, to fix code that relies on this behavior, convert `x.toArray()` to `x.valueSeq().toArray()`.

Likewise, for code that was working around this behavior before, `x.entrySeq().toArray()` can typically be replaced with `x.toArray()`.

Fixes #333

### DIFF
--- a/__tests__/ObjectSeq.ts
+++ b/__tests__/ObjectSeq.ts
@@ -32,13 +32,13 @@ describe('ObjectSequence', () => {
   it('is reversable', () => {
     let i = Seq({a: 'A', b: 'B', c: 'C'});
     let k = i.reverse().toArray();
-    expect(k).toEqual(['C', 'B', 'A']);
+    expect(k).toEqual([['c', 'C'], ['b', 'B'], ['a', 'A']]);
   });
 
-  it('can double reversable', () => {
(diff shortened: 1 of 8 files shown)


msageryd:
Thank you for your great efforts on a great library. I'm trying out version 4 and it's great.

Just wanted to give you a heads-up on the new toArray() behaviour. I bit me in several places in my code base. I think this one needs to be shout out properly in the documentation. A common place for my use of toArray() is in ReduxSaga when I'm forking tasks based on data in Immutable.Map.

PR #1341, opened 2017-10-04
Remove the custom "is" matcher in favor of "toEqual"

leebyron:
jest has dramatically improved over the years, and supports Immutable values directly! No need for our custom checker.


rgbkrk:
This is great to know for nteract's code base. Thanks @leebyron! /cc @lgeiger

PR #1350, opened 2017-10-05
Add `getIn` and `hasIn` methods to Record flow type

torjusb:
Copied these from `_Collection` type. Little unsure if they should be typed as `mixed` vs `any`?

### DIFF
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -1341,6 +1341,8 @@ declare class RecordInstance<T: Object> {
   clear(): this & T;
 
   setIn(keyPath: Iterable<any>, value: any): this & T;
+  getIn(searchKeyPath: Iterable<mixed>, notSetValue?: mixed): any;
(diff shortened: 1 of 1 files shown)


leebyron:
These look great! Could you add them to the .d.ts file as well?

torjusb:
Looks like they are in already, unless I'm mistaken? https://github.com/facebook/immutable-js/blob/master/type-definitions/Immutable.d.ts#L2534-L2535

leebyron:
You're right! Glad to see this discrepancy fixed, thanks for the help!

leebyron:
Got this released as rc.6 in case you're looking to use this fix!

torjusb:
Great. Thank you for the quick response 😃

torjusb:
@leebyron, it doesn't appear to be on yarnpkg.com or npm yet. Does it just take some time before the update propagates?

leebyron:
Yikes, looks like it failed when I published and I didn't notice. Thanks for the heads up, I have versions out on npm

PR #1354, opened 2017-10-05
Fix typos in typescript definitions

TheophileSketo:


leebyron:
Wow, how were the ts tests passing!?

Thanks for this

PR #1362, opened 2017-10-06
V4 get in fix

### LINKED ISSUE #1361: [ v4 ]  getIn throws deprecation warning when encountering null values.
### What happened

Immutable v4 introduced a bug in `getIn` (and other methods as well, like `hasIn`) where attempting to continue on a list of keys after a `null` value was found would result in a deprecation warning.  However, per discussion in #1161 this is not intended.

### How to reproduce

```js
const DB = Immutable.fromJS({ users: [{ name: 'Kyle', pets: null }] });

function getUserPet(userIndex, petIndex) {
  return DB.getIn(['users', userIndex, 'pets', petIndex, 'name'], 'no pet');
}

getUserPet(0,0); // throws warning -- Value at ["users",0,"pets"] does not have a .get() method: null
```

**Expected** although `pets` is null, the expected functionality would be that the `getIn` request will not throw a warning, but rather continue the v3 functionality.

**See also**: https://codepen.io/Pink401k/pen/NayNPp

Kyle-Mendes:
resolves #1361 

A pretty simple fix.  However, it might not be 100% how the Immutable team wants to continue.

There's a default `NOT_SET` constant in `TrieUtils.js`.  [Here](https://github.com/facebook/immutable-js/blob/master/src/TrieUtils.js#L18)

Since `NOT_SET = {}` declaring the `notSetValue` for `null` and `undefined` was left up to the `warn` call in `getIn`.  

However, based on the conversation in #1161, we can safely assume that `undefined` and `null` should be counted as `NOT_SET` values as well.

So, this PR resolves #1361 by including those values in the `NOT_SET` conditonal.

-------------

### Potential Improvements
Perhaps, a better (larger) solution would be to make `NOT_SET` a function that can compare against all `NOT_SET` values (including `null` and `undefined`) so that this conditional can be brought back down to one statement, and this logic wouldn't have to be repeated elsewhere in the code base.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ TODO
 /gh-pages
 /npm
 /dist
+tags
--- a/__tests__/CollectionImpl.ts
+++ b/__tests__/CollectionImpl.ts
@@ -1,0 +1,33 @@
+/**
+ * Copyright (c) 2014-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+///<reference path='../resources/jest.d.ts'/>
+
+import * as jasmineCheck from 'jasmine-check';
+jasmineCheck.install();
+
+import { fromJS, Map } from '../';
+
+describe('CollectionImpl', () => {
+  it('should allow getIn past null values', () => {
+    console.warn = jest.genMockFunction();
+
+    const m = Map({a: { b: null }});
+
+    expect(m.getIn(['a', 'b', 'c', 'd'], 'notSetValue')).toBe('notSetValue');
+    expect(console.warn.mock.calls.length).toBe(0);
+  });
+
+  it('should allow hasIn past null values', () => {
+    console.warn = jest.genMockFunction();
+
+    const m = Map({a: { b: null }});
+
+    expect(m.hasIn(['a', 'b', 'c', 'd'])).toBeFalsy();
+    expect(console.warn.mock.calls.length).toBe(0);
+  });
+});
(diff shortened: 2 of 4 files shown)


Kyle-Mendes:
@leebyron I think I'm struggling to understand how to fix this then.  Take this example:

```js
const collection = Immutable.fromJS({a: {b: null}});

// ---------------------------------v [ value is null] 
console.log(collection.getIn(['a', 'b', 'c'], 'notSetValue'); // should return string: 'notSetValue'
```

As I understand it, the fix for this issue (#1361) is to make it so that when **value is falsy** or **value doesn't have an internal get method** BUT **that value is null or undefined** we want to return the `notSetValue`.  Is that correct?

If so, the only other fix I can think of here is introducing a new conditional, above line 851 that says:

```js
if (value === null || value === undefined) {
  return notSetValue;
}
```

However, this fix seems to have the same result as the current change in this PR

-----------

In Immutable v3, using `getIn` where an intermediary key in the path was `null` would return the `notSetValue` which is what I'm trying to allow here :)

.gitignore
```diff
@@ -12,3 +12,4 @@ TODO
 /gh-pages
 /npm
 /dist
+tags
```
leebyron: Could you remove this unrelated change?

leebyron:
I just put up #1367 which improves the test coverage of getIn and hasIn with tests that would fail for this PR which helps to illustrate my earlier feedback, perhaps you could rebase to include that?

leebyron:
I borrowed your test and tested out a similar change in #1368 - that will both fix this new test while preserving a pass for the rest!

PR #1366, opened 2017-10-09
Flow types for getIn()

leebyron:
This adds a flow type for getIn() using the new $Call type to define two new utility type operators: `$KeyOf` and `$ValOf` which are similar to `$ElementType` but work on both Records and Collections.

To keep this PR small, this only addresses `getIn()` - though in a follow up will do other `*In` methods.

### DIFF
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -1409,7 +1409,7 @@ declare class RecordInstance<T: Object> {
 
   hasIn(keyPath: Iterable<mixed>): boolean;
 
-  getIn(keyPath: [], notSetValue?: mixed): this;
+  getIn(keyPath: [], notSetValue?: mixed): this & T;
   getIn<K: $Keys<T>>(keyPath: [K], notSetValue?: mixed): $ElementType<T, K>;
   getIn<NSV, K: $Keys<T>, K2: $KeyOf<$ElementType<T, K>>>(keyPath: [K, K2], notSetValue: NSV): $ValOf<$ElementType<T, K>, K2> | NSV;
   getIn<NSV, K: $Keys<T>, K2: $KeyOf<$ElementType<T, K>>, K3: $KeyOf<$ValOf<$ElementType<T, K>, K2>>>(keyPath: [K, K2, K3], notSetValue: NSV): $ValOf<$ValOf<$ElementType<T, K>, K2>, K3> | NSV;
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -32,6 +32,21 @@
 // some constructors and functions.
 type PlainObjInput<K, V> = {[key: K]: V, __proto__: null};
 
+// Helper types to extract the "keys" and "values" use by the *In() methods.
+// Exported for experimental use only, may change.
+export type $KeyOf<C> = $Call<
+  & (<K>(?_Collection<K, *>) => K)
+  & (<T>(?RecordInstance<T>) => $Keys<T>),
+  C
+>;
+
+export type $ValOf<C, K = *> = $Call<
+  & (<V>(?_Collection<*, V>) => V)
+  & (<T, K: $Keys<T>>(?RecordInstance<T>, K) => $ElementType<T, K>),
+  C,
+  K
+>;
+
 declare class _Collection<K, +V> /*implements ValueObject*/ {
   equals(other: mixed): boolean;
   hashCode(): number;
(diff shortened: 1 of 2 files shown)


type-definitions/immutable.js.flow
```diff
@@ -43,9 +58,15 @@ declare class _Collection<K, +V> /*implements ValueObject*/ {
...
-  getIn(keyPath: Iterable<mixed>, notSetValue?: mixed): any;
   hasIn(keyPath: Iterable<mixed>): boolean;
 
+  getIn(keyPath: [], notSetValue?: mixed): this;
+  getIn<NSV>(keyPath: [K], notSetValue: NSV): V | NSV;
+  getIn<NSV, K2: $KeyOf<V>>(keyPath: [K, K2], notSetValue: NSV): $ValOf<V, K2> | NSV;
+  getIn<NSV, K2: $KeyOf<V>, K3: $KeyOf<$ValOf<V, K2>>>(keyPath: [K, K2, K3], notSetValue: NSV): $ValOf<$ValOf<V, K2>, K3> | NSV;
+  getIn<NSV, K2: $KeyOf<V>, K3: $KeyOf<$ValOf<V, K2>>, K4: $KeyOf<$ValOf<$ValOf<V, K2>, K3>>>(keyPath: [K, K2, K3, K4], notSetValue: NSV): $ValOf<$ValOf<$ValOf<V, K2>, K3>, K4> | NSV;
+  getIn<NSV, K2: $KeyOf<V>, K3: $KeyOf<$ValOf<V, K2>>, K4: $KeyOf<$ValOf<$ValOf<V, K2>, K3>>, K5: $KeyOf<$ValOf<$ValOf<$ValOf<V, K2>, K3>, K4>>>(keyPath: [K, K2, K3, K4, K5], notSetValue: NSV): $ValOf<$ValOf<$ValOf<$ValOf<V, K2>, K3>, K4>, K5> | NSV;
+
```
leebyron: One thing I'm not sure about is not having a case for supporting keyPaths with an unknown or more than 5 elements. Adding back `getIn(keyPath: Iterable<mixed>, notSetValue?: mixed): any;` causes some of the `$ExpectError` to no longer trigger when checking for valid key paths.

type-definitions/immutable.js.flow
```diff
@@ -32,6 +32,21 @@
 // some constructors and functions.
 type PlainObjInput<K, V> = {[key: K]: V, __proto__: null};
 
+// Helper types to extract the "keys" and "values" use by the *In() methods.
+// Exported for experimental use only, may change.
+export type $KeyOf<C> = $Call<
+  & (<K>(?_Collection<K, *>) => K)
```
calebmer: Instead of `*`, I recommend using `_Collection<K, mixed>` here 😊

type-definitions/immutable.js.flow
```diff
@@ -32,6 +32,21 @@
...
+  & (<T>(?RecordInstance<T>) => $Keys<T>),
+  C
+>;
+
+export type $ValOf<C, K = *> = $Call<
```
calebmer: Would `K = $KeyOf<C>` work here? `K = string` may also work well.

type-definitions/immutable.js.flow
```diff
@@ -32,6 +32,21 @@
...
+  & (<V>(?_Collection<*, V>) => V)
```
calebmer: I recommend `_Collection<any, V>` here 😊

type-definitions/immutable.js.flow
calebmer: I’ve been wanting to introduce a `$Reduce<>` utility type. Such a type would allow you to support key paths of any length 😊. The definition of `getIn()` with `$Reduce<>` may look something like:

```
getIn<Path, NSV>(
  keyPath: Path,
  notSetValue: NSV,
): $Reduce<Path, <Acc, K>(Acc, K) => $ValOf<Acc, K>, this> | NSV;
```

type-definitions/immutable.js.flow
leebyron: Tried using `mixed` at first, but get errors about value types that don't flow to mixed. I may try `any` instead

type-definitions/immutable.js.flow
leebyron: Great idea to use `$KeyOf<C>` as a default

leebyron:
Thanks for the review, @calebmer. Looking forward to your work on `$Reduce`!

Issue #39, opened 2014-08-06
Functional API


joakin:
I've been playing with the library and missed a functional API instead of a class based method one. To play with it i've made a little wrapper around the library that just adds the static methods to the classes making the last param the object on which the method will be invoked on.

[npm - immutable-fns](https://www.npmjs.org/package/immutable-fns)

That way we can do things like:

``` javascript
// On the immutable-js api
Immutable.Map({}).get('key')
// Added on this library
Immutable.Map.get('key', map)
```

``` javascript
var Immutable = require('immutable-fns');
var get = Immutable.Map.get;

// Create an immutable data structure
var m = Immutable.Map({a: 1});

// We can use the traditional api (obj.method(param)) or use the functional API
// (fn(...params, obj))
assert.equal(m.get('a'), get('a', m));
```

This kind of API is really useful sometimes, and doesn't conflict with the current API (that I have seen). This wrapper may serve to evaluate if this is worth pursuing on the library itself, and it does its job for the moment.

Thanks for your work on the immutable data structures, really hope to see them take off in js-land.

ccorcos:
I'd love to see a functional API for immutable js!

bwestergard:
Any developments on this?

ccorcos:
Well this is an interesting option...

https://github.com/paldepind/functionize

MoeSattler:
I'd love this. If there any current solution to a more functional API? I miss `pipe` and `compose` so much. 
Would definetly be up to contribute to a project tackling this.

mikeaustin:
On a side note, what are people using to pair with immutable for compose, curry, flip, always/constant, etc? Ramda seems big and complicated. Crocks looks interesting, and also has ATDs such as Either and Maybe.

iansinnott:
Many of Ramda's collection functions work quite well with immutable actually, since they dispatch to the relevant method if present. For example `R.map` and `R.filter`. That doesn't help with the relevant getters and setters of immutable, but you can use `R.lens` to gain at least some of the functionality you're looking for. [Here's a relevant article](https://medium.com/@drboolean/lenses-with-immutable-js-9bda85674780). This means you can build up functional pipelines using `R.compose` or `R.pipe` that work on immutable objects.

There's definitely a lot going on in Ramda but you can still import only what you need directly from the `src/` directory. Not as convenient as lodash but I believe there's a babel plugin which can automate this. Manual example:

```js
import map from 'ramda/src/map';
```

[...]

PR #1369, opened 2017-10-10
BREAKING: Support for Array and plain Object (adds functional API)

### LINKED ISSUE #1268: Feature Request: Functional API (wrapper around the OOP API)
I realize this was discussed a few years ago, but there's still no functional API. Is there any way we could revive discussion about this?

It's the most painful thing about using Immutable.js for me. I realize the decision to use an OOP API was made in order to be more like native JS, and, therefore, easier for people to get into, but it's a real problem, because it gets in the way of using pipe/compose, currying, partial application, etc.... It becomes a barrier to introducing teammates to those concepts, because there's a major conflict between Immutable & those techniques. So, if they are unfamiliar with currying, for example, but are already using Immutable, it becomes really difficult to sell them on the idea at all.

Ramda & lodash/fp are quite popular nowadays, and more people than ever before are using FP techniques when writing JavaScript now. There seems to be an initiative at Facebook to guide the JS community further into FP (look at what's going on with Reason), but not offering a functional API for a very popular library gets in the way of this for people who are entrenched in the React/Redux/Immutable/etc. world already.

I realize that this wrapper could be created by a 3rd party, but it makes so much more sense to include it as an optional API directly in Immutable, because it would help expose more people to currying, composition, partial application, etc.... They would have less fear of learning new concepts, because they would see a major, popular, Facebook-affiliated library offering an API that encourages the use of them.

There's a real opportunity here, as there aren't many alternatives out there. Mori, for example, is not maintained. It's not kept up to date with updates to ClojureScript. The only other options out there are relatively obscure/unknown to most people.

I'm not even asking for an auto-curried API (although that would be nice). We could do that on the app side. Just having functional wrappers for the data structure methods would be good enough.

Thoughts?

### LINKED ISSUE #1363: Request: Put deprecation warnings behind `process.env === production`
I love the new `console.warn` deprecation notices from `immutable`. Would it possible to put them behind a `process.env === "production"` flag though so they don't get posted when applications are built in production mode? Similar to how React does it.

leebyron:
**BREAKING**: This changes the behavior of a few common methods with respect to Arrays and plain Objects:

- `.merge()` - where previously plain object or array values were considered opaque and replaced completely, they are now merged with the incoming value.

- `.getIn()`/`.setIn()`/`.updateIn()` - where previously key paths that encounter plain object or array values would throw an error, replace a value, or return undefined, they now key into the plain object or array values.

- `.toJS()` - where previously plain object and array values were considered opaque and returned by reference, toJS() will now return copies of such values, and recurse into them converting nested immutable collections to plain JS values.

---

This adds a functional API for the core read/write methods (get, set, update, merge) (originally proposed in #39). I don't believe that simply providing a functional alternative is enough value to warrant the increase in API surface area, but what this functional API enables is to use the immutable read/write API on typical Array and Object values, allowing for a combination of nesting Immutable Collections and typical Arrays and Objects while preserving the ability to use `getIn` and `setIn`.

As an example:

```js
import { getIn, setIn } from 'immutable';
const data = { x: { y: 'z' } }
const z = getIn(data, ['x', 'y'])
const data2 = setIn(data, ['x', 'y'], 'abc') // { x: { y: 'abc' } }
// Original is unchanged
console.log(data); // { x: { y: 'z' } }
```

I think that providing this flexibility will allow for a more reasoned use of Immutable collections based on best fit, rather than the "all or nothing" approach often taken. Specifically, I'm interested in supporting typical Objects as a viable alternative to Record for those that want to use Immutable collections but prefer not to use the Record API.

This does add some byte-weight to the library, but only marginally due to replacing existing implementations. Adds +209 gz-bytes, or +1167 raw-bytes.

Remaining tasks:

* [x] Fix existing tests
* [x] Add new tests for new use cases 
* [x] Add flow types
* [x] Add flow type tests
* [x] Add typescript types
* [x] Add typescript type tests
* [x] Documentation

Closes #1268 
Closes #1363

### DIFF
--- a/src/functional/getIn.js
+++ b/src/functional/getIn.js
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) 2014-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import coerceKeyPath from '../utils/coerceKeyPath';
+import { NOT_SET } from '../TrieUtils';
+import { get } from './get';
+
+export function getIn(collection, searchKeyPath, notSetValue) {
+  const keyPath = coerceKeyPath(searchKeyPath);
+  let i = 0;
+  while (i !== keyPath.length) {
+    // Intermediate null/undefined value along path
+    if (!collection) {
+      return notSetValue;
+    }
+    collection = get(collection, keyPath[i++], NOT_SET);
+    if (collection === NOT_SET) {
+      return notSetValue;
+    }
+  }
+  return collection;
+}
--- a/src/functional/getIn.js
+++ b/src/functional/getIn.js
@@ -13,10 +13,6 @@ export function getIn(collection, searchKeyPath, notSetValue) {
   const keyPath = coerceKeyPath(searchKeyPath);
   let i = 0;
   while (i !== keyPath.length) {
-    // Intermediate null/undefined value along path
-    if (!collection) {
-      return notSetValue;
-    }
     collection = get(collection, keyPath[i++], NOT_SET);
     if (collection === NOT_SET) {
       return notSetValue;
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -33,6 +33,7 @@
     "no-param-reassign": "off",
     "no-plusplus": "off",
     "no-prototype-builtins": "off",
+    "no-restricted-syntax": "off",
     "no-return-assign": "off",
     "no-self-compare": "off",
     "no-sequences": "off",
--- a/__tests__/Seq.ts
+++ b/__tests__/Seq.ts
@@ -85,4 +85,16 @@ describe('Seq', () => {
     expect(list.take(NaN).toJS()).toEqual([]);
   });
 
+  it('Converts deeply toJS after converting to entries', () => {
+    const list = Seq([Seq([1, 2]), Seq({a: 'z'})]);
+    expect(list.entrySeq().toJS()).toEqual(
+      [[0, [1, 2]], [1, {a: 'z'}]],
+    );
+
+    const map = Seq({x: Seq([1, 2]), y: Seq({a: 'z'})});
+    expect(map.entrySeq().toJS()).toEqual(
+      [['x', [1, 2]], ['y', {a: 'z'}]],
+    );
+  });
+
 });
--- a/__tests__/getIn.ts
+++ b/__tests__/getIn.ts
@@ -36,7 +36,7 @@ describe('getIn', () => {
     ).toThrow('Invalid keyPath: expected Ordered Collection or Array: abc');
     expect(() =>
       getIn(Map(), 'abc' as any),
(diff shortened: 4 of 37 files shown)


jpollard-cs:
This is awesome! Why not have the data/collections come after the other arguments to be more curry friendly?

leebyron:
> Why not have the data/collections come after the other arguments to be more curry friendly?

It's a tradeoff.

Use of curry is still pretty uncommon in JavaScript programs, so the argument order intends to better fit the most commonly expected order from JavaScript devs instead. I don't think it would be wise to choose a different order than similar methods in `lodash`, and cause confusion or a difficult learning curve for most. Curry-friendly argument order is nice to have, but not nice enough to warrant a worse learning curve, in my opinion.

Also, if I understand correctly the datastructure-last argument order for curry friendliness makes the most sense within languages which have syntactic support for currying. For languages that do not, like JS, the order should play a less important role - you'll need to use some explicit function to curry anyhow. For example:

```js
const first = coll => get(coll, 0);
first(['a','b','c'])
```

Or if you use a function like `curry` from `ramda` to create curryable versions of functions, just use that opportunity to re-order the arguments, since you'll need to explicitly call it anyhow. For example:

```js
const get = curry((key, collection) => Immutable.get(key, collection));
get(0)(['a', 'b', 'c'])
```

jpollard-cs:
True - one could always provide a wrapper function. You make good points. I guess I was under the assumption that if you're going out of your way to use a functional API you would expect the data to come last, but perhaps Lodash FP is more popular than I think ;)

src/functional/getIn.js
```diff
@@ -0,0 +1,26 @@
...
+import coerceKeyPath from '../utils/coerceKeyPath';
+import { NOT_SET } from '../TrieUtils';
+import { get } from './get';
+
+export function getIn(collection, searchKeyPath, notSetValue) {
+  const keyPath = coerceKeyPath(searchKeyPath);
+  let i = 0;
+  while (i !== keyPath.length) {
+    // Intermediate null/undefined value along path
+    if (!collection) {
```
danprince: Seems like this will cause problems with other falsy values. For example:

```js
let names = ["foo", "bar", ""];

names.map(name => getIn(name, "length"));
// will return
[3, 3, undefined]
```

Maybe it'd be better to use `collection == null` or be completely explicit with `collection === null || collection === undefined`.

leebyron: Thanks for the review! You're right that this is a strange outcome. Instead, I think getin() should treat non-data structure values as opaque, which will control the set of outcomes. I'll follow up

mattzeunert:
Could all the affected functions be added to the changelog? I ran into this with `toJS`, but only `merge` and `setIn` are mentioned.

WorldMaker:
I wanted this functionality in a couple places, but didn't quite realize how much it would impact elsewhere in the existing application as I was using POJOs as "merge barriers" in a bunch of places. I'm trying to square how to mix both functionality now. Now that POJOs are no longer "merge barriers", I'm wondering if there needs to be some other way to set explicit "barriers" when using something like `mergeDeep`?

PR #1373, opened 2017-10-10
BREAKING: Alias `merge` as `concat` for Set and Map collections

leebyron:
These methods are extremely similar, but have differing implementations with different performance characteristics. This makes `concat` an alias of `merge`, sharing the implementation for those methods.

### DIFF
--- a/__tests__/List.ts
+++ b/__tests__/List.ts
@@ -642,10 +642,34 @@ describe('List', () => {
   });
 
   it('concat works like Array.prototype.concat', () => {
-    const v1 = List.of(1, 2, 3);
+    const v1 = List([1, 2, 3]);
     const v2 = v1.concat(4, List([ 5, 6 ]), [7, 8], Seq([ 9, 10 ]), Set.of(11, 12), null as any);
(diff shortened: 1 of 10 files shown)


randallb:
I upgraded to RC9 and at runtime, orderedMaps can't concat. Concat returns as undefined. Working on a test case now.

randallb:
Given a package.json:
```
{
  "name": "immmutable-bug-simple",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "immutable": "^4.0.0-rc.9"
  }
}
```
and an index.js of
```
const immutable = require('immutable');

const orderedMap = immutable.OrderedMap();
const orderedMap2 = immutable.OrderedMap();
orderedMap.concat(orderedMap2)
```
I get:
```
TypeError: orderedMap.concat is not a function
    at Object.<anonymous> (/Users/rb/code/experiment/immmutable-bug-simple/index.js:5:12)
```

randallb:
Working to see if I can help write code to fix.

randallb:
Opening a new issue too.

PR #1377, opened 2017-10-11
Flow types for setIn(), removeIn(), updateIn()

leebyron:
Following up #1366, adds flow types aware of keypath for these three methods for List, Map, and Record


sambathl:
why tuple for updateIn, what shall we do if we resolve keypath dynamically ?

PR #1378, opened 2017-10-12
Add Record.get(v, notSetValue) to type defs

### LINKED ISSUE #1359: Record typedef for get doesn't include notSetValue
Hi,

Trying out 4.0 rc-7. Noticed the new Record types (awesome). But also noticed Record.get() doesn't type the notSetValue so flow complains if I do
```
r.get('myProp', 123)
```
even though it actually works

Thanks! (super excited for this release, keep up the good work)

leebyron:
notSetValue was accidentally missing from type definition for Record. Also include some documentation explaining when this value is used.

Fixes #1359

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -2552,7 +2552,16 @@ declare module Immutable {
     // Reading values
 
     has(key: string): key is keyof TProps;
-    get<K extends keyof TProps>(key: K): TProps[K];
+
+    /**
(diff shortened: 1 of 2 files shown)


Anahkiasen:
@leebyron Should this not be a facultative argument?

julianwong94:
@leebyron +1, shouldn't this be an optional argument? The get functions for the other immutable collections have optional notSetValues.

PR #1393, opened 2017-10-17
BREAKING: Create new List instead of new Map in updateIn keypath.

leebyron:
If any keys in `keyPath` do not exist, a new Map is created. This is often confusing when using numeric keys - which typically imply using a List instead of a Map.

This now looks at the key, if it is a number then a List is created instead of a Map.

This is breaking since previously this would have created a Map instead of a List. Existing code relying on creating `Map<number, *>` should convert code to explicitly create missing Maps before calling `updateIn()`.

### DIFF
--- a/__tests__/updateIn.ts
+++ b/__tests__/updateIn.ts
@@ -7,7 +7,7 @@
 
 ///<reference path='../resources/jest.d.ts'/>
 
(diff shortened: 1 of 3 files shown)


leebyron:
Closing this since an existing use case for numeric keys is representing "ids" - where a Map is a better fit. This breaking change is likely more painful than helpful.

PR #1394, opened 2017-10-17
Adds test to demo key Symbol in Map.set()

abacaphiliac:
Adds test to demonstrate usage of Symbol primitive data type as Map key via set method, which does not work upon Map instantiation.

Split from https://github.com/facebook/immutable-js/pull/1392


abacaphiliac:
@leebyron please review.

I'm working on another request right now for the Map instantiation performance test(s).

leebyron:
Great! thanks :)

PR #1395, opened 2017-10-17
fix: typescript definitions

TheophileSketo:
I had to guess a couple of signatures for the remove function, I hope everything is in order.


leebyron:
Thanks for your help with this. I'm frustrated with the state of our typescript tests, which I expected to help catch these issues.

PR #1399, opened 2017-10-17
Fix flow types for Record setIn/getIn deep values.

### LINKED ISSUE #1398: [v4.0.0-rc.8] Modifying deeply nested Records receives flow type error
Here is a sample reducer that reproduces the issue.  I have some Records that have deeply nested Records
`immutable 4.0.0-rc.8`
`flow-bin 0.57.3`

```
// @flow
import { Record, type RecordOf, type RecordFactory } from 'immutable';

type DeepNest = {
  foo: number,
};

const deepNest: RecordFactory<DeepNest> = Record({
  foo: 0,
});

type Nest = {
  deepNest: RecordOf<DeepNest>,
};

const nest: RecordFactory<Nest> = Record({
  deepNest: deepNest({}),
});

type State = {
  nest: RecordOf<Nest>,
};

const initialState: RecordFactory<State> = Record({
  nest: nest(),
});

function testReducer(state: RecordOf<State> = initialState(), action: Object) {
  switch (action.type) {
    case 'ACTION_TYPE':
      return state.setIn(['nest', 'deepNest', 'foo'], 5);
    default:
      return state;
  }
}

export default testReducer;
```
The error occurs on `return state.setIn(['nest', 'deepNest', 'foo'], 5);` and I receive

```
call of method `setIn`
Method cannot be called on any member of intersection type intersection
Member 1: type application of class `RecordInstance`
Error: call of method `setIn`
Function cannot be called on any member of intersection type intersection
Member 2: type parameter `Values` of call of method `setIn`
Error: property `setIn`
Property not found in object type
```

leebyron:
Unsure what the root issue was, but key types were being incorrectly derived for deeply nested values. Replacing $ElementType with $ValOf seems to fix the issue.

Added the original reported test case that failed before, now passes

Fixes #1398

### DIFF
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -1397,10 +1397,10 @@ declare class RecordInstance<T: Object> {
 
   getIn(keyPath: [], notSetValue?: mixed): this & T;
   getIn<K: $Keys<T>>(keyPath: [K], notSetValue?: mixed): $ElementType<T, K>;
-  getIn<NSV, K: $Keys<T>, K2: $KeyOf<$ElementType<T, K>>>(keyPath: [K, K2], notSetValue: NSV): $ValOf<$ElementType<T, K>, K2> | NSV;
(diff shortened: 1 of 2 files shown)


acusti:
This is great! We had around 30 new flow errors in `rc.8` related to this. Will this fix be available via the `facebook/immutable-js#npm` dependency string?

leebyron:
Yes, although Travis CI is being slow today. I'm also releasing rc.9 now

PR #1401, opened 2017-10-18
test(ts-definitions): test Immutable.d.ts validity using tsc

TheophileSketo:
It seems the current setup for ts-tests doesn't include Immutable.d.ts in the compilation. This is the only way I could find to test Immutable.d.ts for validity.


leebyron:
Hey, even if it's a little redundant, I think we can't get enough coverage of these! Thanks :)

PR #1402, opened 2017-10-18
Fix Map#concat

philipp-spiess:
#1373 made `Map#concat` an alias for `Map#merge` but defined the alias
before `Map#merge` was defined, resulting in `Map#concat` being
undefined.

This fixes the issue and adds a regression test.


leebyron:
Awesome, thank you for adding a test!

acusti:
Seems like this fix might be worth a new `rc`

cchamberlain:
This appears to still be broken in current `rc`.

PR #1403, opened 2017-10-18
Use prettier for tests

philipp-spiess:
When working on projects that use prettier, I enable the format-on-save
option in my editor of choice which run prettier whenever I save a file.

This always updated the formatting of the tests files when working on
Immutable.js, since those files were not formatted using prettier yet.
Prettier added [typescript support][1] in the mean time, so I thought
it was a good idea to keep the whole code base consistent and to allow
everyone to use format-on-save without getting annoyed.

Since there are not a lot of PRs open, I think it's a good time to merge
this.

[1]: https://github.com/prettier/prettier/releases/tag/1.4.0

### DIFF
--- a/__tests__/ArraySeq.ts
+++ b/__tests__/ArraySeq.ts
@@ -10,7 +10,6 @@
 import { Seq } from '../';
 
 describe('ArraySequence', () => {
-
   it('every is true when predicate is true for all entries', () => {
     expect(Seq([]).every(() => false)).toBe(true);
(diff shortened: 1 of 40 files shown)


myitcv:
We've been using https://github.com/vvakame/typescript-formatter for some time for what it's worth.

leebyron:
Awesome, I've been meaning to do this, so thank you for taking the initiative

PR #1404, opened 2017-10-18
Adds Map.mergeDeep Symbol tests

abacaphiliac:
Demonstrates correct handling of Symbol keys in Map.mergeDeep when the Maps are nested, and initialized with Symbol KV tuples instead of plain JS object literals.

Tests were provided in this discussion:
https://github.com/facebook/immutable-js/pull/1392


abacaphiliac:
@leebyron please review.

leebyron:
Excellent, thanks for adding more tests!

PR #1405, opened 2017-10-18
Document regression test steps.

abacaphiliac:
Documents steps necessary to run the regression test suite.


abacaphiliac:
@leebyron I've been through these steps a few times over the past couple of days, and ran through them just now once more for good measure. It looks right to me. How about you?

leebyron:
Awesome! What do you think about moving this to the `CONTRIBUTING.md` file?

abacaphiliac:
Great idea. Coming up.

abacaphiliac:
@leebyron does that look right?

leebyron:
Looks great! Thanks for your help improving documentation!

PR #1408, opened 2017-10-18
Fix code samples in Immutable.d.ts docs

dchristensen:


leebyron:
Awesome, thanks for your careful read and submitting fixes!

PR #1410, opened 2017-10-18
Add example for OrderedSet subtract.

TheBox193:
Add an example for `subtract()`.


leebyron:
Awesome, thanks!

TheBox193:
@leebyron Doesn't look like this went live. Is there a syntax error I missed?
https://facebook.github.io/immutable-js/docs/#/OrderedSet/subtract

PR #1414, opened 2017-10-19
Improved Flow support for Record subclasses

### LINKED ISSUE #1388: Using flow, Record#set and Record#merge return intersection types (v4)
### What happened

In trying to upgrade draft-js to `4.0.0-rc.7` (and flow 0.57.1), I ran into many errors all around the fact that calling `.set()` or `.merge()` on a Record returns an intersection type that then causes errors for any other typing that refers to the original Record, or in the case of draft, that refers to a class that extends the Record, which is the common pattern in that codebase. For example:

`[ContentState.js](https://github.com/facebook/draft-js/blob/master/src/model/immutable/ContentState.js)`, the Record-based class:

```js
const defaultRecord: {
  entityMap: ?any,
  blockMap: ?BlockMap,
  selectionBefore: ?SelectionState,
  selectionAfter: ?SelectionState,
} = {
  entityMap: null,
  blockMap: null,
  selectionBefore: null,
  selectionAfter: null,
};

const ContentStateRecord = Record(defaultRecord);

class ContentState extends ContentStateRecord {
  ...
}
```

`[updateEntityDataInContentState.js](https://github.com/facebook/draft-js/blob/master/src/model/transaction/updateEntityDataInContentState.js)`:

```js
import type ContentState from 'ContentState';

function updateEntityDataInContentState(
  contentState: ContentState,
  key: string,
  data: {[key: string]: any},
  merge: boolean,
): ContentState {
  const instance = contentState.getEntity(key);
  const entityData = instance.getData();
  const newData = merge ?
    {...entityData, ...data} :
    data;

  const newInstance = instance.set('data', newData);
  const newEntityMap = contentState.getEntityMap().set(key, newInstance);
  return contentState.set('entityMap', newEntityMap);
}
```

Which leads to:

```
Error: src/model/transaction/updateEntityDataInContentState.js:32
 32:   return contentState.set('entityMap', newEntityMap);
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ intersection. This type is incompatible with
 23: ): ContentState {
        ^^^^^^^^^^^^ ContentState
  Member 1:
  1395:   set<K: $Keys<T>>(key: K, value: $ElementType<T, K>): this & T;
                                                               ^^^^ `this` type. See: node_modules/immutable/dist/immutable.js.flow:1395
  Error:
  1395:   set<K: $Keys<T>>(key: K, value: $ElementType<T, K>): this & T;
                                                               ^^^^ RecordInstance. This type is incompatible with the expected return type of. See: node_modules/immutable/dist/immutable.js.flow:1395
   23: ): ContentState {
          ^^^^^^^^^^^^ ContentState
  Member 2:
   60: const ContentStateRecord = Record(defaultRecord);
                                  ^^^^^^^^^^^^^^^^^^^^^ type parameter `Values` of function call. See: src/model/immutable/ContentState.js:60
  Error:
  1387:   get<K: $Keys<T>>(key: K): $ElementType<T, K>;
                                    ^^^^^^^^^^^^^^^^^^ element type. Indexable signature not found in. See: node_modules/immutable/dist/immutable.js.flow:1387
   23: ): ContentState {
          ^^^^^^^^^^^^ ContentState
```

Is there a way to cast the result of every `set` and `merge` operation somehow to ensure that the value can still be typed as the original record type? Or any other recommended way to get around this issue?

Also, @leebyron, is opening issues here a useful way for me to report issues I come across as I work through getting draft.js upgraded (facebook/draft-js#1439)? @flarnie mentioned that you were open to suggestions or questions; if there’s a better venue for that, please let me know. Thanks!

leebyron:
Not ideal, but this adds dramatically better support for subclassing Records when using Flow. The cost of this is that the default values passed to Record are no longer checked - they should be separately checked :/. I've updated the documentation to show this in the example, as well as including an example of subclassing a Record when using Flow.

Fixes #1388

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -2228,12 +2228,12 @@ declare module Immutable {
    * **Flow Typing Records:**
    *
    * Immutable.js exports two Flow types designed to make it easier to use
-   * Records with flow typed code, `RecordOf<T>` and `RecordFactory<TProps>`.
+   * Records with flow typed code, `RecordOf<TProps>` and `RecordFactory<TProps>`.
    *
    * When defining a new kind of Record factory function, use a flow type that
    * describes the values the record contains along with `RecordFactory<TProps>`.
    * To type instances of the Record (which the factory function returns),
-   * use `RecordOf<T>`.
+   * use `RecordOf<TProps>`.
    *
    * Typically, new Record definitions will export both the Record factory
    * function as well as the Record instance type for use in other code.
@@ -2243,7 +2243,8 @@ declare module Immutable {
    *
    * // Use RecordFactory<TProps> for defining new Record factory functions.
    * type Point3DProps = { x: number, y: number, z: number };
-   * const makePoint3D: RecordFactory<Point3DProps> = Record({ x: 0, y: 0, z: 0 });
+   * const defaultValues: Point3DProps = { x: 0, y: 0, z: 0 };
+   * const makePoint3D: RecordFactory<Point3DProps> = Record(defaultValues);
    * export makePoint3D;
    *
    * // Use RecordOf<T> for defining new instances of that Record.
@@ -2251,6 +2252,30 @@ declare module Immutable {
    * const some3DPoint: Point3D = makePoint3D({ x: 10, y: 20, z: 30 });
    * ```
    *
+   * **Flow Typing Record Subclasses:**
+   *
+   * Records can be subclassed as a means to add additional methods to Record
+   * instances. This is generally discouraged in favor of a more functional API,
+   * since Subclasses have some minor overhead. However the ability to create
+   * a rich API on Record types can be quite valuable.
+   *
+   * When using Flow to type Subclasses, do not use `RecordFactory<TProps>`,
+   * instead apply the props type when subclassing:
+   *
+   * ```js
+   * type PersonProps = {name: string, age: number};
+   * const defaultValues: PersonProps = {name: 'Aristotle', age: 2400};
+   * const PersonRecord = Record(defaultValues);
+   * class Person extends PersonRecord<PersonProps> {
+   *   getName(): string {
+   *     return this.get('name')
+   *   }
+   *
+   *   setName(name: string): this {
+   *     return this.set('name', name);
+   *   }
+   * }
+   * ```
    *
    * **Choosing Records vs plain JavaScript objects**
    *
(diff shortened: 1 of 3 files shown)


acusti:
The subclassed example looks great: easy, simple, clear. Just to be sure I understand, the trade-off is that regular Records change from being typed like:

```js
type Point3DProps = { x: number, y: number, z: number };
const makePoint3D: RecordFactory<Point3DProps> = Record({ x: 0, y: 0, z: 0 });
```

to:

```js
type Point3DProps = { x: number, y: number, z: number };
const defaultValues: Point3DProps = { x: 0, y: 0, z: 0 };
const makePoint3D: RecordFactory<Point3DProps> = Record(defaultValues);
```

where the default values are defined separately and typed at the point that it is defined? If so, it seems like a reasonable trade-off to me.

leebyron:
Yep, that's the tradeoff

patriqdesigns:
How can I use this in my project right now?
Tried using `"immutable": "git://github.com/facebook/immutable-js.git#d23746f"` as a dependency but since there is no compiled version in the github repository it simply can't find the module.
Any suggestion to use this with a team of developers, without me telling me what do change, or compile the module myself?

acusti:
Here’s what I use:

```json
    "immutable": "facebook/immutable-js#npm",
```

This gives you the latest build for npm always. It reflects whatever is in the `master` branch, and is the most bleeding edge thing you can use. You can also choose a specific pre-release version, the latest being 4.0.0-rc.9:

```json
    "immutable": "^4.0.0-rc.9",
```

patriqdesigns:
Thanks man! Should have looked into the branches instead of trying to use the specific commit...

alaindresse:
I would love to use this, but am a bit held off because I don't know how close we are to a stable 4.0.0.

We don't use immutablejs yet in the current project, and will introduce it, within a limited scope which will be deployed in the coming 4 weeks. Would you recommend I stick to the latest stable version ?

acusti:
@alaindresse Do you use flow or typescript? If so, the v4 release candidates are a massive improvement over v3.x. However, there are known issues, and I don’t have any idea about the timeline for stable v4 being released, so if you aren’t using flow or typescript, you should probably start with latest stable and only upgrade once 4.0.0 is released.

alaindresse:
Thanks for the quick reply @acusti! I am indeed using Flow. Where can I see the known issues ?

acusti:
You can find them in the open issues that explicitly mention immutable 4. They’re all on the first page of issues, and most of them are about types. But there are also a few about functionality, like #1438 and #1437 (which looks like an undocumented breaking change, but the undocumented part is the rub). Also note that the official docs at http://facebook.github.io/immutable-js/ describe the behavior of the latest release candidate (4.0.0-rc.9 right now), rather than the latest stable version.

alaindresse:
Thanks. I had noticed the point on the official docs, which led me to consider using it for my project.

PR #1416, opened 2017-10-20
test merge and mergeDeep of Map instances with Symbol keys

abacaphiliac:
adds tests to cover merge and mergeDeep of Map instances with Symbols as keys.

### DIFF
--- a/__tests__/merge.ts
+++ b/__tests__/merge.ts
@@ -229,4 +229,62 @@ describe('merge', () => {
       Map([[a, Map([[b, Map([[c, 10], [d, 2], [e, 20], [f, 30], [g, 40]])]])]])
     );
   });
+
+  it('merges Maps with Symbol keys via global merge', () => {
+    const c = Symbol('c');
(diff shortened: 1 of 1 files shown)


abacaphiliac:
@leebyron Am I crazy, or is this actually working? I think this proves the use-case I described in #1392, with no changes to source. Did something else in master change?

leebyron:
Is it possible that Symbol is getting compiled by typescript to a string?

leebyron:
I can't reproduce this test passing locally, something is fishy

abacaphiliac:
I'm not sure why you and I are seeing different things, but I did throw a couple of `console.log` statements in one of these tests, and this is what came out...

```
  console.log __tests__/merge.ts:224
    Map {}

  console.log __tests__/merge.ts:225
    Map {}
```

I'll make sure the tests I'm running check for specific values at specific key depths without lazily checking object equivalency. These false positives (for me) were definitely under the radar!

PR #1459, opened 2017-12-13
Fix typo 'ofters' -> 'offers'

jaydh:


type-definitions/Immutable.d.ts
```diff
@@ -2290,7 +2290,7 @@ declare module Immutable {
    *
    * **Choosing Records vs plain JavaScript objects**
    *
-   * Records ofters a persistently immutable alternative to plain JavaScript
+   * Records offers a persistently immutable alternative to plain JavaScript
```
rgbkrk: This should probably be `offer` since Records is plural here.

rgbkrk: "Records offer a..."

Emobe:
I already made a PR for this in #1459

kozlitinaelja:
Thank you for working on this! I did merge #1436 which does the same thing, so I am going to close this PR

PR #1464, opened 2017-12-21
add "& T" to some places where it's missing in RecordInstance

### LINKED ISSUE #1465: Record.withMutations not compatible with RecordOf type
### Repro code

```js
// @flow
import {Record} from 'immutable'
import type {RecordOf} from 'immutable'

type BrokenFields = {
  foo: ?string,
  bar: ?string,
}

const BrokenRecord = Record(({
  foo: undefined,
  bar: undefined,
}: BrokenFields))

type Broken = RecordOf<BrokenFields>

const b: Broken = BrokenRecord()
b.withMutations((b: Broken) => {

})
```

### Expected

No flow errors

### Actual

Flow complains that the `mutable` instance passed by `withMutations` doesn't have the property `bar`:
```
Error: src/reducer.js:19
     v-------------------------------
 19: b.withMutations((b: Broken) => {
 20: 
 21: })
     -^ call of method `withMutations`. Method cannot be called on any member of intersection type
 19: b.withMutations((b: Broken) => {
     ^ intersection
  Member 1:
  1379: type RecordOf<Values: Object> = RecordInstance<Values> & Values;
                                        ^^^^^^^^^^^^^^^^^^^^^^ RecordInstance. See: node_modules/immutable/dist/immutable.js.flow:1379
  Error:
                       v---------------
   19: b.withMutations((b: Broken) => {
   20: 
   21: })
       ^ function. This type is incompatible with the expected param type of
  1479:   withMutations(mutator: (mutable: this) => mixed): this & T;
                                 ^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: node_modules/immutable/dist/immutable.js.flow:1479
    This parameter is incompatible:
      1479:   withMutations(mutator: (mutable: this) => mixed): this & T;
                                               ^^^^ RecordInstance. This type is incompatible with. See: node_modules/immutable/dist/immutable.js.flow:1479
       16: type Broken = RecordOf<BrokenFields>
                                  ^^^^^^^^^^^^ object type
        Property `bar` is incompatible:
           16: type Broken = RecordOf<BrokenFields>
                                      ^^^^^^^^^^^^ property `bar`. Property not found in
          1479:   withMutations(mutator: (mutable: this) => mixed): this & T;
                                                   ^^^^ RecordInstance. See: node_modules/immutable/dist/immutable.js.flow:1479
  Member 2:
       v-------------------------------
   19: b.withMutations((b: Broken) => {
   20: 
   21: })
       -^ Values
  Error:
   19: b.withMutations((b: Broken) => {
         ^^^^^^^^^^^^^ property `withMutations`. Property not found in
   16: type Broken = RecordOf<BrokenFields>
                              ^^^^^^^^^^^^ object type
```

### Workaround

Use the type obtained from `Record(({...}: BrokenFields))` (since `RecordInstance` is not exported) or use `asMutable()`/`asImmutable()`.

jedwards1211:
Not sure why it was missing from `updateIn<U>` or `withMutations` arguments.

Fix #1465

### DIFF
--- a/type-definitions/immutable.js.flow
+++ b/type-definitions/immutable.js.flow
@@ -1454,8 +1454,8 @@ declare class RecordInstance<T: Object> {
   removeIn<K: $Keys<T>, K2: $KeyOf<$ValOf<T, K>>, K3: $KeyOf<$ValOf<$ValOf<T, K>, K2>>, K4: $KeyOf<$ValOf<$ValOf<$ValOf<T, K>, K2>, K3>>>(keyPath: [K, K2, K3, K4]): this & T;
   removeIn<K: $Keys<T>, K2: $KeyOf<$ValOf<T, K>>, K3: $KeyOf<$ValOf<$ValOf<T, K>, K2>>, K4: $KeyOf<$ValOf<$ValOf<$ValOf<T, K>, K2>, K3>>, K5: $KeyOf<$ValOf<$ValOf<$ValOf<$ValOf<T, K>, K2>, K3>, K4>>>(keyPath: [K, K2, K3, K4, K5]): this & T;
 
-  updateIn<U>(keyPath: [], notSetValue: mixed, updater: (value: this) => U): U;
-  updateIn<U>(keyPath: [], updater: (value: this) => U): U;
+  updateIn<U>(keyPath: [], notSetValue: mixed, updater: (value: this & T) => U): U;
+  updateIn<U>(keyPath: [], updater: (value: this & T) => U): U;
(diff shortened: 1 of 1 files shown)


donkeybonks:
@leebyron it would be really good to have this, it's causing a bit of an issue for us. We have to put // $FlowFixMe above all our updateIn's which has caused some very real bugs for us.

leebyron:
Makes sense, thanks for the fix

donkeybonks:
Thank you so much for merging this!

PR #1469, opened 2017-12-31
Improve Typescript support for Record

### LINKED ISSUE #1463: Record.get() with optional NSV (TypeScript)
I'm using `v4.0.0-rc.9` with TypeScript.
I wonder why there is no [get](https://github.com/facebook/immutable-js/blob/v4.0.0-rc.9/type-definitions/Immutable.d.ts#L2391) method in the Record interface without the NSV value.
It results in a lot of `.get('..', null)` calls. Since I know that there is already a value, given that a record factory already enforces that a value is always present for all keys, it seems unnecessary.

Zzortell:
* Make NSV optional in Record.get method

Fixes #1463

* Update documentation

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -2429,8 +2429,9 @@ declare module Immutable {
      *
      * If the requested key is not defined by this Record type, then
      * notSetValue will be returned if provided. Note that this scenario would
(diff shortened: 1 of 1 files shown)


DogPawHat:
This would be a very useful fix for our typescript projects as well.

PR #1486, opened 2018-02-26
Documentation: correct the default reviver

shimaore:
`Seq.Keyed` does not have `.Map`, and [the default reviver](https://github.com/facebook/immutable-js/blob/master/src/fromJS.js#L49) indeed uses `.toMap`.


asazernik:
Duplicate of #1424

kozlitinaelja:
Thank you for fixing this! I did merge  #1424 which does the same thing. Going to close this PR

PR #1489, opened 2018-03-06
fix: fix typo in docs. ES3 -> ES5

Chun-Lin:
## Fix the typo in Immutable.d.ts

* ES3 should be ES5


amiiit:
@Chun-Lin there are some very old browsers that implement only ES3, so in this case the documentation might be right.

PR #1492, opened 2018-03-15
Add API documentation to OrderedMap.set()

amiiit:
Add specific documentation to `OrderedMap.set` stating that it returns a new OrderedMap and maintains the existing order.

This attempts to avoid confusion between users as the current documentation for this method inherits from `Immutable.Map` which can be understood like an `OrderedMap.set() => Map`, which is wrong, and might lead to uncertainty about the order of the entries. This Pull Request adds the missing bit to the documentation.

[Link to relevant documentation part of OrderedMap.set()](https://facebook.github.io/immutable-js/docs/#/OrderedMap/set)

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -1430,6 +1430,27 @@ declare module Immutable {
     readonly size: number;
 
     /**
+     * Returns a new OrderedMap also containing the new key, value pair. If an equivalent
+     * key already exists in this OrderedMap, it will be replaced while maintaining the
+     * existing order.
+     *
+     * <!-- runkit:activate -->
+     * ```js
+     * const { OrderedMapMap } = require('immutable')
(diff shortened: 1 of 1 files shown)


type-definitions/Immutable.d.ts
```diff
@@ -1429,6 +1429,27 @@ declare module Immutable {
...
 
+    /**
+     * Returns a new OrderedMap also containing the new key, value pair. If an equivalent
+     * key already exists in this OrderedMap, it will be replaced while maintaining the
+     * existing order.
+     *
+     * <!-- runkit:activate -->
+     * ```js
+     * const { OrderedMapMap } = require('immutable')
+     * const originalMap = Immutable.OrderedMap({a:1, b:1, c:1})
```
leebyron: In the future - ensure lines like these are correct - this both imports a value that doesn't exist, and then proceeds to use a pattern we don't support in docs. I just added a commit to fix these issues.

Issue #1431, opened 2017-10-30
Code typo in README.md


Ltrlg:
The following code snippet, copied from the readme, has one syntax errors and one logical error:
```js
const { Map } = require('immutable')
const map = Map({ a: 1, b: 2, c: 3 }
const lazySeq = Seq(map)
```

From what I understand, it should be written as follows:
```js
const { Map, Seq } = require('immutable')
const map = Map({ a: 1, b: 2, c: 3 })
const lazySeq = Seq(map)
```
(adding an imported identifier and a parenthesis)

ajomadlabs:
Can I take up this ?

PR #1500, opened 2018-03-28
Typo fix for code sample in docs

acusti:
Existing code sample was missing an import and had a syntax error.


acusti:
Fixes #1431

PR #1531, opened 2018-05-17
Update to latest yarn

leebyron:
This is currently a WIP that doesn't pass tests.

There is some weird behavior happening that may be flow bugs, but I wanted to get it up for early discussion.

### DIFF
--- a/type-definitions/tests/immutable-flow.js
+++ b/type-definitions/tests/immutable-flow.js
@@ -180,7 +180,6 @@ numberList = List.of('a').merge(List.of(1))
 // Functional API
 
 numberList = merge(List([1]), List([2]))
-// $ExpectError - Functional API currently requires arguments to have the same value types.
 numberOrStringList = merge(List(['a']), List([1]))
 // $ExpectError
 numberList = merge(List(['a']), List([1]))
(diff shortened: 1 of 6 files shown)


type-definitions/tests/immutable-flow.js
```diff
@@ -929,7 +926,7 @@ personRecordInstance.set('invalid', 25)
 personRecordInstance.set('name', '25')
 personRecordInstance.set('age', 33)
 
-// $ExpectError
+// TODO: This should be ExpectError, perhaps $KeyOf<> is broken?
 set(personRecordInstance, 'invalid', 25)
```
leebyron: Still investigating this one :(

type-definitions/tests/merge.js
```diff
@@ -29,6 +29,27 @@ import {
   updateIn,
 } from '../../';
 
+// merge: Objects as Maps
```
leebyron: I did nothing but move this block higher in the file, otherwise it complains that `z` isn't found in `XYPoint` despite `objMap` and `XYPoint` never be used together in the file. It seems like flow somehow retains memory of using `XYPoint` with the `merge()` function and incorrectly applies it to a later use on different input arguments.

@calebmer any idea how to reduce or report this?

PR #1533, opened 2018-05-18
isPalinObj not detect a object like {constructor: 'Adam'} as plain ob…

shonny-ua:
…ject

In according to https://stackoverflow.com/questions/5876332/how-can-i-differentiate-between-an-object-literal-other-javascript-objects
I propose next fix.

### DIFF
--- a/src/utils/isPlainObj.js
+++ b/src/utils/isPlainObj.js
@@ -1,16 +1,16 @@
 export default function isPlainObj(value) {
   // Basic check for Type object that's not null
-  if (typeof value == 'object' && value !== null) {
+  if (typeof value === 'object' && value !== null) {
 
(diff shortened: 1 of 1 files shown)


leebyron:
Unfortunately this is failing tests and will severely regress performance (try out the benchmarking scripts), so I don't think we can consider merging this without proof that it's at least net-neutral for performance. It has also been attempted in a few ways before #1056 #452

shonny-ua:
After last commit https://github.com/shonny-ua/immutable-js/commit/fcb08740d668cf24f9e44930254fe304ffe0ac7a
all tests was passed.
If I compare "severely regress performance" and correct work of module, I chose last.
It's not about y our module....

PR #1537, opened 2018-05-23
Create nested records on Record construction

camullen:
If a default value for a key(`innerKey`) of a record type (`OuterRecordType`) is another record (`innerRecord`), upon construction of `OuterRecordType`, if the `innerKey` is set to a plain Javascript object (`innerObject`), the constructed record (`outerRecord`) will have a value for `innerKey` that is the result of deeply merging the `innerObject` into the `innerRecord`.

```javascript
import Immutable from 'immmutable';

const InnerRecordType = Immutable.Record({
  a: 10,
  b: 20,
  c: 30,
});

const OuterRecordType = Immutable.Record({
  one: 1,
  two: 2,
  innerKey: InnerRecordType({ a: 0 }),
});

const outerRecord = OuterRecordType({ innerKey: { c: 40 } });

outerRecord.innerKey instanceof InnerRecordType // true
outerRecord.innerKey.a // 0
outerRecord.innerKey.b // 20
outerRecord.innerKey.c // 40
```

Records can also be deeply nested

### DIFF
--- a/__tests__/Record.ts
+++ b/__tests__/Record.ts
@@ -146,6 +146,52 @@ describe('Record', () => {
     expect(t4).toBe(t2);
   });
 
+  it('creates nested records when when default value is a record', () => {
+    const MyDeepInnerType = Record({
+      oneFish: 'twoFish',
+      redFish: 'blueFish',
+      blackFish: 'blueFish',
+      oldFish: 'newFish',
+    });
+    const MyInnerType = Record({
+      a: 1,
+      b: 2,
+      c: 3,
+      deep: MyDeepInnerType({ redFish: 'snapper' }),
+    });
+    const MyOuterType = Record({
+      one: 1,
+      two: 2,
+      inner: MyInnerType({
+        a: 99,
+        deep: {
+          blackFish: 'Brynden Tully',
+        } as any,
+      }),
+    });
+    const t = MyOuterType({
+      two: 4,
+      inner: { b: 3, deep: { oldFish: 'stinky' } } as any,
+    });
+    expect(t.inner instanceof MyInnerType).toBe(true);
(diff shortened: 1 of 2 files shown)


camullen:
@facebook-github-bot Just signed CLA

camullen:
Rationale for this change is that if you’d like to use a Record type to hold a response from an API that has nested objects, you don’t have to manually traverse the objects to construct the record.

marcelmokos:
@camullen I like your change, it will solve the same issue in our project to create nested records from API calls. Thanks

camullen:
@marcelmokos Thanks! I appreciate it. Any suggestions for improvements?

marcelmokos:
@camullen I would write a test for records with getters as we are using getters as selectors in our project. But the test would just make clear to me if mergeDeep is working as I am expecting it should. 

I expect mergeDeep should merge underlying maps but in records and leave the getters functional. I also expect that flow will digest it correctly since right now we are creating a type for the constructor of a record.

TheHaff:
nice. absolutely want this

leebyron:
Unfortunately this slows down Record construction with the additional merge and makes it more challenging to flow/typescript type.

leebyron:
I'm also concerned about cases where a record may have a record instance as a default value, but supports multiple record types as values for that key. Or cases where a default value is null, but Record instances are expected. Or what about cases which expect Lists of Records or Maps of Records?

I'm worried that this behavior will be too unpredictable. Not apply in cases where you expect it might, and apply in cases you do not want it to.

If there's a really compelling reason to have this behavior, I'd like to discuss what it might look like to move it out of the Record constructor and into some other utility

leebyron:
Closing this PR since I don't think we can merge this as written, however definitely give some thought to my comment above as to what a utility function might look like which has the behavior we're looking for

PR #1539, opened 2018-05-29
type-definitions: fix has() for TypeScript 2.9

bpowers:
I'm not sure this is the _right_ fix, but it is _a_ fix.  TypeScript
2.9 (in RC, due out soon AIUI) has a breaking change around `keyof`:

> TypeScript 2.9 changes the behavior of `keyof` to factor in both unique symbols as well as numeric literal types. ([announcement](https://blogs.msdn.microsoft.com/typescript/2018/05/16/announcing-typescript-2-9-rc/))

This PR fixes things, but I'm not sure of compatability with <= 2.8?
Happy to revise or look into this further.

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -2440,7 +2440,7 @@ declare module Immutable {
 
     // Reading values
 
-    has(key: string): key is keyof TProps;
+    has(key: string | number | Symbol): key is keyof TProps;
 
     /**
      * Returns the value associated with the provided key, which may be the


Huvik:
Any news? :)

type-definitions/Immutable.d.ts
```diff
@@ -2440,7 +2440,7 @@ declare module Immutable {
 
     // Reading values
 
-    has(key: string): key is keyof TProps;
+    has(key: string | number | Symbol): key is keyof TProps;
```
IllusionMH: Shouldn't it be `symbol` (lowercase) as type, not as constructor (interface)?

szagi3891:
@leebyron - Can you this merged? There is now a problem with using the immutable library with Typescript.

zcr1:
Ran into this issue today

SqueezeToyAliens:
@zcr1 Like many others. We'll really appreciate a merge here :)

neilchaudhuri:
It looks like I've run into this issue as well as I posted on [Stack Overflow](https://stackoverflow.com/questions/51750674/build-error-with-typescript-3-and-immutable-js-4).

What's the holdup on merging this PR? Is there we can do in the community to help?

type-definitions/Immutable.d.ts
neilchaudhuri: So is this what is holding up the merge?

neilchaudhuri:
For my issue setting `"skipLibCheck": true` in `tsconfig.json` solves this fwiw.

Methuselah96:
Made a new PR with the proper fix: https://github.com/facebook/immutable-js/pull/1571.

Methuselah96:
Made a PR for a better fix: https://github.com/facebook/immutable-js/pull/1571.

derek126:
When will this be released?

alangpierce:
I was able to work around this using `"skipLibCheck": true`, but it's really inconvenient to keep that config because it means that _all_ `.d.ts` files have their errors completely ignored, not just third-party definition files. It would be great to get a new release that includes a fix for this.

type-definitions/Immutable.d.ts
alangpierce: I think maybe `key is keyof TProps & string;` is best here. `Extract` was introduced in TS 2.8, so using it would break anyone using 2.7 or earlier.

Issue #1503, opened 2018-03-30
Non-obvious behavior when omitting the merger for `mergeWith`


bwaite-co:
### What happened

Using `mergeWith` without a function fails to merge the objects or throw an error. This seems to be non-obvious behavior and out of step with the JavaScript builtin functions throwing a `TypeError` when a function is missing or is missing a required argument.

e.g.:
```javascript
window.addEventListener('keyup')
// TypeError: Failed to execute 'addEventListener' on 'EventTarget': 2 arguments required, but only 1 present.
// or
function myMapper(mapper) {
  return [1,2,3].map(mapper)
}
myMapper({x: 1, y: 2}) // not actually a function by mistake
// Uncaught TypeError: #<Object> is not a function
//  at Array.map (<anonymous>)
//  at <anonymous>:1:11
```

### How to reproduce

```javascript
// 4.0.0-rc9, Chrome 60
var { Map } = require('immutable')
Map({a: 1, b: 2}).merge(Map({c: 3}))
// Map { "a": 1, "b": 2, "c": 3 }
Map({a: 1, b: 2}).mergeWith(Map({c: 3}))
// Map { "a": 1, "b": 2 }
```

Brantron:
[Fixed](https://github.com/facebook/immutable-js/pull/1543) - hopefully it will land in master soon

PR #1543, opened 2018-06-15
throw error in mergeWith method if no merger is passed (#1503)

Brantron:
This ensures users get an informative error when calling `mergeWith` without a merge function

Details can be found at https://github.com/facebook/immutable-js/issues/1503.

### DIFF
--- a/src/functional/merge.js
+++ b/src/functional/merge.js
@@ -38,7 +38,8 @@ export function mergeWithSources(collection, sources, merger) {
     );
   }
   if (isImmutable(collection)) {
-    return collection.mergeWith
+    const shouldMergeWith = collection.mergeWith && merger instanceof Function;
+    return shouldMergeWith
       ? collection.mergeWith(merger, ...sources)
       : collection.concat(...sources);
   }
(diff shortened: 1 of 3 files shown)


Brantron:
@facebook-github-bot Just signed CLA

src/functional/merge.js
```diff
@@ -38,7 +38,8 @@ export function mergeWithSources(collection, sources, merger) {
     );
   }
   if (isImmutable(collection)) {
-    return collection.mergeWith
+    const shouldMergeWith = collection.mergeWith && merger instanceof Function;
```
asazernik: This logic translates all calls to `Immutable.merge(collection, sources)` into calls to `collection.concat(...sources)`. Should instead check for `collection.merge` if `merger` isn't a function. Something like:

```javascript
if (collection.mergeWith && merger instanceOf Function) {
  return collection.mergeWith(merger, ...sources);
} else if (collection.merge && !(merger instanceOf Function) { // REALLY not sure about this condition
  return collection.merge(...sources);
} else {
  return collection.concat(...sources);
}

Brantron:
@asazernik thanks for taking a peak, pushed a commit.

leebyron:
Squash merged in 2ea44e0ed31206de90d9ae6fc8082d63a052e8a3, thanks!

Issue #1542, opened 2018-06-08
Record.withMutations adds extraneous keys


alessandrotilli:
From the docs:

> A record is similar to a JS object, but enforces a specific set of allowed string keys

Using `withMutations` to modify a record can break the previous statement, allowing you to add extraneous keys. 

### How to reproduce

```javascript
const { Record } = require('immutable');

const Rec = new Record({ a: 1 });
const rec = new Rec();

const r1 = rec.set('wtf', '?');
console.log(r1.wtf); // undefined (OK)

const r2 = rec.withMutations((m) => {
  m.a = 2;
  m.wtf = '?';
})

console.log(r2.a); // 2
console.log(r2.wtf); // ? (unexpected)
console.log(r2.has('wtf')) // false;
```
tested on immutable@4.0.0-rc.9.

The same behaviour occurs with immutable@3.8.2, the only difference is that `rec.set('wtf', '?')` (correctly) throws an error.

Brantron:
[Fixed](https://github.com/facebook/immutable-js/pull/1544) - hopefully it will land in master soon

jdeniau:
Fixed by https://github.com/jdeniau/immutable-js/commit/9fa9e6bdfda9dace10f7334f4345550a9290d21c

PR #1544, opened 2018-06-16
prevent withMutations from setting values on non defined record keys

Brantron:
This closes https://github.com/facebook/immutable-js/issues/1542

I created a separate method since this is shared and I didn't think an `isRecord` check would be the most performant solution.

### DIFF
--- a/src/methods/withMutations.js
+++ b/src/methods/withMutations.js
@@ -10,3 +10,10 @@ export function withMutations(fn) {
   fn(mutable);
   return mutable.wasAltered() ? mutable.__ensureOwner(this.__ownerID) : this;
 }
+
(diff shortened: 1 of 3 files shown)


src/methods/withMutations.js
```diff
@@ -10,3 +10,10 @@ export function withMutations(fn) {
   fn(mutable);
   return mutable.wasAltered() ? mutable.__ensureOwner(this.__ownerID) : this;
 }
+
+export function withRecordMutations(fn) {
+  const mutable = this.asMutable();
+  Object.preventExtensions(mutable);
```
rgbkrk: +1

Issue #1547, opened 2018-06-19
undefined and false values have the same hashCode


jackwootton:
### What happened

`{value: false}` and `{value: undefined}` have the same hash code.

### How to reproduce

```
const a = fromJS({value: false}).hashCode();
const b = fromJS({value: undefined}).hashCode();

console.log(a, b);
```
Will log:

> 10995925 10995925

Brantron:
There's a fix here https://github.com/facebook/immutable-js/pull/1549

leebyron:
I commented on the PR - but could you back up a bit and explain how this affects you?

Hash collisions will happen and the immutable collections will do the right thing.

I want to ensure we're balancing complexity and performance correctly here

PR #1549, opened 2018-06-19
differentiate falsy values when hashing (#1547)

Brantron:
this PR fixes https://github.com/facebook/immutable-js/issues/1547


leebyron:
Unfortunately this slows down the hash function for these values. Ideally we just return fixed hashes for these fixed values.

To back up a bit though, can you share more about how this change is affecting you?

PR #1556, opened 2018-07-09
Added optional value in first and last

yasserhussain1110:


rgbkrk:
Looks solid to me!

__tests__/Seq.ts
```diff
@@ -10,6 +10,14 @@
 import { isCollection, isIndexed, Seq } from '../';
 
 describe('Seq', () => {
+  it('has a default first value', () => {
+    expect(Seq().first({})).toEqual({});
+  });
+
+  it('has a default first value', () => {
+    expect(Seq().last({})).toEqual({});
+  });
```
rgbkrk: Should probably also have a test that ensures that calling `first` and `last` on an empty `Seq` with no default set will return `undefined`. Only to make sure the behavior is consistent with the past.

yasserhussain1110:
@rgbkrk Made the required changes.

asazernik:
Add this optional argument to the type definitions (`type-definitions/Immutable.d.ts` and `type-definitions/immutable.js.flow`) and describe it in the docs (the javadoc-ish comments in `type-definitions/Immutable.d.ts`).

yasserhussain1110:
Added type definitions. Please let me know if further changes are required.

rgbkrk:
@flarnie @leebyron -- this PR has been put through the paces -- would you be willing to check it out for incorporation into Immutable 4.0?

hrldcpr:
Note that this PR leads to new type errors since `NSV` defaults to `{}`—see this comment for example https://github.com/facebook/immutable-js/commit/37ae5beef11664371ecb231b422a25ffa11126eb#r31116027

PR #1557, opened 2018-07-11
Update isPlainObj to workaround Safari bug

cvazac:
While debugging an issue in the wild with our library [boomerang](https://github.com/akamai/boomerang), we uncovered a [bug in webkit](https://bugs.webkit.org/show_bug.cgi?id=187411) that causes `xhr.response` to no longer work with Immutable.js.

You can see the issue reproduce here: 
https://cvazac.netlify.com/xhr-response-constructor-immutable

This small code change works around version 605.1.15 of Safari.


cvazac:
Do you know who the right person to review this is @leebyron?

leebyron:
This is great, thanks for this fix!

PR #1577, opened 2018-08-18
Fix TypeScript issue with `Record` `has` method.

cchamberlain:
The old version was causing type errors with my TypeScript configuration due to key being `string` / `number` / `Symbol`.

**The issue:**
image

- This is closer to the `get` function definition below it:
```ts
has<K extends keyof TProps>(key: K): boolean;
```

**After the fix:**
image

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -2440,7 +2440,7 @@ declare module Immutable {
 
     // Reading values
 
-    has(key: string): key is keyof TProps;
+    has<K extends keyof TProps>(key: K): boolean;
 
     /**
      * Returns the value associated with the provided key, which may be the


errendir:
Hello

I believe the suggested change could break some existing code. Consider the following code (compiled with `--keyofStringsOnly`):
```
interface Record_<TProps extends object> {
  has(key: string): key is keyof TProps
}

type Shape = { a: number, b: boolean, [Symbol.iterator]: any }

function defaultKey(a: Record_<Shape>, key: string) {
  let z: "a" | "b"
  if (a.has(key)) {
    z = key
  } else {
    z = "a"
  }
  return z
}
```

The `z = key` assignment is allowed, since inside of the `if` block, the type of `key` is constrained to be extending `keyof TProps`. Your change removes this constraint.

I think a better solution is to change:
```js
has(key: string): key is keyof TProps
``` 
to
```js 
has(key: string): key is ((keyof TProps) & string)
```

IllusionMH:
Looks like "duplicate" on PR #1539 and as mentioned above - breaks current code and intention of the method.

Suggested fix is https://github.com/facebook/immutable-js/pull/1539/files#r195696659

PR #1578, opened 2018-08-19
Add `RecordOf<TProps>` type alias for TypeScript

cchamberlain:
The `Record.Factory` API is great but the current definitions lack a simple way to reference instances of a record (created from a record factory). This adds a `RecordOf<TProps>` which can be used to create interfaces containing records:

```ts
import { Record, RecordOf } from "immutable";

interface Foo {
    flag: boolean;
    message: string;
}

const FooRecord = Record<Foo>({
    flag: false,
    message: "DEFAULT"
});

interface Bar {
    foo: RecordOf<Foo>;
}

const bar: Bar = {
    foo: new FooRecord({ flag: true })
}
```

Currently I've been creating this type in every library so that I can refer to instances and I think Flow has support for something similar.

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -2565,7 +2565,7 @@ declare module Immutable {
    * 
    * This is equivalent to an instance of a record created by a Record Factory.
    */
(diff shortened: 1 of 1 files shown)


leebyron:
Nice

Issue #1453, opened 2017-12-04
Performance problem with toJS function in Immutable 4.0.0-rc.8


rafalsep:
### What happened
Converting Immutable Map to pure JS object via toJS function in release 4.0.0-rc.9 is very slow as compared to previous release (3.8.2) and 4.0.0-rc.7

### How to reproduce
Please take a look at project created below to ilustrate the issue. Running toJS on large Map object is more than 2x slower with rc.8 (as compared to rc.7). 
https://github.com/hopeisnear/immutable-v4-toJS-speed-test

release 4.0.0-rc.7: TOTAL toJS: 446.493ms
release 4.0.0-rc.8: TOTAL toJS: 947.584ms

Tested on:
node: 8.9.1
npm: 5.3.0

Execution time measured using console.time.

lukaswelinder:
After doing some investigating, it looks like the changes [here](https://github.com/facebook/immutable-js/pull/1369) are to blame.

[This change](https://github.com/facebook/immutable-js/pull/1369/files#diff-94ed586fbea7c838707c5beb81e6e57dR12) in particular is what caught my eye; the `isDataStructure` function results in a lot of conditional function calls if the data is a primitive value. 

Checking if the value is an object before calling `isDataStructure` helps quite a bit. The following results in comparable performance to 4.0.0-rc.7:

``` js
function toJS(value) {
  return (typeof value === 'object' && isDataStructure(value))
    ? Seq(value)
      .map(toJS)
      .toJSON()
    : value;
}
// > TOTAL toJS: 457.498ms
```

Ultimately, I think it would be better to just do the check inside of the `isDataStructure` function. This approach is cleaner and yields the same performance.

``` js
function isDataStructure(value) {
  return typeof value === 'object' && (isImmutable(value) || Array.isArray(value) || isPlainObj(value));
}
```

With that change and a little more experimenting, I managed to implement `toJS` that is ~33% faster than 4.0.0-rc.7 and ~69% faster than 4.0.0-rc.8 & 4.0.0-rc.9.

```js
function toJS(value) {
  if(isDataStructure(value)) {
    value = Seq(value);
    const result = isIndexed(value) ? [] : {};
    value.forEach((v, k) => {
      result[k] = toJS(v);
    });
    return result;
  }
  return value;
}
// > TOTAL toJS: 295.861ms
```

Not as pretty, but probably worth the performance gains.

PR #1581, opened 2018-08-25
Improve performance of toJS (#1453)

lukaswelinder:
This PR addresses #1453, significantly improving performance of `toJS` method; more details can be found in my comment on that issue.

It might be worth adding a performance regression test for the `toJS` method.

### DIFF
--- a/src/toJS.js
+++ b/src/toJS.js
@@ -12,6 +12,7 @@ import isDataStructure from './utils/isDataStructure';
 export function toJS(value) {
   if (isDataStructure(value)) {
     value = Seq(value);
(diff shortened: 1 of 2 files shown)


leebyron:
Nice! I was able to speed things up a bit further by avoiding the Seq cast for collections and using the internal __iterate method instead of forEach - squash merge in 0082ff78295fb818eca80e898edbea5af758f9db

lukaswelinder:
@leebyron Awesome! Nice work; over 60% faster than it was even before the issue cropped up in `4.0.0-rc.8` (using the same test linked in #1453).

PR #1589, opened 2018-09-08
[BREAKING] Remove IteratorSequence. Do not attempt to detect iterators.

### LINKED ISSUE #1456: merge breaks on objects with a 'next' key
### What happened

There is a fatal error when running the code shown below.
In the simple example that causes the error I expect the following result instead:

```javascript
fromJS({next: () => 1}).merge({next: () => 2}).toJS() // {next: () => 2}
```

In the browser the error was not fatal (for the browser) but resulted in some different behavior as well.

(It seems to be related to generators expecting a `next` prop and might thus be a transpilation concern.)

### How to reproduce

This code doesn't work when running with node 8.9.2 on macOS Sierra 10.12.6 

```javascript
const { fromJS } = require('immutable@4.0.0-rc.9')
const original = {next: () => 1}
console.log(fromJS(original).mergeDeep({next: () => 2}))
// merge instead of mergeDeep also results in a crash.
```

It runs just fine when `next` is replaced by e.g. `a`.

Here is what node printed:

```
<--- Last few GCs --->

[5789:0x104001600]    26953 ms: Mark-sweep 578.7 (586.7) -> 578.7 (586.7) MB, 249.2 / 0.0 ms  allocation failure GC in old space requested
[5789:0x104001600]    27194 ms: Mark-sweep 578.7 (586.7) -> 578.7 (583.2) MB, 240.9 / 0.0 ms  last resort GC in old space requested
[5789:0x104001600]    27442 ms: Mark-sweep 578.7 (583.2) -> 578.7 (583.2) MB, 247.3 / 0.0 ms  last resort GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x22e4d6825ee1 <JSObject>
    1: /* anonymous */ [.../node_modules/immutable/dist/immutable.js:~489] [pc=0x284a5c20a083](this=0x22e4735d4849 <IteratorSeq map = 0x22e424ad4111>,fn=0x22e4735d4801 <JSFunction (sfi = 0x22e4fe830141)>,reverse=0x22e473582311 <undefined>)
    3: seqIterate(aka seqIterate) [.../node_modules/immutable/dist...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [/usr/local/bin/node]
 2: node::FatalException(v8::Isolate*, v8::Local<v8::Value>, v8::Local<v8::Message>) [/usr/local/bin/node]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [/usr/local/bin/node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [/usr/local/bin/node]
 5: v8::internal::Factory::NewUninitializedFixedArray(int) [/usr/local/bin/node]
 6: v8::internal::(anonymous namespace)::ElementsAccessorBase<v8::internal::(anonymous namespace)::FastPackedObjectElementsAccessor, v8::internal::(anonymous namespace)::ElementsKindTraits<(v8::internal::ElementsKind)2> >::GrowCapacity(v8::internal::Handle<v8::internal::JSObject>, unsigned int) [/usr/local/bin/node]
 7: v8::internal::Runtime_GrowArrayElements(int, v8::internal::Object**, v8::internal::Isolate*) [/usr/local/bin/node]
 8: 0x284a5c08463d
Abort trap: 6
```

conartist6:
Fixes #1456


Brantron:
Out of curiosity, what is the speed/perf impact of this change?

conartist6:
There could be a [very] small win since sequences are sometimes constructed internally as part of other implementations, but my motivation for this change is correctness not perf.

leebyron:
This is the right thing to do. Thank you for your investigation and work.

All built in iterables define `[Symbol.iterator]() { return this }` and most libraries do as well, so in practice this should be a very minor breaking change.

Issue #1604, opened 2018-09-20
set.has is incorrect if constructed via set.map in rc.10


jessezhang91:
### What happened
Mapping over a set produces a set that does not correctly evaluate `has` checks. Confirmed it is only happening in rc.10, not rc.9.

### How to reproduce

```
const { Set } = require("immutable");

const set1 = Set([1, 2, 3]);
console.log(set1.has(4)); // rc.10: false, rc.9: false
console.log(set1._map.keySeq().toArray()); // rc.10: [1, 2, 3], rc.9: [1, 2, 3]
console.log(set1._map.valueSeq().toArray()); // rc.10: [1, 2, 3], rc.9: [ 1, 2, 3 ]

const set2 = set1.map((a) => a + 1);
console.log(set2.has(4)); // rc.10: false, rc.9: true
console.log(set2._map.keySeq().toArray()); // rc.10: [1, 2, 3], rc.9: [2, 3, 4]
console.log(set2._map.valueSeq().toArray()); // rc.10: [2, 3, 4], rc.9: [ 2, 3, 4 ]
```

It looks like the mapper in 
https://github.com/facebook/immutable-js/blob/master/src/Set.js#L85
should map both key and value.

The has in
https://github.com/facebook/immutable-js/blob/master/src/Set.js#L65
uses the key of the underlying map to check which was not mapped.

PR #1606, opened 2018-09-20
fix: set.map produces valid underlying map

jessezhang91:
Fixes https://github.com/facebook/immutable-js/issues/1604

As explained in issue, `set.map` would produce a set whose underlying `_map` had mismatched key-value pairs (expected to always be equal for sets). The original code only mapped the value and kept the original set's keys which caused `has` checks (which uses the `_map` keys) to be incorrect.

- Updated the code to use `mapEntries` on the underlying `_map`
- `updateSet` uses `equals` instead of `===` since `mapEntries` produces a different map reference
- Added test to check for this case

### DIFF
--- a/src/Set.js
+++ b/src/Set.js
@@ -83,13 +83,17 @@ export class Set extends SetCollection {
 
   map(mapper, context) {
     return this.withMutations(set => {
+      const removes = [];
+      const adds = [];
       this.forEach(value => {
         const mapped = mapper.call(context, value, value, set);
         if (mapped !== value) {
-          set.remove(value);
-          set.add(mapped);
+          removes.push(value);
+          adds.push(mapped);
         }
       });
+      removes.forEach(value => set.remove(value));
+      adds.forEach(value => set.add(value));
     });
   }
 
--- a/src/Set.js
+++ b/src/Set.js
@@ -82,13 +82,15 @@ export class Set extends SetCollection {
   // @pragma Composition
 
   map(mapper, context) {
-    return updateSet(
-      this,
-      this._map.mapEntries(([, v]) => {
-        const mapped = mapper(v, v, this);
-        return [mapped, mapped];
-      }, context)
-    );
+    return this.withMutations(set => {
+      this.forEach(value => {
+        const mapped = mapper.call(context, value, value, set);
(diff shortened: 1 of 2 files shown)


src/Set.js
```diff
@@ -200,7 +206,7 @@ function updateSet(set, newMap) {
     set._map = newMap;
     return set;
   }
-  return newMap === set._map
+  return newMap.equals(set._map)
```
leebyron: For performance reasons, it's important that this is a cheap reference equality check. Specifically, we expect `newMap` to be a new reference if it has been altered

jessezhang91: Gotcha. I'll see what I can do to maintain the same reference in the map. I did notice that `map` and `mapEntries` have different code paths. Is there a recommendation on how I should approach this?

jessezhang91: I tested two approaches:
- doing `mapEntries` and keeping track if there was any alterations and only `updateSet` if changed
- using `withMutations` and `remove`/`add` when there is a change

The latter was faster in some micro-benchmarks so I've opted for that solution.

src/Set.js
```diff
@@ -82,7 +82,15 @@ export class Set extends SetCollection {
   // @pragma Composition
 
   map(mapper, context) {
-    return updateSet(this, this._map.map(v => mapper(v, v, this), context));
+    return this.withMutations(set => {
+      this.forEach(value => {
+        const mapped = mapper.call(context, value, value, set);
+        if (mapped !== value) {
+          set.remove(value);
```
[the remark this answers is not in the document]
jessezhang91: Great point! Adjusted the tests to capture this point and updated the code per your recommendation.

src/Set.js
```diff
@@ -82,7 +82,19 @@ export class Set extends SetCollection {
   // @pragma Composition
 
   map(mapper, context) {
-    return updateSet(this, this._map.map(v => mapper(v, v, this), context));
+    return this.withMutations(set => {
+      const removes = [];
+      const adds = [];
+      this.forEach(value => {
+        const mapped = mapper.call(context, value, value, set);
```
leebyron: The last value here should be `this` - the original immutable map instead of the transiently mutable one.

PR #1609, opened 2018-09-24
New Option type

diwadidu:
This PR is a proof-of-concept for a new data type which I've been missing in Immutable: the Option type. Coming from a Scala background, I'm used to returning an Option from get requests and use it throughout my computations. 

The implementation here is not complete, and might need some modifications to align better with your coding patterns. I'm making this PR to gauge your interest in pursuing this further. I added unit tests for the currently implemented behavior as well as tests demonstrating the interoperability with the Immutable List, which should demonstrate how I imagine the three new types Option, None and Some to be used.

### DIFF
--- a/__tests__/Option.ts
+++ b/__tests__/Option.ts
@@ -1,26 +1,303 @@
 ///<reference path='../resources/jest.d.ts'/>
 
 declare var Symbol: any;
-import { None, Option, Set, Some } from '../';
+import { List, None, Option, Set, Some } from '../';
 
 describe('Option', function() {
-  it('should return a None if null is given on instantiation', () => {
-    const result = Option(null);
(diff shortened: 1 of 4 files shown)


leebyron:
I like using Option types, but I think this warrants it's own library. Mostly because there's not much you need to change about the existing library to introduce this, so it's easily decoupled. Also because there is very reasonable preference for and against using Option types (typescript/flow undefined-checking is a reasonable alternative) so I'd prefer this library stay agnostic to people's preferences.

I'm closing this PR since I don't think this library is the appropriate home, but I'd encourage you to start a new project which offers these types for those interested

diwadidu:
Thanks for taking the time to look at this and the suggestion to start a separate project. It didn't occur to me to do that, but I'll look into it.

PR #1617, opened 2018-10-10
Improve type infererence for fromJS

KSXGitHub:
With this pull request, `fromJS(value)` will no longer return `any`.

**Broken test:** This does not work in TypeScript 2.2, so I upgrade `dtslint`, which breaks linter. I don't want to mess with files that are not related to the change, so I reverted my attempt in fixing linter.

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -4827,12 +4827,43 @@ declare module Immutable {
    */
   export function fromJS(
     jsValue: any,
-    reviver?: (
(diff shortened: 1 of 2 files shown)


leebyron:
Awesome, thanks for doing this. Let's get TypeScript updated to prove support for 3.0 so this can be merged

jdeniau:
Superseded by https://github.com/immutable-js/immutable-js/pull/1927

Thanks @KSXGitHub

PR #1632, opened 2018-10-27
update security vulnerabilities

beast911:


leebyron:
I believe the travis ci check is failing because your version of yarn is outdated. I think if you update to the latest version and run `yarn add` for this new version bump that the yarn lock file will be generated correctly

leebyron:
actually - I take that back. Your change looks correct. It looks like Travis is failing because *it* has an outdated version of yarn. That's quite annoying!

beast911:
@leebyron shouldnt Travis CI pick the new yarn file from the commit ?

Issue #1634, opened 2018-10-29
Flow RecordOf<T> ignores T type 


a-shtubov:
Flow: 0.84.0
Immutable: 4.0.0-rc.11

The following code doesn't produce an error.
```
// @flow
import { Record } from 'immutable';
import type { RecordOf, RecordFactory } from 'immutable';

type Foo = {| foo: string |};

const fooFactory: RecordFactory<Foo> = Record({ foo: 1, bar: 2 });
const foo: RecordOf<Foo> = fooFactory();
```

acusti:
@leebyron Tagging this for inclusion in the 4.0 milestone

jdeniau:
Hi @a-shtubov ,

Do you still use flow ?
I'm considering dropping support as [typescript is far more used](https://npmtrends.com/flow-bin-vs-typescript) and I do not have any knowledge in flow, but I'd like your input on that.
Thanks

PR #1635, opened 2018-10-30
Fix flow RecordFactory type usage

leebyron:
This ensures use of Records result in appropriate flow types

Tested by ensuring the flow type tests continue to pass.


leebyron:
Should probably fix #1634

leebyron:
Trying a different approach

PR #1637, opened 2018-10-30
[BREAKING] Rename forEach to forSome, add forEach matching spec

conartist6:
This is an attempt to make Immutable more spec compliant. Currently a naive user would assume that `forEach` has the same implementation as the `forEach` function on native types, which does not involve early exit.

Also since this forEach now returns this, it can be chained!

### DIFF
--- a/__tests__/List.ts
+++ b/__tests__/List.ts
@@ -660,17 +660,30 @@ describe('List', () => {
     expect(a).toEqual([0, 1, 2, 3, 4]);
   });
 
-  it('forEach iteration terminates when callback returns false', () => {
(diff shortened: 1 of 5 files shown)


conartist6:
@tolmasky This is one of those changes that's very relevant to our discussion of breaking changes. I am proposing it specifically after thinking about what I said to you about considering the es6 API of Map and Set to be a gold standard that can be relied upon to be stable across versions. I'm going to need to do some research on whether Immutable itself has ever relied on being able to leverage this particular behavior of `forEach` when using `forEach` as an instance method of an object whose Immutable version is unknown.

PR #1653, opened 2018-11-27
Add separate types for Collection first()/last() with zero arguments

augustl:
When the type is `first<NSV>(notSetValue?: NSV): V | NSV`, the type of
the return value will always be `V | NSV` and you will have to handle
the possibility of the return value as being both of type V and
NSV. When you call first without specifying the type, it will default
to the type {}.

An example:

    List<String>(["a", "b", "c"]).first().length // Will fail

    List<String>(["a", "b", "c"]).first<string>().length // Will succeed

This patch will make the first succeed, by letting the type system
know that if no argument is passed to first(), the type will be String
in this case.

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -3878,7 +3878,8 @@ declare module Immutable {
      * In case the `Collection` is empty returns the optional default
      * value if provided, if no default value is provided returns undefined.
      */
-    first<NSV>(notSetValue?: NSV): V | NSV;
+    first<NSV>(): V | undefined;
+    first<NSV>(notSetValue: NSV): V | NSV;
 
     /**
      * In case the `Collection` is not empty returns the last element of the
@@ -3886,7 +3887,8 @@ declare module Immutable {
      * In case the `Collection` is empty returns the optional default
      * value if provided, if no default value is provided returns undefined.
      */
-    last<NSV>(notSetValue?: NSV): V | NSV;
+    last<NSV>(): V | undefined;
+    last<NSV>(notSetValue: NSV): V | NSV;
 
     // Reading deep values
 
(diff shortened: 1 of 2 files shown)


augustl:
Glad to help :) Force pushed an updated commit!

I'm not much of a flow user (and, frankly, quite the beginner in typescript as well), so the tests are passing, and "it looks good to me (tm)", but the flow part of this patch comes without warranty :)

(Not quite sure why the travis build failed, it worked yesterday, and it works on my machine if I do `npm run test`)

asazernik:
Looks good to me! On the more fundamental stuff like this, flow and TS are basically identical.

augustl:
Cool!

Not sure what to do about the failing tests. `npm run test` works fine on my machine, if I do `yarn install` I get the same error as in travis (`error An unexpected error occurred: "https://registry.yarnpkg.com/flatmap-stream/-/flatmap-stream-0.1.1.tgz: Request failed \"404 Not Found\"".`).

augustl:
Found the cause of the build error, created a separate pull request #1656 for that. Will rebase this pull request against master whenever master is updated :)

type-definitions/Immutable.d.ts
```diff
@@ -3878,15 +3878,17 @@ declare module Immutable {
      * In case the `Collection` is empty returns the optional default
      * value if provided, if no default value is provided returns undefined.
      */
-    first<NSV>(notSetValue?: NSV): V | NSV;
+    first<NSV>(): V;
```
leebyron: Should this be `V | undefined`?

augustl:
Very good point! That makes this patch much less important, I agree that the correct type is `V | undefined` and not just `V`. I updated the pull request and rebased it against master so the build doesn't fail, and I suppose `V | undefined` is marginally better than `V | NSV`.

augustl:
I just found a case where the explicit typing of `first()` with no arguments to `V | undefined` is needed, actually.

```
const xs = List<Foo>(...)
const firstX = xs.first()

firstX ? firstX.thing() : somethingElse
```

Currently, the type of firstX is `{} | Foo`, whereas with my patch the type is `Foo | undefined` so Typescript knows that firstX is now just `Foo` because of the ternary.

Temporary workaround is to use `get(0)` instead.

augustl:
Fixed in #1676 ftr :)

PR #1654, opened 2018-11-27
Correctly handle hashing of objects that return non-objects from valueOf()

### LINKED ISSUE #1643: hashing breaks for custom objects with "Invalid value used as weak map key"
I think the [latest changes to `Hash.js`](https://github.com/facebook/immutable-js/commit/a74700d22fa94b0bc1253fe166ee00487ca001d8#diff-8fbd09a266961312c5b404bb2a549e10) introduce a weird bug when hashing custom objects, more specifically, a class with `valueOf()` overridden. The error being thrown is `Uncaught TypeError: Invalid value used as weak map key`. 

It's strange because it doesn't happen until a certain point, a very specific point. It happens when invoking `.set()` on a `Map` on the 9th attempt. For the first 8 attempts, all is good. On the 9th attempt, error.

To reproduce, I ran this bit of code in the console at https://facebook.github.io/immutable-js.

```
class Example {
  constructor(id, key) { this.id = id; this.key = key; }
  valueOf() { return `${this.id}.${this.key}`; } 
  toString() { return `${this.id}.${this.key}`; }
}

const m = Immutable.Map().asMutable();
m.set(new Example('test', (new Date()).toISOString()), Math.random()); // 1, ok
...
m.set(new Example('test', (new Date()).toISOString()), Math.random()); // 8, ok
m.set(new Example('test', (new Date()).toISOString()), Math.random()); // 9, ERROR
```

This has worked for me for quite a while. It breaks with release `v4.0.0-rc.11`.

Here's a screenshot of the console:

screen shot 2018-11-06 at 12 17 27 pm

adamshone:
Fixes https://github.com/facebook/immutable-js/issues/1643.

JSFiddle showing the issue: https://jsfiddle.net/ey7opm1w

The problem is described well by @joakim-hagglund in this comment: https://github.com/facebook/immutable-js/issues/1643#issuecomment-437795012

The return value from `valueOf` was being passed to `hashJSObj`, and subsequently failing because non-objects cannot be used as keys in a `WeakMap`.

### DIFF
--- a/__tests__/issues.ts
+++ b/__tests__/issues.ts
@@ -101,3 +101,27 @@ describe('Issue #1293', () => {
     expect(secondState).toEqual(firstState);
   });
 });
+
+describe('Issue #1643', () => {
+  [
+    ['a string', 'test'],
+    ['a number', 5],
+    ['null', null],
+    ['undefined', undefined],
+    ['a boolean', true],
+    ['an object', {}],
+    ['an array', []],
+    ['a function', () => null],
+  ].forEach(([label, value]) => {
+    class MyClass {
+      valueOf() {
+        return value;
+      }
+    }
+
+    it(`Collection#hashCode() should handle objects that return ${label} for valueOf`, () => {
+      const set = Set().add(new MyClass());
+      set.hashCode();
+    });
+  });
+});
--- a/src/Hash.js
+++ b/src/Hash.js
@@ -10,6 +10,10 @@ import { smi } from './Math';
 const defaultValueOf = Object.prototype.valueOf;
 
 export function hash(o) {
+  if (o && o.valueOf !== defaultValueOf && typeof o.valueOf === 'function') {
+    o = o.valueOf();
+  }
+
   switch (typeof o) {
     case 'boolean':
       // The hash values for built-in constants are a 1 value for each 5-byte
@@ -30,9 +34,6 @@ export function hash(o) {
       if (typeof o.hashCode === 'function') {
         // Drop any high bits from accidentally long hash codes.
         return smi(o.hashCode(o));
-      }
-      if (o.valueOf !== defaultValueOf && typeof o.valueOf === 'function') {
-        o = o.valueOf(o);
       }
       return hashJSObj(o);
     case 'undefined':


timvracer:
Is this ever going to be incorporated?  Or is the position of immutable that they will not support this use case (in particular, using moment() )

adamshone:
I don't know :/

For me it wasn't moment, it was a [Decimal](https://github.com/MikeMcl/decimal.js/), but it breaks for anything that implements `valueOf`. So it definitely needs fixing.

@asazernik do you have any insight about what I can do to get this merged?

jelder:
The problem is easy to reproduce with Date. It looks like any value greater than `Date(8)` will crash Immutable. This is true even if the Date instance is part of a Map or Record, which is in turn a member of a Set. 

This makes Immutable completely break for our application, and we've had to refactor away all uses of Immutable Set. Hopefully this PR can get merged and a new rc released soon. 

```js
const { Set } = require("immutable")

const bad = []
let set = Set()
let i = 0

while (true) {
  i++
  try {
    set = set.add(new Date(i))
    console.log(i, "is ok")
  } catch (err) {
    console.log(i, err.message)
    bad.push(i)
  }
  if (bad.length > 100) {
    console.log(bad)
    break
  }
}
```

TikiTDO:
@leebyron You were the last one to complete a merge into this repo, so is there any chance to get this fix merged in? Or is there anyone else we can tag to get this resolved?

aaronwhite:
Lack of activity and commit history suggest this repo died on Valentine’s Day 2019 :/ RIP, even true, immutable-love isn’t forever 💔

asazernik:
@aaronwhite it kind of goes into hibernation, and then gets a burst of pulls every month or three.

EricTheRed-VT:
Can someone please merge this? The regression is 9 months old, and this fix was requested 8 months ago.

TikiTDO:
@EricTheRed-VT It honestly looks like this project is proper dead. It's been 5 months since a commit, and I haven't seen a single person with merge authority so much as comment. My recommendation is to move off of this library entirely sooner rather than later.

EricTheRed-VT:
@TikiTDO We're pretty happy with it in general, aside from the rare bug like this. Dropping Immutable would be too costly for my team, not to mention that there isn't a good replacement library (that I'm aware of). If this really never gets fixed, we might have to fork it an maintain a private version. 😞

jdeniau:
Included in #1833 via https://github.com/immutable-js/immutable-js/pull/1833/commits/45196e3c7c793e64daa134b72bb89b529f6d8318

PR #1656, opened 2018-11-30
Fix build error caused by flatmap-stream-0.1.1 yarn registry removal

augustl:
It was removed from the yarn registry due to a security issue, and
this project transitively depends on it via npm-run-all -> ps-tree ->
event-stream.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "marked": "0.3.19",
     "microtime": "2.1.8",
     "mkdirp": "0.5.1",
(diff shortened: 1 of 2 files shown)


augustl:
Gently bumping, anyone want to look into this? :) Currently the build breaks, because of a module that has been removed from the registry due to a security issue.

It's also blocking #1653

leebyron:
Thanks for fixing this

PR #1658, opened 2018-11-30
 Fix use of notSetValue in updateIn

### LINKED ISSUE #1657: `updateIn` doesn't properly handle `notSetValue`
### What happened

`updateIn` doesn't properly handle `notSetValue`.

### How to reproduce

```es6
const { updateIn } = require('immutable')
const original = { x: { y: { z: 123 } } }
const NOT_SET = { NOT_SET: true }
console.log(updateIn(original, ['x', 'y'], NOT_SET, val => NOT_SET)) // { x: { y: { NOT_SET: true } } }
```

Expected the log output to be `{ x: {} }` instead. Is this intended behavior?

migueloller:
Closes #1657

### DIFF
--- a/__tests__/updateIn.ts
+++ b/__tests__/updateIn.ts
@@ -31,6 +31,14 @@ describe('updateIn', () => {
     });
   });
 
+  it('remove in raw JS', () => {
+    const m = { a: { b: { c: 10 } } };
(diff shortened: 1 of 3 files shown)


migueloller:
The second commit is a bit out of band from the main issue. It seemed like a sensible change though. Would love to hear some thoughts on this.

PR #1661, opened 2018-12-10
You added es bundle, but forgot about sideEffects:false

a-x-:


IllusionMH:
Compromised `flatmap-stream` and `event-stream` were unpublished from npm. More info: https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident

ablakey:
@a-x- have you tested if this change causes tree shaking to work? Does it work well with the current structure of Immutable?  There's so much unused code I drag along in projects. I would love to see this capability be supported. 

Any help I can offer in testing, updating documentation?

PR #1673, opened 2019-01-09
Add new pick() method to Map

TLausmaa:
This adds a new `pick()` method to Map. This does the opposite of `deleteAll`, by returning a new Map which includes only the provided keys.

I felt like this sort of a convenience is missing, and would complement the existing functionality quite well in my opinion.

Thank you for taking the time to review this.

### DIFF
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -473,4 +473,27 @@ describe('Map', () => {
       Map([[a, Map([[b, Map([[c, 10], [d, 2], [e, 20], [f, 30], [g, 40]])]])]])
     );
   });
+
+  it('discards everything but the provided keys', () => {
+    const NOT_SET = undefined;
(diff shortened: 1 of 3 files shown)


__tests__/Map.ts
```diff
@@ -473,4 +473,27 @@ describe('Map', () => {
...
+    expect(m2.size).toBe(2);
+    expect(m2.get('b')).toBe(NOT_SET);
+    expect(m2.get('d')).toBe(NOT_SET);
+    expect(m2.get('a')).toBe(1);
+    expect(m2.get('c')).toBe(3);
+  });
+
+  it('does not alter the map if no keys are provided', () => {
+    const m1 = Map({ A: 1, B: 2, C: 3 });
+    const m2 = m1.deleteAll([]);
```
mmeinzer:
```suggestion
    const m2 = m1.pick([]);
```

PR #1675, opened 2019-01-12
[docs] Remove redundant double quotes

### LINKED ISSUE #1674: [docs] Redundant double quotes
### What happened
I think I found typos in [`Set`](https://github.com/facebook/immutable-js/blob/master/type-definitions/Immutable.d.ts#L1585) and [`List`](https://github.com/facebook/immutable-js/blob/master/type-definitions/Immutable.d.ts#L3477) operation results.
```js
const { Set } = require('immutable')
const intersected = Set.intersect([
  Set([ 'a', 'b', 'c' ])
  Set([ 'c', 'a', 't' ])
])
// Set [ "a", "c"" ]   <-- redundant double quote
```

metreniuk:
Fixes #1674

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -1582,7 +1582,7 @@ declare module Immutable {
      *   Set([ 'a', 'b', 'c' ])
      *   Set([ 'c', 'a', 't' ])
      * ])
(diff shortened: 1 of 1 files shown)


leebyron:
Not sure how that happened, but I'm glad you fixed it

PR #1676, opened 2019-01-16
Provide default type to first() and last()

johanblumenberg:
### Before

`new OrderedSet().first()` gives a compiler error, because the `NSV` type is not known.
`new OrderedSet<V>('a string').first()` returns type `V | string`

You need to explicitly provide the type to `first()`.

Intuitively you would pass type `V`:

`new OrderedSet().first<V>()` returns type `V`, although it can return `undefined`

The correct way would be:

`new OrderedSet().first<undefined>()` or `new OrderedSet().first(undefined)`, which would return type `V | undefined`

### After

`new OrderedSet<V>().first()` returns type `V | undefined`
`new OrderedSet<V>('a string').first()` returns type `V | string`

Less likely to make misstakes.
Less typing in the common case when you want `undefined` when there is no first value.

### DIFF
--- a/type-definitions/Immutable.d.ts
+++ b/type-definitions/Immutable.d.ts
@@ -3878,7 +3878,7 @@ declare module Immutable {
      * In case the `Collection` is empty returns the optional default
      * value if provided, if no default value is provided returns undefined.
      */
-    first<NSV>(notSetValue?: NSV): V | NSV;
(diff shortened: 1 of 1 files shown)


rgbkrk:
If I could merge this, I would.

leebyron:
This is the right thing. thanks!

PR #1679, opened 2019-01-21
Fix flip() example under "Seq is lazy"

dharkness:
The `Seq` example using `flip()` has an incorrect result (all 1s).


leebyron:
Thanks for improving the docs!

PR #1686, opened 2019-01-30
Replace any by unknown in typescript types

ThomasdenH:
This PR replaces all usage of `any` by `unknown`. Using any in Typescript is unsafe because it allows code that does no type checking. Because it could require type checking where it didn't before, it is a breaking change. Also, `unknown` requires [Typescript 3.0](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12340 @@
+{
+  "name": "immutable",
+  "version": "4.0.0-rc.12",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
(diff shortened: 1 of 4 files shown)


ThomasdenH:
It should by noted that older code can easily be fixed either by casting or explicitly specifying generics as `any`.

leebyron:
How does unknown work in input positions? Like the arguments to a function?

Also it looks like this PR is failing tests, could you please investigate and add a fix? You should be able to run `npm test` locally

ThomasdenH:
In input positions `unknown` works no different than `any`, anything is assignable to it. Of course, when writing the actual function bodies, `any` would allow you to perform any operation on the arguments, while `unknown` would force you to be type safe. Using it in definitions is mainly for consistency.

`npm test` did not work on Windows. I get the following error:
```
> prettier --single-quote --trailing-comma=es5 --write "{__tests__,src,pages/src,pages/lib,perf,resources}/**/*{\.js,\.ts}"

[error] No matching files. Patterns tried: {__tests__,src,pages/src,pages/lib,perf,resources}/**/*{\.js,\.ts} !**/node_modules/** !./node_modules/** !**/.{git,svn,hg}/** !./.{git,svn,hg}/**
```

I did run it on bash for Windows and pushed the result.

leebyron:
Awesome, thank you!

PR #1693, opened 2019-02-15
create new data structure called stretch

eitanschreiber123:
create new data structure called stretch(base, ends, front) that returns an IndexedSeq with 3 values base - ends, base, base + front and if front is not it uses ends


jdeniau:
Hi @eitanschreiber123 

I'm digging some old PR and issues of immutable.
I must say that I do not understand the purpose of this method, and I do not remember another structure like this.

Could you help me understand a use case for that?
Thank you!

PR #1684, opened 2019-01-29
add 'Complementary Tools' section to website

eitanschreiber123:
add a new section to the website with a list of libraries and tools to use with immutable-js

PR #1695, opened 2019-02-19
add new data structure called IterateUntil

eitanschreiber123:
add a new data structure called IterateUntil(callback, condition, seed = 1) that calls callback with seed as the argument and passes the result to callback as long as condition returns true and returns a Seq.Indexed of the results 
For the type-definitions i wrote for flow but i still have to figure out for typescript for defining function types i tried Function, and ((_: any) => any) but both returned errors and the only other option is to use interfaces or the type keyword

### DIFF
--- a/__tests__/IterateUntil.ts
+++ b/__tests__/IterateUntil.ts
@@ -1,0 +1,32 @@
+/**
+ * Copyright (c) 2014-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+///<reference path='../resources/jest.d.ts'/>
+
+import { IterateUntil } from '../';
+
+describe('IterateUntil', () => {
+  it('fixed iterateuntil', () => {
+    const v = IterateUntil(n => n + 1, a => a < 5);
+    expect(v.size).toBe(3);
+    expect(v.first()).toBe(2);
+    expect(v.rest().toArray()).toEqual([3, 4]);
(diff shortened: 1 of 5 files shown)


conartist6:
The community is trying to spark the project back into maintenance, but I do not think your PR will be merged. In general the great thing about iterable sequences is that there's very little penalty for chaining more than one operation together to achieve a desired result. `flatMap` is a great example! On an array, calling `arr.map().flat()` would generate an intermediate array of arrays which is immediately discarded. When working with iterables `seq.map().flat()` is just as idiomatic as `seq.flatMap()`, though `flatMap` is usually included as a convenience since the name and meaning of the operation are familiar from arrays.

So that said, I see no reason that what you want to accomplish can't be done with several operations over a sequence, which also I think would make the code clearer to anyone who isn't you.

I think it will probably look something like this:

```js
const result = Seq.Indexed(data)
  .takeWhile((item, i) => shouldTakeItem(item, i));
```
No you can iterate over your results using a `for (const item  of result)` loop, or using `result.forEach(callback)`.

eitanschreiber123:
yes thnk you that works better and acheives the same result so i will close this pull request, also can i ask about my pull request #1684 there was a problem with the styling but i fixed it

conartist6:
Lee is currently the only person with commit privileges. He has indicated that he's willing to transition to community maintenance, and a community has formed and is ready to carry out the maintenance. We've created [a slack](htps://immutable-js.slack.com) to organize. All we need now are privs but we're still waiting on Lee there.

PR #1696, opened 2019-02-19
add new data structure called Times

eitanschreiber123:
add a new data structure called 
times(times, callback, seed = 1) that calls callback with seed passed as the argument and passes the result to callback the times number of times and returns a Seq.Indexed of the results 
For the type-definitions i wrote for flow but i still have to figure out for typescript for defining function types i tried Function, and ((_: any) => any) but both returned errors and the only other option is to use interfaces or the type keyword

### DIFF
--- a/__tests__/Times.ts
+++ b/__tests__/Times.ts
@@ -1,0 +1,32 @@
+/**
+ * Copyright (c) 2014-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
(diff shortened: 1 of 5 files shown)


conartist6:
See my comments on #1695. I think this idea makes more sense, but personally I'd still be inclined not to do it. It's just more difficult to read than `Range().map(_ => ...)`.

jdeniau (commented):
Hi,
It looks like your data structure is just a syntaxic sugar over Range.
The "seed" seems to be a really specific use-case.

I will close this as I do not see a real use case for that and I do not want to extend the external API for too specific code.

PR #1697, opened 2019-03-02
Fix typescript definition tests

berickson1:
I was preparing another change to improve some typings in this library but discovered with the introduction of the `unknown` return type in immutable's d.ts, tests were broken. (#1686)

This PR contains the following:
Upgrade tslint, typescript & dtslint to the latest versions
Run tests against typescript 3.0 (first version of typescript compatible with the `unknown` type)
Add default generic to most constructors that return Class<unknown> when no type can be inferred
Fix lint rule violations (caused by upgrading dtslint)
Update d.ts tests to reflect usage of `unknown` type

Unfortunately strict function types had to be left disabled for the time being, but I hope to post another PR that'll add more typings & enable this typescript compiler flag


berickson1:
Closing as I no longer have a project using immutable to validate against & it's been left un-reviewed for ~8 months. Happy to let someone else revive this if needed.

PR #1728, opened 2019-08-11
Change the type of set container depending on the type of key.

volkanunsal:


jdeniau:
@volkanunsal I am digging up this issue, but it seems to be the logical thing to do. 
I will try to implement that in 6.0 (it's breaking so I'm not updating the 5.x branch)

PR #1764, opened 2020-03-17
fixed Invalid value used as weak map key #1643

### LINKED ISSUE #1643: hashing breaks for custom objects with "Invalid value used as weak map key"

abrcdf1023:
fixed https://github.com/immutable-js/immutable-js/issues/1643

### DIFF
--- a/__tests__/issues.ts
+++ b/__tests__/issues.ts
@@ -101,3 +101,27 @@ describe('Issue #1293', () => {
     expect(secondState).toEqual(firstState);
   });
 });
+
+describe('Issue #1643', () => {
+  [
+    ['a string', 'test'],
+    ['a number', 5],
+    ['null', null],
+    ['undefined', undefined],
+    ['a boolean', true],
+    ['an object', {}],
+    ['an array', []],
+    ['a function', () => null],
+  ].forEach(([label, value]) => {
+    class MyClass {
+      valueOf() {
+        return value;
+      }
+    }
+
+    it(`Collection#hashCode() should handle objects that return ${label} for valueOf`, () => {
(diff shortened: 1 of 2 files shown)


Methuselah96:
Closing since this is an exact duplicate of https://github.com/immutable-js/immutable-js/pull/1654.

PR #1744, opened 2019-10-24
Update link to docs in readme

Emobe:
This fixes #1743

PR #1775, opened 2020-07-11
Fix documentation link

amdw:


amdw:
Seems this is a duplicate of #1744, which is quite old now. A consequence of #1689 I suppose. :(

PR #1777, opened 2020-08-14
Fix OrderedMap#deleteAll()

Methuselah96:
Resolves https://github.com/immutable-js/immutable-js/issues/1701.

Explanation of bug fix:
`deleteAll` uses `withMutations` to delete keys from an ordered map. `OrderedMap` tracks whether it's been mutated by checking to see if its internal `_map` or `_list` have been mutated. This is not sufficient because when we resize the list [here](https://github.com/immutable-js/immutable-js/blob/91c7c1e82ec616804768f968cc585565e855c8fd/src/OrderedMap.js#L138-L148), we create a new map and a new list that don't realize that they are part of a mutation. Therefore, `OrderedMap` needs to be tracking its own `__altered` state instead of relying on the underlying altered state of its map and/or list. This bug was introduced in https://github.com/immutable-js/immutable-js/pull/1331 because we actually fixed the way that we were tracking mutations in a `List` which ended up causing `OrderedMap` to break.

### DIFF
--- a/__tests__/OrderedMap.ts
+++ b/__tests__/OrderedMap.ts
@@ -7,7 +7,7 @@
 
 ///<reference path='../resources/jest.d.ts'/>
 
-import { OrderedMap, Seq } from '../';
+import { OrderedMap, Range, Seq } from '../';
 
 describe('OrderedMap', () => {
   it('converts from object', () => {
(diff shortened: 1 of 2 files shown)


jdeniau:
This has been merged in the OSS fork #1833

PR #1779, opened 2020-08-14
Remove incorrect size property from Record in doc example

Methuselah96:
Resolves https://github.com/immutable-js/immutable-js/issues/1763.


jdeniau:
Integrated in #1833

PR #1840, opened 2021-07-06
Prevent mutable structures when merging Immutable objects using mergeDeep

### LINKED ISSUE #1475: DeepMerge transforms Map to [key, value] Array in some conflicts
Hello.
Using ```immutable@4.0.0-rc.9```
I was doing a ```.deepMerge``` and I ran into a conflict between a List and Map like this one: 
```javascript
const a = Immutable.fromJS({
    ch: [{
        code: 8
    }]
}); 
const b = Immutable.fromJS({
    ch: {
        code: 8
    }
});
const result = a.mergeDeep(b); 
/* result:
Map {
    'ch': List[
        Map { 'code': 8 },
        Array [ 'code', 8 ]
    ]
}
*/
```
Then I tried this:
```javascript
const a = Immutable.fromJS({
    ch: []
}); 
const b = Immutable.fromJS({
    ch: { code: 8 }
});
const result = a.mergeDeep(b); 
/* result:
Map {
    'ch': List[
        Array [ 'code', 8 ]
    ]
}
*/
```
I guess there's a problem when merging a key with a List value and a key with a Map value (the Map has 
 to have at least one key inside).
So I'm baffled by that Array there. 
Is this a bug?

Thank you.

jdeniau:
Migrate https://github.com/immutable-js-oss/immutable-js/pull/220

Resolves #1475 and https://github.com/immutable-js-oss/immutable-js/issues/70.


rgbkrk (approved):

bdurrer:
Also see conversation of immutable-js-oss#215

bdurrer:
We discussed the issues of merging unmergable types a few times, and it boils down to "there is no general _fixed_ solution that makes everyone happy".

It [probably] started at immutable-js-oss#215 and was discussed on slack a few times, the latest discussion was [in june](https://immutable-js.slack.com/archives/C018PFV575G/p1625059417123700).

Problem: Merging Lists (collections without keys) and Maps (keyed Collection) almost certainly results in unwanted behaviour

### A few possible options:

1. Deny merging them and throw a TypeError
2. Silently skip merging these and keep the old values.
3. Sliently overwrite the whole conflicting collection with the new data
4. Add a merger param (like `mergeWith`) that is called for every value, which means also for every [sub]collection
5. Add a conflict param that is called for every incompatible data type combination

### We concluded that
- mergeDeep is probably not widely used
- None of us was very comfortable with mergeDeep, we think it is a dangerous function anyway
- some [wrote their own](https://immutable-js.slack.com/archives/C018PFV575G/p1625062094142300) mergeDeep using recursive mergeWith, which matched their needs (see below)
- We know of one live use-case: storing/retrieving of data from browser store, which could contain incompatible data from a previous app version. Solution 2 would be their for this


### Pro / Cons of these solutions

#### 1 - Type Error
- Works well with explicit typing (TS)
- Breaks with the current behavior of methods (nothing ever throws)
- The user must deal with Errors by catching, which probably is an annoying behaviour in a React reducer
- Doesn't solve the issue, only uncovers it

#### 2 - keep existing collection

- Predictable behavior
- Works well with static typing (TS)
- Bad for legit use cases where types _do_ change, e.g. collection with mixed data types
- If you don't like the behavior, you must implement your own mergeDeep (like Andrew's team did)

#### 3 - keep new collection

- Predictable behavior
- Bad for static typing (TS)
- If you don't like the behavior, you must implement your own mergeDeep (like Andrew's team did)

#### 4 - merger param
- (positive) you have full control
- (negative) you have full control. Changing the behaviour means you have to provide a full merger function, you might be faster by implementing a custom solution with `mergeWith`
- In contrast to `mergeWith` it is called for both the collection(s) and the valules of these collections. Implementing it would probably be confusing.

#### 5 - conflict param:
- control over conflicts
- default could be "keep old", which is good for static typing (TS)
- default could log to console to make users aware of the issue
- we could make the default overwriteable, e.g. something like `Immutable.conflictFunction = (a, b) => a;`


Custom solution provided by Andrew Patton:

```
const isMergeable = (a) =>
    a && typeof a === 'object' && typeof a.mergeWith === 'function' && !List.isList(a);
export const mergeDeep = (a, b) => {
    // If b is null, it would overwrite a, even if a is mergeable
    if (isMergeable(a) && b !== null) {
        return a.mergeWith(mergeDeep, b);
    }
    if (!List.isList(a) || !List.isList(b)) {
        return b;
    }
    return b.reduce((acc, nextItem, index) => {
        const existingItem = acc.get(index);
        if (isMergeable(existingItem)) {
            return acc.set(index, existingItem.mergeWith(mergeDeep, nextItem));
        }
        return acc.set(index, nextItem);
    }, a);
};
```

jdeniau:
My preferred option according to the current API of immutable goes to 3 : I don't think it's bad in strictly typed land. In fact, you probably don't want to go that in TS/flow as it is really an opposite to strict typing.
If TS/flow does throw an error here, I think it is fine.

In JS-land, I find more logical that :

```js
Map({ l: List(['foo']) })
  .mergeDeep({
    l:  Map({foo: bar }
  })

// { l: {foo:'bar'} }
```

Without the current API of immutable, I think that every weird comportment should throw instead of behing silent. Explicit is better that implicit. And in this case I do prefer option 1.

leebyron:
I agree with all the analysis here. `mergeDeep` is not always predictable and while some iterations on improving that behavior should be welcomed, we should probably be careful not to change its existing semantics too much. Better behavior might be best reserved for a new function (if there's demand)

> 3 - keep new collection

This seems like the most reasonable choice to me. `mergeDeep` is already a not very explicit type friendly operation in the first place, so I don't put too much weight on that. The key semantic of "merge" is "set multiple times" and "mergeDeep" is "set multiple times, in potentially deep positions in a data structure" - I can't reconcile that semantic with a behavior that keeps an old value (does not set) or throws an error for something that could otherwise be merged, albeit in a non-type-safe way.

At a high level I think a mergeDeep algorithm should be:

* If `a` and `b` are mergeable:
  * Let `ab` have all `(k, v)` entries of `a`
  * For each entry `(k, v)` of `b`:
    * If `k` exists in `a`: let `ab[k]` be `mergeDeep(a[k], v)`
    * Otherwise let `ab[k]` be `v`
  * Return `ab`
* Otherwise return `b`

src/functional/merge.js
```diff
@@ -68,7 +72,14 @@ export function mergeWithSources(collection, sources, merger) {
         }
       };
   for (let i = 0; i < sources.length; i++) {
-    Collection(sources[i]).forEach(mergeItem);
+    const source = sources[i];
+    if (
+      (isArray && (isPlainObject(source) || isKeyed(source))) ||
+      (isObject && (Array.isArray(source) || isIndexed(source)))
+    ) {
+      continue;
```
leebyron, line 80: Similar here - is data being lost?

leebyron (changes requested):
I suspect the correct fix will require changing the `mergeDeep` behavior of determining if two deep values are mergeable rather than changing the existing behavior of `merge` and `concat` (though it does seem like `concat` needs behavior refinement)

Methuselah96:
Thanks for the helpful feedback. Some of the shortcomings mentioned in the review were intentional since this was just a proof of concept to get the conversation started.

There's one class of cases I'm still not sure about. How should the following behave:
```
Map({ foo: 'bar' }).merge(List(['baz']))
merge({ foo: 'bar' }, ['baz'])
Map({ foo: 'bar' }).mergeDeep(List(['baz']))
mergeDeep({ foo: 'bar' }, ['baz'])
```

Should these return the `Map` or the `List`? Presumably if this were nested in a `mergeDeep` scenario it should return the `List`, but do we want to make an exception for the top level for either/both `merge` or/and `mergeDeep` so that it's still a `Map`/`object` or do we want to go for more across-the-board consistency and have these return the `List`?

bdurrer:
> There's one class of cases I'm still not sure about. How should the following behave:
> 
> ```
> Map({ foo: 'bar' }).merge(List(['baz']))
> merge({ foo: 'bar' }, ['baz'])
> Map({ foo: 'bar' }).mergeDeep(List(['baz']))
> mergeDeep({ foo: 'bar' }, ['baz'])
> ```
> 
> Should these return the `Map`/object or the `List`/array? Presumably if this were nested in a `mergeDeep` scenario it should return the `List`/array, but do we want to make an exception for the top level for either/both `merge` or/and `mergeDeep` so that it's still a `Map`/object or do we want to go for more across-the-board consistency and have these return the `List`/array?

Interesting question. keeping the toplevel makes sense, but it adds a special case that might be confusing too.

I just realized that `Object.assign` solves this in an clever way:
```
var obj = { hello: 'world' };
var arr = ['ooh'];
 Object.assign(obj, arr);
// Object { 0: "ooh", hello: "world" }

Object.assign(arr, obj);
// Array [ "ooh" ]
```

Maybe we also could just treat indexes as keys? 
`Map({ foo: 'bar' }).merge(List(['baz']))` would produce `Map({ foo: 'bar', 0: List(['baz']) })`?

Methuselah96:
Nice observation, I'd be fine with that behavior as well with the added benefit that it's consistent with `Object.assign`.
@leebyron What's your opinion?

leebyron:
I like the appeal to consistency with core JS API, and this behavior makes sense.

For documentation and our own sanity sake, do we have a high level algorithm for what the final behavior should be?

Seems like this special case is: if the merge destination is a keyed collection, convert the merge source to a keyed sequence before merging?

leebyron:
My one concern though is whether this is the behavior most will want. It makes sense for a shallow merge where you need to imply _some_ reasonable behavior. Which is what Object.assign is doing. But when merging deeply, is this what most would expect?

My thought was the main use for mergeDeep was to apply an update to a complex data structure with a known shape. It's not too uncommon to have `Thing | Array<Thing>` in that shape, and if your intent with a deep merge was to replace a `Thing` with a `Thing[]` then this new behavior would be frustrating

Methuselah96:
Yeah, I share that concern. I think my slight leaning is to special-case the top level to mimic `Object.assign` and below that replace the data structure if there's a conflict.

bdurrer:
I think Methuselah96's proposal (object.assign on top level, replace behavior for rest) is the best we can do, considering that there is no golden bullet. 
Letting the user decide could be the next step, but I believe it would not be so implement for both library and user - and as Lee suggested, might/should be an entirely different function.

The only non-derpy use-case we heard of was storing and retrieving data into session storage (oder some other serialized form), where you never can't be entirely sure the data still matches the current app's data structure.
I believe the proposal should work with that use case (maybe?).

leebyron:
I think the Object.assign behavior at the top level makes sense. However for the sake of keeping this specific PR/issue focused, I might suggest limiting changes to the top level merge strategy that aren't directly solving this bad behavior.

Another thing to consider is that the top level merge behavior should be as similar as possible to `merge()` behavior and we should avoid breaking changes.

Right now at top level this just throws a TypeError in many cases, but there is existing support for Lists of tuples we should avoid breaking (at least until a future major version)

```
Immutable.Map({a:'X'}).merge(Immutable.List(['Y']))
> Uncaught TypeError: Expected [K, V] tuple: Y
Immutable.Map({a:'X'}).merge(Immutable.fromJS([['b', 'Y']]))
> Map { "a": "X", "b": "Y" }
```

Methuselah96:
@bdurrer @leebyron Alright I've redone it. Let me know what you think. It's a much more localized change now that will only affect `mergeDeep`.

Methuselah96:
Note that this only overwrites the value when we're merging a list into a map (as in https://github.com/immutable-js/immutable-js/issues/1475) and not vice versa (as in https://github.com/immutable-js/immutable-js/issues/1719). I'm not sure what to do if we're merging a list into a map because we need to continue to support merging a list of tuples into a map and I can't think of any obvious way to know whether a list is intended to be a list of tuples or not. It seems like the current approach is inconsistent if we leave it like it is. Any thoughts of how to overcome this?

Methuselah96:
I guess we could determine if it's a list of tuples by checking to see if each element in the list has a length of two. But I'm not sure if that's getting too complicated or not.

leebyron:
Yeah I agree with your concern, I don't think sniffing for tuples is sustainable. Considering that the root concern here is that this behavior is too complicated then we should probably bias to a simpler solution that is more explainable, even if that solution covers less useful surface area as a result.

We might even want to make the behavior for "list into map" and "map into list" the same, total replacement. What's the cost of that?

leebyron:
Very nice and focused change in this update. Great work

bdurrer:
Just realized I can't review, I probably don't have any rights 🤷 

Since we are discussing it all the time, I suggest to also add test case(s) for partial and absolute conflicts:

e.g. (untested):

```
    const a = fromJS({
      ch: [
        {
          code: 8,
        },
      ],
      banana: 'good',
    }) as Map<unknown, unknown>;
    const b = fromJS({
      ch: {
        code: 8,
      },
      apple: 'anti-doctor',
    });
    expect(a.mergeDeep(b).equals({
      ch: {
        code: 8,
      },
      apple: 'anti-doctor',
      banana: 'good',
    })).toBe(true);
```

and absolute conflict:

```
    const a = Map({
      banana: 'good',
    }) as Map<unknown, unknown>;
    const b = List(['banana']);
    expect(a.mergeDeep(b).equals(b)).toBe(true);
```

Methuselah96:
> We might even want to make the behavior for "list into map" and "map into list" the same, total replacement. What's the cost of that?

Looks like the only failing test with making that change would be [this test](https://github.com/immutable-js/immutable-js/pull/1404/files#diff-a3728fcf53279f3b747758c3692e1286a839156267bccbf8eda3b122fa59ed4dR203-R220) checking that you can merge tuples of symbols into a `Map` using `mergeDeep`.

Methuselah96:
> Just realized I can't review, I probably don't have any rights 🤷

@leebyron Can we add @bdurrer as a maintainer? @bdurrer's been extremely helpful since the beginning.

Methuselah96:
> And please, could we stop nesting ternary conditions and just write if statements? That stuff is gonna blow up in our face on day 😅

Personally I find nested ternaries a lot easier to read and much more succinct in many situations. Now if only Prettier formatted them without flattening them... (although I am getting used to the flattening)

Methuselah96:
> and absolute conflict:
> 
> ```
>     const a = Map({
>       banana: 'good',
>     }) as Map<unknown, unknown>;
>     const b = List(['banana']);
>     expect(a.mergeDeep(b).equals(b)).toBe(true);
> ```

This test currently throws with:
```
  ● merge › mergeDeep merges absolute conflicts

    TypeError: Expected [K, V] tuple: banana

      839 | function validateEntry(entry) {
      840 |   if (entry !== Object(entry)) {
    > 841 |     throw new TypeError('Expected [K, V] tuple: ' + entry);
          |           ^
      842 |   }
      843 | }
      844 | 
```

I would be inclined to think that throwing an error for this situation is fine. The impetus for this PR and for the attached issues is specifically that people expect `mergeDeep` to replace nested collections if the collections aren't mergeable. I think it's reasonable for the top-level of `mergeDeep` to behave identically to `merge` and it doesn't seem like people are bothered by the fact that merging a List into a Map at the top level results in a `TypeError` if it's not a list of tuples.

bdurrer:
Sorry if my test implied an expected behavior. I agree with the reasoning (and if you are brave enough to use `mergeDeep`, at least wrap it in a try-catch and deal with conflicts).

Methuselah96:
Build failing due to https://github.com/nodejs/node/issues/40030.

src/functional/merge.js
```diff
@@ -69,10 +72,19 @@ export function mergeWithSources(collection, sources, merger) {
...
       : merger
       ? merger(oldValue, newValue, key)
       : newValue;
   }
   return deepMerger;
 }
+
+function areMergeable(oldValue, newValue) {
+  return !(
+    (isIndexed(oldValue) || Array.isArray(oldValue)) &&
```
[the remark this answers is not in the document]
Methuselah96: Hmm, I'm open to that. The reason I kept it so narrow is because these checks match the checks in `isDataStructure` which is what is used in `deepMergerWith`. Do you still think it's worth updating?

src/functional/merge.js
```diff
@@ -69,10 +72,19 @@ export function mergeWithSources(collection, sources, merger) {
...
+        ? mergeWithSources(oldValue, [newValue], deepMerger)
+        : newValue
       : merger
       ? merger(oldValue, newValue, key)
       : newValue;
   }
   return deepMerger;
 }
+
+function areMergeable(oldValue, newValue) {
```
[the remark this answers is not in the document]
Methuselah96: Yeah, no problem. I might be changing this anyway depending on what we decide on https://github.com/immutable-js/immutable-js/pull/1840#issuecomment-915296624. Should we go ahead and make it so that lists replace maps as well?

src/functional/merge.js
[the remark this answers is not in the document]
leebyron: I have low confidence, but that was my instinct. Maybe something you can confirm with test coverage.

src/functional/merge.js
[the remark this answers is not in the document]
leebyron: IMO that's probably the right behavior since it is most simple and I think having a narrower behavior for the deep merge vs the top level is right.

I think we should be able to keep the top level support for merging lists-of-tuples while using the simpler replace behavior for deep conflicts.

src/functional/merge.js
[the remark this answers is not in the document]
Methuselah96: Done.

src/functional/merge.js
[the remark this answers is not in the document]
Methuselah96: Done.

src/functional/merge.js
```diff
@@ -69,10 +72,22 @@ export function mergeWithSources(collection, sources, merger) {
...
 }
+
+/**
+ * The data structures are considered not to be mergeable if one of them is indexed and the other is keyed.
+ */
+function areMergeable(oldDataStructure, newDataStructure) {
+  return !(
+    (isIndexed(Seq(oldDataStructure)) && isKeyed(Seq(newDataStructure))) ||
+    (isKeyed(Seq(oldDataStructure)) && isIndexed(Seq(newDataStructure)))
+  );
```
leebyron, line 92: How does this handle `Set` types?

Should this be something like

```suggestion
  return (
    (isIndexed(Seq(oldDataStructure)) === isIndexed(Seq(newDataStructure))) &&
    (isKeyed(Seq(oldDataStructure)) === isKeyed(Seq(newDataStructure)))
  );
```

src/functional/merge.js
```diff
@@ -69,10 +72,22 @@ export function mergeWithSources(collection, sources, merger) {
 function deepMergerWith(merger) {
   function deepMerger(oldValue, newValue, key) {
     return isDataStructure(oldValue) && isDataStructure(newValue)
-      ? mergeWithSources(oldValue, [newValue], deepMerger)
+      ? areMergeable(oldValue, newValue)
+        ? mergeWithSources(oldValue, [newValue], deepMerger)
+        : newValue
```
leebyron, line 77: Curious what you think about calling the custom merger in this scenario? Previous logic is to do this

```suggestion
      && areMergeable(oldValue, newValue)
      ? mergeWithSources(oldValue, [newValue], deepMerger)
```

Methuselah96: Yeah, I definitely considered it. My main thought was that it would be an additional breaking change since the user would need to handle more cases (although I can't imagine it would have that big of an impact), but it would provide more flexibility for the user. I'm in favor of it if we're not concerned about it being a breaking change.

leebyron: Yeah the domain in which we're changing behavior is already in the weird cases. My thought was that it would be easier to explain the behavior of the function. Essentially both before and after the change we could correctly say "if we can merge em we do, if not we call your merge conflict function which has a default of `(old, new) => new`

leebyron: I think we also have some license to make very modest breaking changes in the name of fixing bad behavior since the plan is land pre final 4.0

src/functional/merge.js
Methuselah96: > How does this handle `Set` types?

It doesn't. 😄

I'm open to that change, I was just erring on the side of being more permissive in case there were cases I wasn't considering. How *do* we want to handle sets? Should they be mergeable with lists, but not maps?

leebyron: If we're optimizing for "easy to explain" then I think we probably want to follow the model here. Lists only merge into Lists, Maps into Maps, Sets into Sets. What do you think?

Methuselah96: Yeah, that seems the most straightforward rule. Although Maps into Maps and Sets or Lists into Sets or Lists could also be fairly easy to explain. I'm trying to think what people would be expecting and I'm having a hard time guessing since I can see it going either way as far as desired behavior. I guess the nice thing about keeping it more strict is that it will call the merger function and so that gives the user the ability to easily override it.

Methuselah96: Updated.

src/functional/merge.js
Methuselah96: Updated.

src/functional/merge.js
leebyron: >  I guess the nice thing about keeping it more strict is that it will call the merger function and so that gives the user the ability to easily override it. The other way it would be much harder to override.

That really convinces me!

src/functional/merge.js
```diff
@@ -68,11 +72,23 @@ export function mergeWithSources(collection, sources, merger) {
...
   return deepMerger;
 }
+
+function areMergeable(oldDataStructure, newDataStructure) {
+  const oldSeq = Seq(oldDataStructure);
+  const newSeq = Seq(newDataStructure);
+  return (
+    isIndexed(oldSeq) === isIndexed(newSeq) &&
+    isKeyed(oldSeq) === isKeyed(newSeq) &&
+    isSet(oldSeq) === isSet(newSeq)
```
leebyron, line 92: You may need some test coverage to cover this, but I believe `isSet` only checks if something is actually `Set` or `OrderedSet` and doesn't include `SetSeq`. We don't have a great first class function for this, something is set-like if it is both not indexed and not keyed

Methuselah96: I wondered about that, I should've looked into that more, makes sense.

leebyron:
This is looking great! Any additional tests you want to add? Documentation improvements part of this PR or a follow up?

What do we need to do to unblock the unit test CI spurious failure?

Methuselah96:
I'll probably add some more tests to make sure the overriding is working as expected.

I'm willing to add more documentation. Any ideas of what would be helpful? Would some examples be helpful or maybe expanded clarification on what we consider mergeable?

The build will be fixed once https://github.com/nodejs/node/issues/40030 is released. We could either:
 - Wait until that is released in a patch release to merge this PR.
 - Set the Node version in the CI to 16.8 and then revert the change once a patch is released for 16.9.
 - Merge this without the CI passing since we know it passes locally and just wait for the patch to be released.

Methuselah96:
Looks like the release of the fix [is imminent](https://github.com/nodejs/node/pull/40069), so we can probably just wait it out since I still have some more tests and possibly documentation to add.

leebyron:
Sounds good. Seems like the build is going to be fixed upstream today.

For documentation, given it's been mentioned a few times that this function's behavior can be confusing, I think it's worth adding:

* An explanation of when `mergeDeep` will recurse vs when it will consider something a "conflict" (ie the collections have to be of similar type - List/Map/Set)
* How `mergeDeep` will treat conflicts by always taking the merged in value and that this behavior can be replaced by `mergeDeepWith`
* How Lists and Sets are "merged" via concat/union

I think it would also be helpful to do a scan for some consistency across the few places we document these both as [instance methods](https://immutable-js.com/docs/latest@main/Map/#mergeDeep()) and as the [top level function](https://immutable-js.com/docs/latest@main/mergeDeep()/)

For tests, it would be great to explicitly include a test for a scenario that looks like a deep-edit but is actually another example of #1475

```js
a = Map({ a: List([ Map({ x: 1 }) ]))
b = Map({ a: Map([[ 0, Map({ y: 1 }) ]]) })
a.mergeDeep(b)
```

leebyron:
We should probably also add a test for the functional case:

```js
a = { a: [ { x: 1 } ] }
b = Map({ a: Map([[ 0, Map({ y: 2 }) ]]) })
mergeDeep(a, b)
```

Methuselah96:
Added tests and documentation.

src/functional/merge.js
```diff
@@ -68,11 +71,22 @@ export function mergeWithSources(collection, sources, merger) {
...
+      areMergeable(oldValue, newValue)
       ? mergeWithSources(oldValue, [newValue], deepMerger)
       : merger
       ? merger(oldValue, newValue, key)
       : newValue;
   }
   return deepMerger;
 }
+
+function areMergeable(oldDataStructure, newDataStructure) {
```
leebyron, line 85: Nit: it would still be nice to include a comment block above this function explaining why this implementation determines things are considered mergeable

leebyron (approved):
Excellent work. Great tests and docs improvements

src/functional/merge.js
Methuselah96: Done. Let me know if that's what you were thinking or if you were asking for a different explanation.

leebyron: Yeah this looks good to me. It's the connection between the collection categorization and this somewhat nuanced boolean logic that will be helpful when we look at this function again after our collective memory is lost

leebyron:
One small note before this squash-merge lands:

We should make sure the commit message includes a note about the aspect of this PR which is a breaking change so we can clearly communicate that in the release notes for the next RC.

I think that's just the behavior related to the one removed unit test (eg. lists of tuples no longer deep merge into maps)?

Methuselah96:
Yeah, that sounds like the only "expected" behavior that has been changed in this PR.

PR #1841, opened 2021-07-06
Improve Typescript definition for Map

jdeniau:
Imagine the following code:

```ts
const m = Map({ length: 3, 1: 'one' });
```

This was previously typed as `Map<string, string | number>`

and return type of `m.get('length')` or `m.get('inexistant')` was typed as `string | number | undefined`.

This made `Map` really unusable with TypeScript.

Now the Map is typed like this:

```ts
MapOf<{
    length: number;
    1: string;
}>
```

and the return type of `m.get('length')` is typed as `number`.

The return of `m.get('inexistant')` throw the TypeScript error:

> Argument of type '"inexistant"' is not assignable to parameter of type '1 | "length"

#### If you want to keep the old definition

**This is a minor BC for TS users**, so if you want to keep the old definition, you can declare you Map like this:

```ts
const m = Map<string, string | number>({ length: 3, 1: 'one' });
```

#### If you need to type the Map with a larger definition

You might want to declare a wider definition, you can type your Map like this:

```ts
type MyMapType = {
  length: number;
  1: string | null;
  optionalProperty?: string;
};

const m = Map<MyMapType>({ length: 3, 1: 'one' });
```

#### Are all `Map` methods implemented ?

For now, only `get`, `set` and `getIn` methods are implemented. All other methods will fallback to the basic `Map` definition. Other method definition will be added later, but as some might be really complex, we prefer the progressive enhancement on the most used functions.


mbullington:
Working today on seeing if `getIn` can be improved or implementing the other functions

mbullington:
@jdeniau I spent a lot of time on `getIn` 😆 but finally(!) found a solution using a new TS feature called variadic tuple types: https://github.com/microsoft/TypeScript/pull/39094

mbullington:
Happy to work more on this with you as well, if you have a good idea of what else needs implementing in order to review & merge.

jdeniau:
@mbullington I gave you access on my fork. You can push on the [jd-feat-tsMapObject](https://github.com/jdeniau/immutable-js/tree/jd-feat-tsMapObject) branch.

If you want to continue on the subject, there are still some issues with constructor like `Map(List(List(['a', 'A'])))`, but the worst thing for me is the `ObjectLikeMap` interface, which seems really weird.
The best solution for me would be to have a unique `interface Map` that handle both key/value sequence and also the object constructor, but I didn't handle that without breaking actual interface.
If we keep the two different interfaces, then all other method should be declared and are missing here (especially `set` and `setIn`), but can be declared later.

mbullington:
Was able to push my changes:

Thinking about `ObjectLikeMap`, I think it's hard to integrate them because it's limited to `string | number` (I should add Symbol!), while `Map` can have arbitrary key values.

What's the issue with `Map(List(List(['a', 'A'])))`? Haven't tested it but curious which constructor does it use?

jdeniau:
About `Map(List(List(['a', 'A'])))`, IIRC, It should match `Map(collection: Iterable<[K, V]>)` but it doesn't as `List(['a', 'A'])` does not match the tuple `[K, V]` (as it's a List).

EduardoAraujoB:
hey @jdeniau the `Map` doesn't deep convert objects right? for an example if I do `Map({ a: 'foo', b: { c: 'boo' } })` it will only convert the first object right?

jdeniau:
@EduardoAraujoB exactly. Map only convert one "level". `fromJS ` does convert deeply on the other side.

jdeniau:
> About Map(List(List(['a', 'A']))), IIRC, It should match Map(collection: Iterable<[K, V]>) but it doesn't as List(['a', 'A']) does not match the tuple [K, V] (as it's a List).

I just read [the 4.0.0-rc.11](https://github.com/immutable-js/immutable-js/releases/tag/v4.0.0-rc.11) release message that says:

> [TypeScript] Remove Iterable<T> as tuple from Map constructor types (#1626)
> 
>    Typescript allowed constructing a Map with a list of List instances, assuming each was a key, value pair. While this runtime behavior still works, this type led to more issues than it solved so it has been removed. (Note, this may break previous v4 rcs, but is not a change against v3)

So I think we can ignore the tuple depclaration with a List instance

comatory:
I don't think this is something Immutable should be solving. Like @bdurrer said, map is a map. In TS you wouldn't be able to type JS map neither.

What's stopping you from using `Record`?

EduardoAraujoB:
@jdeniau what about using a own definition instead of `Record` or `Map` ?

it seems for me that the current implementation doesn't match with those structures, so creating a spefic type for that could be a good solution, something like `ImmutableMap`

EduardoAraujoB:
@jdeniau I was thinking about using a different type instead of the typescript `Map` that I saw you extending, because the keys can be more than `string | number | symbol` or whatever problems `Map` could cause

bdurrer:
I think you are mostly trying to solve the "Map as [redux]state" case, where you likely know most keys and the type of the values differ. But Map is a collection too and people (eg like me) use it to hold API data too, where the map holds a specific type, e.g. Map<string, Record>. Hell, I even have Map<Record, Record> types.
Of course I expect all shallow operations to correctly use the stated types on these data collections.

If you can solve problem #1 without breaking simple types of #2 be my guest, that would be cool! However as already mentioned, if the key type is just wrong, then either TS needs improvements or Immutable (or this task is unsolvable). The compiler incorrectly telling me "no you can't, that object is a number" is worse than using _any_

bdurrer:
For the record, I fully agree that it certainly is a common use-case and we should try to solve it. If the "standard collection" and "state construct" cannot be unified in a single type, maybe we want to export Map a second time under a different name and with another type? Although that would probably lead to more problems than it solves...

We could try to ease the pain by letting the callers of methods define what they expect? E.g. like `getIn<T>(...)` That wouldn't be real type safety, but at least make the fictional casting easier

EduardoAraujoB:
@bdurrer Remember that type definitions only affects development environment, but in runtime, it would not cause a new error, since the changes in this PR doesn't affect the current runtime implementation of `Map`, only the type definitions.

Even in cases where "the key is not knowned" it'll works at the same way, with the difference of alerting dangerous things that you're doing in your code, but as always, you can skip that warning and compile and it'll work at the same way

bdurrer:
Huh I wrote an answer but apparently didn't submit, oops 😀 

> @bdurrer do you have a reproducible example of what might break with this implementation?
Nope, since both variants kinda coexist, I am fine with what the PR offers. 
Althrough I still think `get` shouldn't throw on unset/undefined keys, that's totally expected behaviour of a collection type. If your code follwing the `get` can't deal with undefined it's your (the programmer) fault, but nothing we should enforce. That would lead to many conflicts with `merge`, `fromJS`, etc, which can add keys that TypeScript would have a hard time detecting.


> with the difference of alerting dangerous things that you're doing in your code, but as always, you can skip that warning and compile and it'll work at the same way
yeah well, all the teams I know are kinda strict with warnings, put too many not-entirely-correct warnings in it and they'll hate your framework for no reason ;)

bdurrer (changes requested):
Sorry, I forgot the side-job: Please update the CHANGELOG.md!
It needs a new "unreleased" section with this change being the first entry

jdeniau:
@acusti I took your name that is way better.

@bdurrer I updated the CHANGELOG. I also understood your comments as mine was only focusing on the error for undefined key, and I did not pointed out the type resolution for each key (but I think that you understood it later).

@leebyron Do you have an opinion on this one before merge ?

bdurrer:
IMO the changelog should be short so it can be read quickly, not offer extensive explanations. But maybe that would make a great wiki page that it could link to?

jdeniau:
@bdurrer I will open a discussion or a wiki page once this is merged to keep the CHANGELOG light

mbullington:
Thanks all and @jdeniau for working on this!

I can understand the argument that `Map` isn't the place for these and wanting to keep the typing aligned with the ES6 Map, so it's not without a good discussion.

I think the main goal of this is to enable strong typing for the "immutable POJOs," which in a lot of cases due to `fromJS` is how people use Immutable. There's no ES6 Map equivalent for that.

Right now it's hard to justify using the Immutable typing with `fromJS` since it'll sometimes get in the way, in the sense of "I know this is a `string` but I have to check because TS says it's a `string | number`" So this PR makes a world of difference there.

For people like @bdurrer 's use case I'd hate to flip this around, trading one alienated use case for another. If we can strike a balance (which I hope we've done?) both groups will be able to work productively.

---

I've been playing around with an internal version for Mapbox Studio that recursively types `fromJS` and `toJS`, as well as adding support for `update`.

I'd love some feedback on if these should be added or if the from/toJS logic is even correct: 

https://gist.github.com/mbullington/631dd21f910e594b29fb97dd6f3e62e8

__tests__/Map.ts
```diff
@@ -97,6 +100,7 @@ describe('Map', () => {
   it('accepts non-collection array-like objects as keyed collections', () => {
     const m = Map({ length: 3, 1: 'one' });
     expect(m.get('length')).toBe(3);
+    // @ts-expect-error -- type error, but the API is tolerante
```
leebyron, line 103: This should not be a type error. Similarly, we should ensure that `m.get(1)` is well typed to return NSV (or be a type error if that arg is omitted)

I filed an issue about this a while back and it was closed as "Working as intended" despite being clearly incorrect. 🤷  - https://github.com/microsoft/TypeScript/issues/45170#issuecomment-909751159

A workaround type to fix this looks like:

```ts
type ObjectKeys<T> = { [K in keyof T]: K extends number ? `${K}` : K }[keyof T]
```

__tests__/Map.ts
```diff
@@ -60,8 +60,11 @@ describe('Map', () => {
     const l = List([List(['a', 'A']), List(['b', 'B']), List(['c', 'C'])]);
     const m = Map(l);
     expect(m.size).toBe(3);
+    // @ts-expect-error -- Not supported by typescript since 4.0.0 https://github.com/immutable-js/immutable-js/pull/1626
     expect(m.get('a')).toBe('A');
+    // @ts-expect-error -- Not supported by typescript since 4.0.0 https://github.com/immutable-js/immutable-js/pull/1626
     expect(m.get('b')).toBe('B');
+    // @ts-expect-error -- Not supported by typescript since 4.0.0 https://github.com/immutable-js/immutable-js/pull/1626
     expect(m.get('c')).toBe('C');
```
leebyron, line 68: I'm confused - is this not currently failing? Why does this PR cause this to start producing errors?

type-definitions/ts-tests/map.ts
```diff
@@ -61,6 +71,31 @@ import { Map, List } from 'immutable';
...
+  // $ExpectType boolean
+  Map({ a: 4, b: true }).get('b');
+
+  // $ExpectType boolean
+  Map({ a: Map({ b: true }) })
+    .get('a')
+    .get('b');
+
+  // $ExpectError
+  Map({ a: 4 }).get('b');
```
leebyron, line 87: Can you add `.get('b', undefined)` to test the alternate NSV definition?

type-definitions/ts-tests/map.ts
```diff
@@ -61,6 +71,31 @@ import { Map, List } from 'immutable';
...
+}
```
leebyron, line 88: Can you add tests for the number and symbol keys as well?

type-definitions/ts-tests/map.ts
```diff
@@ -61,6 +71,31 @@ import { Map, List } from 'immutable';
...
+{
+  // Minimum TypeScript Version: 4.1
+  // #getIn
+
+  // $ExpectType number
+  Map({ a: 4, b: true }).getIn(['a']);
+
+  // $ExpectType number
+  Map({ a: Map({ b: Map({ c: Map({ d: 4 }) }) }) }).getIn(['a', 'b', 'c', 'd']);
 }
```
leebyron, line 99: Can you add getIn tests where nested Map are constructed by other means (eg not only `ObjectFromMap` type?) as well as having `List` within?

type-definitions/ts-tests/map.ts
```diff
@@ -80,6 +115,28 @@ import { Map, List } from 'immutable';
 
   // $ExpectType Map<number, string | number>
   Map<number, number | string>().set(0, 'a');
+
+  // $ExpectError
+  Map({ a: 1 }).set('b', 'b');
```
leebyron, line 120: I'm surprised by this, I'd expect either `MapFromObject<{ a: 1, b: 'b' }>` or `Map<string, number | string>` as a result?

type-definitions/immutable.d.ts
```diff
@@ -760,9 +760,93 @@ declare namespace Immutable {
...
+    remove<K extends keyof R>(
+      key: K
+    ): Extract<R[K], undefined> extends never ? never : this;
+  }
+
+  // Loosely based off of this work.
+  // https://github.com/immutable-js/immutable-js/issues/1462#issuecomment-584123268
+
+  /** @ignore */
+  type GetMapType<S> = S extends MapFromObject<infer T> ? T : S;
```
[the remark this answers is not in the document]
EduardoAraujoB: I think that it will just return the type `Map` or `List`, am I correct? @jdeniau

type-definitions/immutable.d.ts
```diff
@@ -760,9 +760,93 @@ declare namespace Immutable {
...
+    get<K extends keyof R>(key: K, notSetValue?: unknown): R[K];
+    get<NSV>(key: string, notSetValue: NSV): NSV;
+
+    // https://github.com/microsoft/TypeScript/pull/39094
+    getIn<P extends ReadonlyArray<string | number | symbol>>(
+      searchKeyPath: [...P],
+      notSetValue?: unknown
+    ): RetrievePath<R, P>;
+
+    set<K extends keyof R>(key: K, value: R[K]): this;
```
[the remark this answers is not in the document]
EduardoAraujoB: if `key` is not in `keyof R` immutable will return a `undefined` right?

type-definitions/immutable.d.ts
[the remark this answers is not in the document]
jdeniau: On something that is not a `MapFromObject`, it will return the type directly. I do not really know in the end what will happen though. Maybe @mbullington who coded this part ?

I thing that:
```ts
MapFromObject<{ m: Map<string, number>; l: List<number>; }>
GetMapType<m['m']> // Map<string, number>
GetMapType<m['l']> // List<number>
```

Reading this particular line, the function only extract the initial object of a MapFromObject, that's all.

So in a chain:

```ts
MapFromObject<{
  m: Map<string, List<number>>
}>
```
Calling `RetrievePath<R, ['m', 'some-string', number]>`  _should_ return `number`

type-definitions/ts-tests/map.ts
jdeniau: See comment above about the consistency with plain objects in TS

type-definitions/immutable.d.ts
```diff
@@ -760,9 +760,93 @@ declare namespace Immutable {
...
+     *
+     * Note: it is possible a key may be associated with an `undefined` value,
+     * so if `notSetValue` is not provided and this method returns `undefined`,
+     * that does not guarantee the key was not found.
+     */
+    get<K extends keyof R>(key: K, notSetValue?: unknown): R[K];
+    get<NSV>(key: string, notSetValue: NSV): NSV;
+
+    // https://github.com/microsoft/TypeScript/pull/39094
+    getIn<P extends ReadonlyArray<string | number | symbol>>(
```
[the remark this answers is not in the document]
jdeniau: I do not know why, but if I do that, calling `getIn(['a', 'b', 'c', 'd'])` is not handled and fallback to be a `never` type

type-definitions/ts-tests/map.ts
jdeniau: Unfortunatly, it's only working for `MapFromObject` for now.

mbullington:
I took a stab at making sure `merge` & co work properly. 

@leebyron Thanks for looking and identifying these issues with `getIn`.

In honesty I changing `never` to `unknown` would be best here, and keeping the current object-all-the-way-down path as helpful but not stopping you from using `getIn` otherwise.

My reasoning is we're dealing with both string literal types like `"car"` and then generic types like `string`. With the `[...T]` syntax (https://github.com/microsoft/TypeScript/pull/39094) I had to make it `ReadonlyArray<string | number | symbol>` to get the path type looking like `["a", "b", "c"]` using literals. Otherwise it would use `[string, string, string]`, so when we check if `string extends "a" | "b" | "c"`, it doesn't and returns never.

This could present issues with typing `getIn` for non-object-all-the-way-down structures.

type-definitions/immutable.d.ts
```diff
@@ -760,9 +760,93 @@ declare namespace Immutable {
...
+      searchKeyPath: [...P],
+      notSetValue?: unknown
+    ): RetrievePath<R, P>;
+
+    set<K extends keyof R>(key: K, value: R[K]): this;
+
+    update(updater: (value: this) => this): this;
+    update<K extends keyof R>(key: K, updater: (value: R[K]) => R[K]): this;
+    update<K extends keyof R, NSV>(
```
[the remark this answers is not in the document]
mbullington: Tried a fix for this, thanks!

eesaber:
Hi @mbullington 

You can get the type of the given key path of a plain JS object like
```ts
type Arr = readonly unknown[];

type GetInType<T, KeyPath extends Arr> =
  KeyPath extends [infer U]
    ? U extends keyof T 
      ? T[U]
      : 'fail 1'
    : KeyPath extends [infer K, ...infer Rest]
      ? K extends keyof T 
        ? GetInType<T[K], Rest>
        : 'fail 2'
      : 'fail 3'

// ---
const obj = {
  foo: {
    list: [1,2,3],
    listObj: [{name: "alex"},{name: "sam"}],
  },
  num: 123,
};

type test = GetInType<typeof obj, ['foo', 'list']>;
```
So, the type of `test` is `number[]`

[Playground Link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggTnKBeKcIEMAmB7AdgGxCgFdcBrXbAd1wG0BdAbgChnRIoBxCYASVwAq4CAB4BAGigBpCCAAK6YAAsoEAB7AIuTAGdYCAHzJmUabIXLVGrbqi0AlrgBmERAFV6J01AD8UN1aa2npksthOUAJQXt6+kbQeMd4AXFAA5E7o9vhQAIxpSaky8ooq6kG2Ds6u0pIAdA2OLogAShA6wJ6xpn5SgTYhYRFRSbF+3HyCwmK0UvSSbR0GoynpmdlQAEwF3VCpGVk5AMwFzADGeB1Q2ABGAFbIUADeXk7Y2Kkvsfj2Ham0uXEm3ER3mSR+HQA8vd-k9cOgALYQVIAInQ+HUKIAvuI4YjkVAUTpEdiwaYcV5cMQEalcpsjuJmFiWGxhFBNFcUBN+EJICJ2BBwtd7pJaBl3mlJGkIcA0vQDCwBez2sBco9uVM+QKhbc7qLxdhJekZdC7nKFayOBzgJt1TwedNtRFdfq3oapSb7kbcvKWUrKKEiFz7ZrRE7hXq7Gk6SdfZboAHZGrg5NeWHhDqRVG3Ubpb9gKajZs46wgA)

For ImmutableJS, I guess we could replace `T[U]` by something like `T.get(U)`.

jbaca303:
What's the status on this? Looks like it would be highly beneficial for a project my team is working on

jdeniau:
@jbaca303 it's a work in progress but it's really complex to make this work properly.

Don't expect it to be done in the next few weeks, but if you would like to try it on your project, every return would be appreciable.

jbaca303:
> @jbaca303 it's a work in progress but it's really complex to make this work properly.
> 
> Don't expect it to be done in the next few weeks, but if you would like to try it on your project, every return would be appreciable.

Thanks for the quick update. Totally understand that this complex. Thank you guys for working on this, really appreciate it!

jdeniau:
I am using this for nearly 2 years now at https://github.com/mapado internally. 

I think it is time to go public now. I will release a RC on a 5.0 version to have some community return on this. 

Thanks you all for your patience.

PR #1849, opened 2021-07-13
test: replace make-synchronous with synckit which is much faster

JounQin:
[`synckit`](https://github.com/rx-ts/synckit) uses `worker_threads` under the hood while `make-synchronous` uses `child_process` which is much slower

See [benchmark](https://github.com/rx-ts/synckit/blob/perf/make-synchronous/benchmarks/benchmark.txt) for details


resources/jestPreprocessor.js
```diff
@@ -66,6 +25,8 @@ module.exports = {
       return transpileTypeScript(src, path);
     }
 
-    return transpileJavaScript(src, path);
+    // Need to make this sync by calling `synckit.createSyncFn`
+    // while https://github.com/facebook/jest/issues/9504 is not resolved for `cjs`
+    return createSyncFn(require.resolve('./transpile-javascript'))(src, path);
```
JounQin, line 30: `createSyncFn` is called lazily and cached in `synckit` itself.

jdeniau:
@JounQin The perf improvement seems great, but as https://github.com/facebook/jest/issues/9504 is resolved, can't we remove this part ? 🤔

JounQin:
@jdeniau Unfortunately not, `processAsync` is only available for all files as esm, I tried locally and failed.

jdeniau:
What do you think @immutable-js/maintainers ? I have really no opinion about this change.

PR #1850, opened 2021-07-13
Issue with IE11 and missing Symbol.iterator

jdeniau:
Fixes https://github.com/immutable-js/immutable-js/issues/1711


bdurrer (approved):
Simple and nicely documented 👍

PR #1854, opened 2021-07-16
Simplify typescript definition files

leebyron:
This removes the build step which converts an "ambient" type definition file into a "non-ambient" one. Since our local tests all run against the source file, this ensures the tested file is exactly the same as the exported one, and updates the exports to support all "UMD" use cases, both ambient (sourced file with a `/// <reference>` and non-ambient module exports.

As a result, we no longer need to explicity mark each type within the exported module

I'm ~90% sure this is the correct way to set up the module and UMD exports, but would be very happy to have a second set of eyes on this.

This also changes the upper-case source files to lower-case so they can be directly copied during build.

Tested by creating a new vanilla typescript project and importing the built npm directory and tested a bunch of imports of Immutable to ensure the types were being correctly referenced.


jdeniau:
I really love the change : I did not understand the complexity of the ambiant / nonambiant files.

It does work great with `import`.

I tried to test with require (which I do not use really often), the definition does not seems to work BUT it did not work with the previous file either. It may be a ts configuration issue on my side.

Just to be sure, here is the test I made:

```ts
import { List as ListImport } from 'immutable';
import ImmutableImport from 'immutable';
const { List: ListRequire } = require('immutable');
const ImmutableRequire = require('immutable');

const ii = ImmutableImport.List(['a']); // ListImport<string>
const li = ListImport(['a']); // ListImport<string>
const lr = ListRequire(['a']); // any
const ir = ImmutableRequire.List(['a']); // any
```

jdeniau (approved):

leebyron:
I think there's a TS config you need to set up to properly support the CJS `require` semantics, but the `export = Immutable` should cover them all dependent on how the config is set up it will interpret this as CJS or a default+ns export as CJS are typically interop'ed to by webpack and TS. So the combo of `exports =` and `export as namespace` covers all UMD: CJS, Modules, and global.

https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

PR #1858, opened 2021-07-19
accept plain JS Map and Set in fromJS

### LINKED ISSUE #1723: Map and Set are not converted to Immutable.Map and Immutable.Set via Immutable.fromJS()
### What happened

There is no way to create Immutable.Map and Immutable.Set from Map and Set via Immutable.fromJS()

### How to reproduce

```js
const set = new Set([1, 2, 3]);
const immutableSet = Immutable.fromJS(set);

console.log(set === immutableSet) // true
``` 

And the same with Map

jdeniau:
Fixes #1723

### DIFF
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -1,11 +1,25 @@
-import { is, List, Map, Range, Record, Seq } from 'immutable';
+import { is, List, Map as ImmutableMap, Range, Record, Seq } from 'immutable';
 
 import * as jasmineCheck from 'jasmine-check';
 jasmineCheck.install();
 
 describe('Map', () => {
   it('converts from object', () => {
-    const m = Map({ a: 'A', b: 'B', c: 'C' });
+    const m = ImmutableMap({ a: 'A', b: 'B', c: 'C' });
+    expect(m.size).toBe(3);
+    expect(m.get('a')).toBe('A');
+    expect(m.get('b')).toBe('B');
+    expect(m.get('c')).toBe('C');
+  });
+
+  it('converts from Map', () => {
+    const map = new Map();
+    map.set('a', 'A');
+    map.set('b', 'B');
+    map.set('c', 'C');
+
+    const m = ImmutableMap(map);
+    expect(ImmutableMap.isMap(m)).toBe(true);
     expect(m.size).toBe(3);
     expect(m.get('a')).toBe('A');
     expect(m.get('b')).toBe('B');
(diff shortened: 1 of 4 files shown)


jdeniau:
ping @conartist6 as you commented on #1723

__tests__/Map.ts
```diff
@@ -1,27 +1,41 @@
-import { is, List, Map, Range, Record, Seq } from 'immutable';
+import { is, List, Map as ImmutableMap, Range, Record, Seq } from 'immutable';
```
leebyron, line 1: Instead of having to convert this whole test file to use an alias, how do you feel about just making a new test file for Map to IMap?

src/fromJS.js
```diff
@@ -13,12 +14,25 @@ export function fromJS(value, converter) {
   );
 }
 
+function getToSeq(value) {
+  if (Array.isArray(value)) {
+    return IndexedSeq;
+  }
+
+  if (value instanceof Set) {
+    return ImmutableSet;
```
leebyron, line 23: This should be determining a kind of Seq to use

```suggestion
    return SetSeq;
```

src/fromJS.js
```diff
@@ -13,12 +14,25 @@ export function fromJS(value, converter) {
   );
 }
 
+function getToSeq(value) {
+  if (Array.isArray(value)) {
+    return IndexedSeq;
+  }
+
+  if (value instanceof Set) {
```
leebyron, line 22: To disambiguate, you might want to specifically reference the global Set class

```suggestion
  if (value instanceof global.Set) {
```

src/fromJS.js
```diff
@@ -41,5 +55,9 @@ function fromJSWith(stack, converter, value, key, keyPath, parentValue) {
 }
 
 function defaultConverter(k, v) {
+  if (ImmutableSet.isSet(v)) {
+    return v;
+  }
+
   return isKeyed(v) ? v.toMap() : v.toList();
```
leebyron, line 62:
```suggestion
  return isKeyed(v) ? v.toMap() : isSet(v) ? v.toSet() : v.toList();
```

leebyron (changes requested):
This doesn't handle `WeakSet` and `WeakMap`, and other collections.

Another way you might handle determining which kind of Seq to use is to look at the iterator methods for Iterable collections (all modern JS collections). eg if `value[ITERATOR_SYMBOL] === value.entries` then it's a Keyed collection. If `value[ITERATOR_SYMBOL] === value.keys` it's a Set collection. Otherwise probably safe to assume anything iterable is an Indexed collection (though it's often the case that `value[ITERATOR_SYMBOL] === value.values` - also careful since that's also true for `Set`)

leebyron:
Trying out an alternative approach in #1865 that has a bit of a broader scope, curious for your thoughts

__tests__/Map.ts
jdeniau: I did not knew how to call global constructor, I think the best way is to call global.Map as you did in another PR

jdeniau:
Your implementation is way better, and more generic. Closing in favor of #1865

PR #1859, opened 2021-07-19
Accept Symbol as Map key.

### LINKED ISSUE #1731: Map constructor ignores symbols
`Map({ [MySymbol]: 5 })` generates an empty map despite `Map([[MySymbol, 5]])` generating a single item `Map`. I imagine `Map` uses `Object.keys` on object argument instead of a combination of `Object.keys` and `Object.getOwnPropertySymbols`.

jdeniau:
Fixes #1731

### DIFF
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -1,4 +1,4 @@
-import { is, List, Map, Range, Record, Seq } from 'immutable';
+import { fromJS, is, List, Map, Range, Record, Seq } from 'immutable';
 
 import * as jasmineCheck from 'jasmine-check';
 jasmineCheck.install();
@@ -465,6 +465,25 @@ describe('Map', () => {
     expect(m.get(a)).toBe('a');
     expect(m.get(b)).toBe('b');
     expect(m.get(c)).toBe('c');
+  });
+
+  it('supports Symbols as object constructor keys', () => {
+    const a = Symbol.for('a');
+    const b = Symbol('b');
+    const c = Symbol('c');
+    const m = Map({
+      [a]: 'a',
+      [b]: 'b',
+      [c]: 'c',
+    });
+    expect(m.size).toBe(3);
+    expect(m.get(a)).toBe('a');
+    expect(m.get(b)).toBe('b');
+    expect(m.get(c)).toBe('c');
+
+    const m2 = fromJS({ [a]: 'a' });
+    expect(m2.size).toBe(1);
+    expect(m2.get(a)).toBe('a');
   });
 
   it('Symbol keys are unique', () => {
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -472,7 +472,6 @@ describe('Map', () => {
     const b = Symbol('b');
     const c = Symbol('c');
     const m = Map<symbol, string>({
-      // @ts-expect-error -- typescript definition does not handle symbol for now
       [a]: 'a',
       [b]: 'b',
       [c]: 'c',
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -82,14 +82,14 @@ describe('Map', () => {
 
   it('does not accept strings (collection, but scalar)', () => {
     expect(() => {
-      // @ts-expect-error
+      // @ts-expect-error -- constructor does not accept strings, this is expected to throw
       Map('abc');
     }).toThrow();
   });
 
   it('does not accept non-entries array', () => {
     expect(() => {
-      // @ts-expect-error
+      // @ts-expect-error -- not an array of entries, this is expected to throw
       Map([1, 2, 3]);
     }).toThrow('Expected [K, V] tuple: 1');
   });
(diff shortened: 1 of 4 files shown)


type-definitions/Immutable.d.ts
```diff
@@ -761,7 +761,9 @@ declare namespace Immutable {
    */
   export function Map<K, V>(collection?: Iterable<[K, V]>): Map<K, V>;
   export function Map<V>(obj: { [key: string]: V }): Map<string, V>;
-  export function Map<K extends string, V>(obj: { [P in K]?: V }): Map<K, V>;
+  export function Map<K extends string | number, V>(
```
leebyron, line 764: `string | symbol` ?

__tests__/Map.ts
```diff
@@ -447,6 +447,26 @@ describe('Map', () => {
     expect(m.get(c)).toBe('c');
   });
 
+  it('supports Symbols as object constructor keys', () => {
+    const a = Symbol.for('a');
+    const b = Symbol('b');
+    const c = Symbol('c');
+    const m = Map<symbol, string>({
+      // @ts-expect-error -- typescript definition does not handle symbol for now
```
leebyron, line 455: This should work - ideally you can add something to the ts-tests as well to assert this

https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMAeAaXhAA9DVgj4SYtUBzeAHyYE8BbAEY4IAGngA1AHwAKHAIBWALngBveAG0ACvFbxyAXQD8y8fAC+ASmUBZKAAcKYqQG4AsACgPYPCSbwAvPAAyvxCENIA5DgA1hEWXj4YBAHBocKRqDhJMXEeHgD0+cEAFjjIEMA6qIggMDb2pBFQEVwEPHYgOIhMYqjIgrWSCai+cETlSYGEJNJqUMoATGLqRPrKAMzm8Z7uYxMAdGwgGJHN23sQGIfH0kTbhfAAAhhEALQ0HZjvMDA4MB4XK5HE4RAS5dwPZ5vD7gDDfX7-XbEA7A6QYbZAA

type-definitions/Immutable.d.ts
jdeniau: I think it should probably be `string | number | symbol` as number can be an object index. But in that case `Map('abc')` and `Map([1,2,3])` do not throw an error anymore. 

The result type is something like that : 
> const abc: Map<number | keyof String, string | number | (() => string) | ((pos: number) => string) | ((index: number) => number) | ((...strings: string[]) => string) | ((searchString: string, position?: number | undefined) => number) | ... 36 more ... | (() => IterableIterator<...>)>

Example here : https://www.typescriptlang.org/play?ssl=1&ssc=51&pln=1&pc=57#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMAeAaXhAA9DVgj4SYtUBzeAHyYE8BbAEY4IXeKmSCQMADTwAagD4AFDgEArAFzwA3vADaABXit45ALoB+LXPgBfAJRaAslAAOFWYoDcAWABQ-mB4JEzwALzwAMr8QhBKAOQ4ANbx9oHBGAThUTHCCag4mcmp-v4A9GVRABY4yBDAxqiIUs5upPFQ8aIYPK4gOIhMsuKSMArpqCFwRHWZEYQkSrpQWgBMsnpEZloAzLIALFoArHZpAX7TswB0bCAYCZ1nlxAYN3dKRGcV8AACGEQAWhofUwQJgMBwMH8z1et3u8QEJT83z+gOB4AwYIhUIuxGucKUGDOpWRlVRQOoIMxUmxExCUAEYGyC3hDLASPKZP+FKpWMhdKKqBAABUAO44YVVOAIebEe56ACMsnW8B2ZjOQA

leebyron: I'm not sure if that's a bug in Typescript, or a quirk they're forced to endure because of how JS Array exotics are weird, but in that play link I see `const result = test({ a: 2, [s]: 3, 4: 5 })` inferred to return `Map<'a' | typeof s | 4, number>` when I would have expected `Map<'a' | typeof s | '4', number>` (note the string `"4"` not the number `4`)

It also reports that `result.get(4)` is valid but `result.get("4")` is a type error, but you can try this yourself:

```ts
const result = Immutable.Map({ a: 2, 4: 5 })
result.get(4) === undefined
result.get("4") === 5
```

Javascript Object property names are *always* strings

leebyron: This is by far not the first time I've been bitten by this or have seen someone else bitten by this. I searched through TS's issue log and saw some things roughly related but none calling the direct inference issue out, so I opened one:

https://github.com/microsoft/TypeScript/issues/45170

jdeniau:
Just a reminder for me to look for https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-rc/#symbol-template-signatures as TS4.4 does change a thing about symbol.

PR #1865, opened 2021-07-23
Generalize fromJS() and Seq() to support Sets

### LINKED ISSUE #1723: Map and Set are not converted to Immutable.Map and Immutable.Set via Immutable.fromJS()

leebyron:
- When calling `Seq(value)` on an iterable value, use the convention of the default iterator method being equal to one of the prototype methods `value.entries()` and `value.keys()` as a clear indication of Keyed or Set style collections.
  - Introduces `isEntriesIterable()` and `isKeysIterable()` to capture these patterns.
  - Uses them within `Seq()` to produce an entries sequence or set sequence.
- Use the general `Seq()` in the implementation of `fromJS()`, rather than repeating the logic of determining which Seq conversion to apply.
  - Replaces this logic with constraints on which values can be converted
  - Introduces allowing iterable objects
  - Introduces allowing array-likes
- Add `toSet()` as a possible return from the default converter of `fromJS()` in the case the generated Seq is a SetSeq.

Fixes #1723

### DIFF
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -12,6 +12,21 @@ describe('Map', () => {
     expect(m.get('c')).toBe('C');
   });
 
+  it('converts from JS (global) Map', () => {
+    const m = Map(
+      new global.Map([
+        ['a', 'A'],
+        ['b', 'B'],
+        ['c', 'C'],
+      ])
+    );
+    expect(Map.isMap(m)).toBe(true);
+    expect(m.size).toBe(3);
+    expect(m.get('a')).toBe('A');
+    expect(m.get('b')).toBe('B');
+    expect(m.get('c')).toBe('C');
+  });
+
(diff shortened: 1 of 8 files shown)


jdeniau (approved):

amr-labib:
Hi, when do you expect to release this change?

Thanks
Labib

jdeniau:
Hi @amr-labib. There will probably be a release in september.

jdeniau:
@amr-labib the `v4.0.0-rc.15` has been released !

PR #1857, opened 2021-07-16
keep sidebar fixed + merge top and bottom sidebar

jdeniau:
I tried to improve the documentation style a little.

Preview is available here : https://jdeniau-immutable-js.vercel.app/docs/latest@main/

PR #1868, opened 2021-07-23
keep sidebar fixed + merge top and bottom sidebar

jdeniau:
Replaces #1857 (some issue with git push force, sorry)


jdeniau:
@leebyron I removed the border : it's cleaner this way. 

I kept the collapse functionality and the arrow, mainly inspired by [the react documentation](https://reactjs.org/docs/getting-started.html) :
- the collapse functionality seems really useful to me to navigate between two main items (I am reading the `Map.set` documentation, and then I want to go to `List` documentation, it's more easy to close all `Map` group than to scroll all the way up or down as there is a huge list of items in `Map`
- the arrow make more understandable where we are.

leebyron:
Reacts docs are a good source of inspiration, but I'd have suggested some design improvements to them, and also there are some important differences we have here.

IMHO using a bolded font alone would be more clear of where you are than an inverted arrow glyph. Collapsing the section for React makes less sense where each section is not so big, but I see why it's useful for us where our sidebar can be quite large. However right now once a section is collapsed the bold font goes away - which is still trying to show you where you are.

React also has only a few sections, so the repetition is less stark there, but our docs have dozens so it's more distracting. Also, more than half are just functions, which don't have anything to expand, so it's a bit confusing to see an expand/collapse indicator for something that can't be expanded or collapsed.

My suggestion would be to only show the expand/collapse arrow for the particular page currently being viewed (and there are members to expand and collapse). That would immediately solve the visual noise problem and actually draw attention to the expand/collapse functionality for that section since there's only one glyph to interact with instead of dozens.

Some minor visual feedback, the member elements within each section should not have additional left-margin since that makes it harder to visually track the edge of the sidebar, and if that had less top-margin then it would be more clear that the members are associated with the page section header.

jdeniau:
Changes are up and running : https://immutable-js-git-feat-docsidebar-jdeniau.vercel.app/docs/latest@main/

PR #1869, opened 2021-07-29
Wrap immutable.d.ts type declarations in a module to avoid a global declaration

vilemj-Viclick:
## The problem:
When using the type-definitions, you provide right now, this is possible and does not get detected not even by running `tsc`:
image
Notice, there's no import statement. When the code is run, it errors out, unless ProvidePlugin is used in WebPack to add the necessary import.

This is due to the global nature of the namespace declaration.

## The solution:
Take the namespace declaration out of the global.
With the type definitions proposed in the pull request this happens:
image
Despite the form of import being proposed when imported the prefered way, it works just fine.
image

Tests pass as far as they can on a windows machine. Your npm scripts for `build:prepare` are not windows compatible as the shebang notation has no effect on windows. But maybe adding the shebang to the package.json directly, resulting in `"build:prepare": "sh -e ./resources/prepare-dist.sh"` might help. Most unix-based devices have alias for "sh" to the preferred shell. Be it `ksh` on a mac or `bash` on linux machines.

### DIFF
--- a/type-definitions/immutable.d.ts
+++ b/type-definitions/immutable.d.ts
@@ -89,2850 +89,497 @@
  * [Flow]: https://flowtype.org/
  * [Iterable]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols
  */
-
-declare namespace Immutable {
-  /**
-   * Lists are ordered indexed dense collections, much like a JavaScript
-   * Array.
-   *
-   * Lists are immutable and fully persistent with O(log32 N) gets and sets,
-   * and O(1) push and pop.
-   *
-   * Lists implement Deque, with efficient addition and removal from both the
-   * end (`push`, `pop`) and beginning (`unshift`, `shift`).
-   *
-   * Unlike a JavaScript Array, there is no distinction between an
-   * "unset" index and an index set to `undefined`. `List#forEach` visits all
-   * indices from 0 to size, regardless of whether they were explicitly defined.
-   */
-  namespace List {
+declare module 'immutable' {
+  namespace Immutable {
     /**
-     * True if the provided value is a List
+     * Lists are ordered indexed dense collections, much like a JavaScript
+     * Array.
      *
-     * <!-- runkit:activate -->
-     * ```js
-     * const { List } = require('immutable');
-     * List.isList([]); // false
-     * List.isList(List()); // true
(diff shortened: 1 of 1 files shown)


jdeniau:
Hi !

Can you check that #1854 does not fix this issue ? It is not releases for now but @leebyron made some work that seems to be related.

Thanks

vilemj-Viclick:
Hi. Thanks for looking at this.

There is still the global declaration. This causes typescript to be fine with it:
image
Just the IDE hinting at a missing import. But it compiles and does not run.
But it's definitely better than the released version. I just sought to make it another notch better.

leebyron:
The current definition file having a global namespace export is intentional since Immutable is a UMD package, and can be used as a global Immutable variable.

The alternative would be to host two separate definition files for the two different use cases, but this single file pattern is fairly common across definitely-typed. For example the React definitions use this same pattern.

vilemj-Viclick:
Alright then. I did not study the problem of module type definitions very much. I see you did a good job on the types and surely have your reasons.
All I know is, `prop-types` for example does not have this problem. I did not check React. But it might be, that `prop-types` is not a UMD package.

I suspected it was done this way for a reason. Well anyway... Question is, how big of a crowd is the Typescript + UMD crowd?
I'm not saying it should not be UMD on JS level. I'm questioning the necessity to reflect the global UMD variable in typescript type definitions as I wouldn't even know how to use typescript with UMD. Babel/Webpack uses ESM. But you @leebyron seem to be on top of things. You may know how much this is required.

Well, I suspect this is as much as I can contribute. Whatever you decide, I don't think further discussion (me pushing my way) would be productive. Thanks for considering my proposal.

And thanks for taking care of the package! You, guys (or whatever, one never knows), rock!

bdurrer:
As mentioned in slack, maybe it is worth testing the same-name pattern instead of the "types" definitions in the package.json.
That would allow to have multiple definitions for different entry points.

Place separate definitions [name].d.ts besides the umd and es6 scripts and it should be automatically picked up. Maybe it only works when working with folders (e.g. /dist/es6/ and /dist/umd/) or when directly referencing it (e.g. `import { List } from 'immutable/umd'`). I haven't tested it, but might be worth checking!

[@angular/material](https://github.com/angular/components) uses this pattern

PR #1870, opened 2021-07-31
Record Flow type improvements

### LINKED ISSUE #1722: Record definitions are sloppy
Flow version: 0.97.0 / immutable 4.0.0-rc12

### What happened
When defining records, you have to do all the typecasting because `Record()` has a incorrect type definition.

### How to reproduce
```
type Product = { address: string };
type App = { name: string };
const AppDefaults: App = { name: '' };

const AppRecordImplicit = Record(AppDefaults);        // class RecordInstance<T: any = any>
const AppRecordExplicit = Record<App>({ name: '' });  // class RecordInstance<T: any = any>
const AppRecordBorked: RecordFactory<App> =           // RecordFactory<App>
  Record<Product>({ address: '' });

const appImplicit = AppRecordImplicit({ name: 'i' }); // RecordOf<empty>
const appExplicit = AppRecordExplicit({ name: 'e' }); // RecordOf<empty>
const appBorked = AppRecordBorked({ name: 'b' });     // RecordOf<App>
```

In the above code, none of the ways of creating records work. The only way that actually works is to say:
```
const AppRecordExtraExplicit: RecordFactory<App> = Record<App>(AppDefaults);
```

Ideally, I'd expect either of the first two to work (`AppRecordImplicit` or `AppRecordExplicit`). However, they create `RecordOf<empty>` since they are not typed themselves (they receive `class RecordInstance<T: any>`).
The third call illustrates how that is a bit crazy, since you can successfully cast a record factory of one type to another (which leads you to be able to make records of the wrong type entirely).

Only by using `AppRecordExtraExplicit` -- and using it correctly, since it is the same form as `AppRecordBorked` -- can you get a `RecordFactory<App>` and produce actual `RecordOf<App>`.

mwiencek:
Fixes #1722

I moved the `Values` type parameter to `Record<Values>`, so calling `Record(foo)` can infer `Values` from `foo` and now returns `RecordFactory<Values>`.

This is a breaking change, though I believe it provides much better type checking for typical uses (and fixes at least one TODO).  Unfortunately, this PR introduces issues when calling `RecordInstance` methods from a subclass (which I'm not sure how to fix), but that probably isn't super common, so the tradeoff might be worth it.

### DIFF
--- a/type-definitions/flow-tests/record.js
+++ b/type-definitions/flow-tests/record.js
@@ -14,11 +14,9 @@ const Point3: RecordFactory<{ x: number, y: number, z: number }> = Record({
 type TGeoPoint = { lat: ?number, lon: ?number };
 const GeoPoint: RecordFactory<TGeoPoint> = Record({ lat: null, lon: null });
 
-// TODO: this should be FlowExpectedError - 'abc' is not a number
-// However, due to support for the brittle support for subclassing, Flow
-// cannot also type check default values in this position.
 const PointWhoops: RecordFactory<{ x: number, y: number }> = Record({
   x: 0,
+  // $FlowExpectedError[incompatible-type-arg]
   y: 'abc',
 });
 
@@ -146,14 +144,15 @@ const mistakeNewInstance = new MakePointNew({ x: 'string' });
 // Note use of + for Read Only.
 type TPerson = { +name: string, +age: number };
 const defaultValues: TPerson = { name: 'Aristotle', age: 2400 };
-const PersonRecord = Record(defaultValues);
+const PersonRecord = Record<TPerson>(defaultValues);
 
-class Person extends PersonRecord<TPerson> {
+class Person extends PersonRecord {
   getName(): string {
     return this.get('name');
   }
 
   setName(name: string): this & TPerson {
+    // $FlowIssue[incompatible-return]
     return this.set('name', name);
   }
 }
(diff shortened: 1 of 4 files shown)


type-definitions/flow-tests/record.js
```diff
@@ -146,14 +144,15 @@ const mistakeNewInstance = new MakePointNew({ x: 'string' });
...
-class Person extends PersonRecord<TPerson> {
+class Person extends PersonRecord {
   getName(): string {
     return this.get('name');
   }
 
   setName(name: string): this & TPerson {
+    // $FlowIssue[incompatible-return]
     return this.set('name', name);
   }
```
mwiencek, line 157: With the new changes, Flow doesn't like the `this & TPerson` return type one bit:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ type-definitions/flow-tests/record.js:155:12

Cannot return this.set(...) because: [incompatible-return]
 • Either RecordInstance [1] is incompatible with this [2].
 • Or TPerson [3] is incompatible with this [2].

     type-definitions/flow-tests/record.js
 [3]  147│ const PersonRecord = Record<TPerson>(defaultValues);
         :
      152│   }
      153│
 [2]  154│   setName(name: string): this & TPerson {
      155│     return this.set('name', name);
      156│   }
      157│ }
      158│

     type-definitions/immutable.js.flow
 [1] 1659│   set<K: $Keys<T>>(key: K, value: $ElementType<T, K>): this & $ReadOnly<T>;



Found 1 error
```

type-definitions/flow-tests/record.js
```diff
@@ -170,23 +169,6 @@ person.get('unknown');
 // $FlowExpectedError[prop-missing]
 person.set('unknown', 'Thales');
 
-// Note: not <TPerson>
```
mwiencek, line 173: I'm not sure what the use case here is (and `PersonRecord` is no longer parameterized anyway) so I removed the test altogether.  If someone can explain the use case I can perhaps update it instead. :)

PR #1874, opened 2021-09-16
upgrade to jest 27 and jasmine-check to 1.0.0-rc

jdeniau:
I tried to replace https://github.com/immutable-js/immutable-js/pull/1849 by a better version now that jest should [handle async transform](https://jestjs.io/docs/configuration#transform-objectstring-pathtotransformer--pathtotransformer-object) since 27.0.

I did not manage easily to drop this `make-synchronous` or `synckit` call, I got a lot of different issues and I spent a lot of time already on it, but jest 27 does work fine 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "arrowParens": "avoid"
   },
   "devDependencies": {
-    "@types/jest": "^27.0.1",
+    "@types/jest": "27.0.1",
     "@types/react": "17.0.11",
     "benchmark": "2.1.4",
     "colors": "1.4.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -85,22 +85,8 @@
...
-      "^.+\\.(js|ts)$": "<rootDir>/resources/jestPreprocessor.js"
-    },
-    "testRegex": "/__tests__/.*\\.(ts|js)$",
-    "unmockedModulePathPatterns": [
-      "./node_modules/react"
-    ]
-  },
   "devDependencies": {
-    "@types/jest": "26.0.24",
+    "@types/jest": "^27.0.1",
```
leebyron, line 89: can you set these to exact versions?

leebyron:
Nice!

PR #1880, opened 2021-09-28
Changelog 4.0.0

bdurrer:
PR for discussions 🎉


bdurrer:
I removed `List.merge` from the diff, as it is still around as alias of concat.

The newly added `update` on several types is not mentioned in the changelog yet (besides the diff). I think it should be mentioned as new method in the keychanges... or maybe in the new section?

Methuselah96 (approved):

PR #1886, opened 2021-10-25
update typings to indicate `Seq()` faux-iterator support

ShawnTalbert:
To support pre-ES2015 environments where `Symbol.iterator` does not exist, recognizing
that the `Seq()` implementation does accept a 'false' Iterable in the form of an object that has a
property named "@@iterator" which returns the Iterator.

Discussion on PR immutable-js#1589

### DIFF
--- a/type-definitions/immutable.d.ts
+++ b/type-definitions/immutable.d.ts
@@ -3125,6 +3125,13 @@ declare namespace Immutable {
   }
 
   /**
+   * For pre-ES2015 support where `Symbol.iterator` does not exist.
+   */
+  interface FauxIterable<T> {
+    ['@@iterator'](): Iterator<T>;
+  }
+
+  /**
    * Creates a Seq.
    *
    * Returns a particular kind of `Seq` based on the input.
@@ -3147,7 +3154,11 @@ declare namespace Immutable {
   function Seq<K, V>(collection: Collection.Keyed<K, V>): Seq.Keyed<K, V>;
   function Seq<T>(collection: Collection.Set<T>): Seq.Set<T>;
   function Seq<T>(
-    collection: Collection.Indexed<T> | Iterable<T> | ArrayLike<T>
+    collection:
+      | Collection.Indexed<T>
+      | Iterable<T>
+      | FauxIterable<T>
+      | ArrayLike<T>
   ): Seq.Indexed<T>;
   function Seq<V>(obj: { [key: string]: V }): Seq.Keyed<string, V>;
   function Seq<K = unknown, V = unknown>(): Seq<K, V>;


ShawnTalbert:
The issue here was the difference between the ES spec (which is reflected in the TypeScript bundled lib definition for `Iterable<T>`) and what Immutable `Seq()` actually supports. From that context, whether the custom faux iterable implementation chosen by Immutable happens to coincide with the some pre-standard property name in browsers seems immaterial to the problem at hand? 

Succinctly, `Seq()`  accepts  a specific faux iterator shape but the immutable.d.ts declarations for that function do not reflect that. 

Lee, if the decision is yours, would you kindly confirm that Immutable will  __not__ be updating the type definitions to include the full range of acceptable inputs to `Seq()`? 

At least with that cemented I can know whether I'll be able to use 4.x, or be forced to stick with 3.8 forever or choose another library entirely.

conartist6:
I'll say one last thing: immutable is in no small part a legacy library. It does some pretty incredible things given that it also works in some pretty barebones environments. At some point though I think it should transition from being an es3 library to being an es6 library, and the 4.0 release is an ideal point to make that pivot. The API has been feature-stable for a long time, and 4.0 was largely a polish release getting rid of most known flaws in the implementation. What I'm saying is that I don't think we have to try to bury the idiosyncrasies of 4.x, because they're not long term. We want the people who care about beautiful design and integration to adopt 5.x, which of course we haven't even begun writing.

leebyron:
>  Lee if I understand what you're saying your point about confusing errors is that people who intended to pass an iterable as Seq(iterable) will be getting a big error detailing all the types and overloads that couldn't be matched, and now there will be an extra one.

Yes this plus my agreement with your previous point about `Iterator` being the correct type and avoiding that union confusion across all 100+ uses.

> From that context, whether the custom faux iterable implementation chosen by Immutable happens to coincide with the some pre-standard property name in browsers seems immaterial to the problem at hand?

The case I was trying to make before is that this is exactly material. It wasn't that Immutable.js wanted some custom faux iterable implementation and happened to choose `"@@iterator"`, that support is there precisely because this was the browser standard before ES6 and Immutable.js wanted to support a broad range of environments. It's not a faux iterable, it's a real iterable standard. It's just not the one defined by ES6, it's the one defined by Firefox years earlier (and adopted by other browsers).

Your case of having an ES5 environment to support is exactly the reason why this behavior exists in Immutable.

> The issue here was the difference between the ES spec (which is reflected in the TypeScript bundled lib definition for Iterable<T>) and what Immutable Seq() actually supports.

Totally agree. Runtime behavior is supporting ES5 correctly, so this is a purely a matter of supplying the correct TypeScript types.

Maybe this seems like a subtle difference, but I think it's actually quite important to get right - `Seq()` (and other functions that operate on iterables) have runtime support for ES5 and ES6 Iterables, but we shouldn't conceptualize these as different things - they're both different runtime implementations of one kind of type. We would not expect ES5 and ES6 iterators to be floating around the same runtime environment, and so we probably should not write type libs that allow that.

Also note that Immutable's typedefs are not specifically using the ES6 Iterable type lib. It just expects that you'll have have a TypeScript environment that has this type defined. Since there is a specification for them in ES6, TypeScript supplies that typelib directly for you to include in your environment via `lib.es2015.iterable.d.ts`. I suspect if the ES5 spec had formally defined them, then TypeScript would have defined `lib.es5.iterable.d.ts` for use. Then this issue would be very easily resolved - you would just use import the ES5 typelib, instead of the ES6 typelib, since that would correctly describe your execution environment. 

My recommendation stands that since TypeScript does not offer this directly, you can provide one yourself. Luckily it is a really small lib to write.

<details>
<summary>lib.es5.iterable.d.ts</summary>

```ts
interface Iterator<T> {
  next(): { done: boolean, value: T };
}

interface Iterable<T> {
  ['@@iterable'](): Iterator<T>;
}

interface IterableIterator<T> extends Iterator<T> {
  ['@@iterable'](): IterableIterator<T>;
}

interface Array<T> {
  ['@@iterable'](): IterableIterator<T>;
}

interface ReadonlyArray<T> {
  ['@@iterable'](): IterableIterator<T>;
}
```
</details>

> Lee, if the decision is yours, would you kindly confirm that Immutable will not be updating the type definitions to include the full range of acceptable inputs to Seq()?

Since Immutable.js's type defs expect `Iterable` to be ambiently defined, it already does describe the full range of acceptable inputs as long as your environment uses one of the two implementations our runtime definition supports (ES6 or ES5).

I don't think it would be the right thing to do to have Immutable.js directly define the ES5 Iterables type. Especially since that would impact the `Array` and `ReadonlyArray` ambient types, which would not be correct for those using an ES6 environment.

leebyron:
While directly including ES5 iterators this in the immutable.d.ts file could have breaking changes to the environment, I could still be reasonable to offer this somewhere easy for people to import?

conartist6:
@ShawnTalbert Does your environment define `Array.prototype.@@iterator`?

ShawnTalbert:
Nay, there is no `Array.prototype['@@iterator']` in this ES5 world I'm running.

conartist6:
Right. So I think that means I'm settled in disagreement with @leebyron. If you want to make a change to all environments, the process exists: you get it put into the spec. Since `@@iterator` is non-spec, you must assume that it is present in only some environments. Thus to me the only path forward that provides correctness in the type system is to widen the type even if errors get uglier.

leebyron:
> If you want to make a change to all environments, the process exists: you get it put into the spec

Today yes. In 2012... specs were stagnant and browsers just shipped shit and copied each other. My point above is that Iterators existed before ES6, even if not uniformly so (browsers did not used to be uniform) and their implementation is the one Immutable.js is designed to support.

> Thus to me the only path forward that provides correctness in the type system is to widen the type even if errors get uglier.

I'm not sure why this is our only path forward. It has significant downsides around type complexity and degraded error messages.

It also doesn't solve the key problem that `Iterator` is an undefined ambient type. For example if we introduce `IteratorLike<T> = Iterator<T> | FauxIterator<T>` and you tell TS that you're using ES5, then it will still complain that `Iterator` is not defined.

Perhaps another option to solve this could be done with two steps:

1. Expand any input (fn arg) of type `Iterator<T>` to `Iterator<T> | ReadonlyArray<T>` so that we remove the assumption that Arrays are iterable, since that might not be a safe one in many ES5 environments. This should be non-breaking. That would further simplify defining your own `Iterator` ambient so that it would not require extending other globals types.
2. Introduce `immutable.es5.d.ts` which imports `immutable.d.ts` and defines the `Iterator` global ambient using `"@@iterator"`. This would preserve high quality types for all ES6 users, which is the vast majority, while still offering a type definition file to import for ES5 users. This would still require a little bit of config to ensure the right type file was being imported, but perhaps that's preferable to defining your own `Iterator` type.

Especially after the first step, then an ES5-Firefoxish type def for Iterator becomes simply:

<details>
<summary>lib.es5.iterator.d.ts</summary>

```ts
interface Iterator<T> {
  next(): { done: boolean, value: T };
}

interface Iterable<T> {
  ['@@iterable'](): Iterator<T>;
}

interface IterableIterator<T> extends Iterator<T> {
  ['@@iterable'](): IterableIterator<T>;
}
```
</details>

ShawnTalbert:
Alas, the runtime environment I'm in is server-side, __not a browser__ and has no de-facto concept of `Iterator<T>` whatsoever. From that perspective, `Iterator<T>` is a type from Immutable just as much as `Seq` is.

In my scenario I have two runtimes to support - an ES5 compatible environment and an ES2019 environment. Also indirectly, ES6+ browsers). To support this with one build, I've been targeting ES5 which functions under both environments due to good backwards compatibility.

I don't see a way to resolve a single global `Iterable<T>` that works cleanly across ES5/ES6+ without introducing something external (and even then it would be an approximation for ES5). I totally forgot about the point from your last post about the global dependency on an `Iterable<T>`; seems insurmountable from ES5 without external help. I tried some things but couldn't come up with any way to represent the concept of `type Iterable<T> = Iterable<T> ?? FauxIterable<T>`

I'm inclined to perhaps just lock the current version of our library to Immutable < 4 and it would remain compatible in all runtimes. Then do a major version bump to target the newer runtime only and hence Immutable 4+.  If my use case is really fringe I hate to waste your time.

Admittedly, the majority of what we do now is in the newer runtime, but there's so much legacy code still running in that old runtime.... It's been nice over the years to be able to upgrade  library and use it freely in either environment due to ES5 broad compatibility.

conartist6:
@ShawnTalbert It sounds like perhaps you should just fork immutable 4 and publish `@shawntalbert/immutable` or whatever would be an acceptable name. Personally I think forking has gotten a stigma that it really doesn't deserve, so when I need something different and can't get maintainers on board, I just go for it. Obviously it's harder though for libraries where I expect there might be a major amount of work keeping my fork up to date (or serious consequences for failing to do so) but I don't think either is the case here.

leebyron:
Less costly than a fork would be just a fork of the type defs, since the runtime itself doesn't need to change.

Considering that ES6+ Iterables depend on Symbol, and one of your constraints is that you don't want to introduce types that do not accurately reflect the environment you're operating in, then think it's going to be hard to describe one type for `Iterable`, even as a union.

Options for you going forward:

1. Define two different TS environments, one for ES5 and one for ES6+.
  * Pro: Will give you accurate types per environment. Code shared across environments will be required to conform to each.
  * Con: Complicates your setup having multiple TS configs.
2. Define Iterable exclusively with the pre-symbol (ES5 friendly) definition.
  * Pro: Simple lowest common denominator that will definitely work in your ES5 env.
  * Con: Won't provide TS support for any native iterables. May create an issue if your ES2019 environment does in fact include ES6 TS type defs such that `Iterable` is in conflict.
3. Fork the typedefs and replace the Iterable type with something else
  * Pro: Lets you define the type as you wish. Avoids any collision with environmental types.
  * Con: Won't provide TS support for any native iterables. Requires you to maintain your own typedef.
4. Keep using v3.x
  * Pro: If it's working for you, it will keep working for you
  * Con: Won't provide TS support for any native iterables. v4 fixed many logic and api issues that you forego.

PR #1903, opened 2022-05-08
Fix bug in Record.equals when comparing against Map

jmtoung:
This commit fixes a bug introduced by this [change](https://github.com/immutable-js/immutable-js/commit/a9a1612897627a6ad1217bfa0a76203c95e710b1), which removed a check that '._keys' exists on the object being compared in Record.equals. This leads to a runtime uncaught error when a Record instance is compared against a Map instance.

### DIFF
--- a/__tests__/Record.ts
+++ b/__tests__/Record.ts
@@ -88,6 +88,12 @@ describe('Record', () => {
     expect(t1.equals(null)).toBeFalsy();
   });
 
+  it('if compared against Map should return false', () => {
+    const MyType = Record({ a: 1, b: 2 });
+    const t1 = MyType();
+    expect(t1.equals(Map({ a: 1, b: 2 }))).toBeFalsy();
+  });
+
   it('merges in Objects and other Records', () => {
     const Point2 = Record({ x: 0, y: 0 });
     const Point3 = Record({ x: 0, y: 0, z: 0 });
--- a/src/Record.js
+++ b/src/Record.js
@@ -120,7 +120,8 @@ export class Record {
 
   equals(other) {
     return (
-      this === other || (other && recordSeq(this).equals(recordSeq(other)))
+      this === other ||
+      (isRecord(other) && recordSeq(this).equals(recordSeq(other)))
     );
   }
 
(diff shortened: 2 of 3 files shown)


src/Record.js
```diff
@@ -120,7 +120,11 @@ export class Record {
 
   equals(other) {
     return (
-      this === other || (other && recordSeq(this).equals(recordSeq(other)))
+      this === other ||
+      (other &&
+        this._keys &&
```
leebyron, line 125: `this._keys` should always be set because the instance doing the comparison is a Record

__tests__/Record.ts
```diff
@@ -88,6 +88,12 @@ describe('Record', () => {
     expect(t1.equals(null)).toBeFalsy();
   });
 
+  it('if compared against Map should return false', () => {
+    const MyType = Record({ a: 1, b: 2 });
+    const t1 = MyType();
+    expect(t1.equals(Map({ a: 1, b: 2 }))).toBeFalsy();
```
leebyron, line 94: For completeness can you add a test which flips the operands to ensure the symmetric property?

leebyron (changes requested):
This is great! Thank you for spotting and resolving this

src/Record.js
```diff
@@ -120,7 +120,11 @@ export class Record {
...
+        other._keys &&
```
leebyron, line 126: I wonder if the proper way to do this should be `isRecord(other)`?

jmtoung: It's definitely semantically more obvious what the check is doing but it depends on the guarantee that Record will always have _keys instantiated, which we will be alerted on if that is ever violated with this newly added unit test.

__tests__/Record.ts
jmtoung: Sure

src/Record.js
jmtoung: Will fix

src/Record.js
```diff
@@ -120,7 +120,8 @@ export class Record {
 
   equals(other) {
     return (
-      this === other || (other && recordSeq(this).equals(recordSeq(other)))
+      this === other ||
+      (isRecord(other) && recordSeq(this).equals(recordSeq(other)))
```
leebyron, line 124: More correct and more clear. Nice!

leebyron (approved):

PR #1913, opened 2022-11-14
fix Set.concat to concatenate strings as items

rezaa91:
https://github.com/immutable-js/immutable-js/issues/1910

This is my first time contributing to open source so please let me know what else is required


jdeniau:
Thanks. 

I just updated the immutable-js types accordingly

leebyron:
Coming to this late, but thank you for fixing this!

There may be an opportunity to generalize this using Symbol.isConcatSpreadable if available, which we should consider if a different variant of this issue appears

PR #1917, opened 2022-12-19
better type for Record.toJS

jdeniau:
Record.toJS loose the Record type and return something like:

Actually the following code:

```ts
Record({ keyA: 'a', keyB: 'b' });
```

will return this type.

```ts
{
    keyA: unknown,
    keyB: unknown,
}
```

This PR implement the `DeepCopy` functionnality to handle the deep convertion made by `toJS`

It will now return this:
```ts
{
    keyA: string,
    keyB: string,
}
```

It does handle recursive types:

```ts
type NestedObject = Record({ map: Map<string, string>(); list: List<string>(); set: Set<string>() });
// type will be : { map: { [key: string]: string }; list: string[]; set: string[]; }
```

### DIFF
--- a/type-definitions/immutable.d.ts
+++ b/type-definitions/immutable.d.ts
@@ -91,6 +91,9 @@
  */
 
 declare namespace Immutable {
+  /**
+   * @ignore
+   */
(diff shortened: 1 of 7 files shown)


leebyron:
This is so good

PR #1920, opened 2022-12-23
Add partition to flow spec

Dagur:


jdeniau:
> Cannot use type without exactly 1 type argument. [nonpolymorphic-type-app]

It seems that tuples are handled this way in flow:

```js
[this, this]
```

and not 

```
Array<this, this>
```

[source](https://flow.org/en/docs/types/tuples/)

Or it might be an issue with an old version of flow. If you want, you can upgrade the flow version to a most recent one

Dagur:
Yeah this was completely wrong, I'm not sure why it wasn't flagged.
I'm fixing this now

Dagur:
Should be alright now btw

jdeniau:
Thanks @Dagur . Released in `4.2.2`

Dagur:
Thank you too. 
One question though, the latest version on npm is still 4.2.1. Is that just them being slow?

jdeniau:
@Dagur thanks for notifying me, the released process did fail. The `4.2.2` is now available on npmjs

Dagur:
awesome. Thanks!

PR #1924, opened 2023-01-09
groupBy return either a Map or an OrderedMap: make the type more precise

jdeniau:


type-definitions/immutable.d.ts
```diff
@@ -4600,7 +4600,7 @@ declare namespace Immutable {
     groupBy<G>(
       grouper: (value: V, key: K, iter: this) => G,
       context?: unknown
-    ): /*Map*/ Seq.Keyed<G, /*this*/ Collection<K, V>>;
+    ): Map<G, /*this*/ Collection<K, V>>;
```
jdeniau, line 4603: We might want to override each class to be even more precise as 

```ts
List<number>().groupBy(v => v); // is a Map<G, List<K, V>>
Set<number>().groupBy(v => v); // is a Map<G, Set<V, V>>
// etc.
```

jdeniau: I tried to implement this, but got a lot of error:

> Type instantiation is excessively deep and possibly infinite.

I did not manage to solve this, so I merge this as the current implementation is nearly right: we lose the fact that the Map is, sometimes, ordered, but we still gain all Map API.

PR #1927, opened 2023-02-03
Improve type infererence for fromJS

jdeniau:
Originally from [KSXGitHub](https://github.com/KSXGitHub) in https://github.com/immutable-js/immutable-js/pull/1617

> With this pull request, fromJS(value) will no longer return any.

### DIFF
--- a/type-definitions/immutable.d.ts
+++ b/type-definitions/immutable.d.ts
@@ -5160,12 +5160,39 @@ declare namespace Immutable {
    */
   function fromJS(
     jsValue: unknown,
-    reviver?: (
+    reviver: (
       key: string | number,
       sequence: Collection.Keyed<string, unknown> | Collection.Indexed<unknown>,
       path?: Array<string | number>
     ) => unknown
   ): Collection<unknown, unknown>;
+  function fromJS<JSValue>(
+    jsValue: JSValue,
+    reviver?: undefined
+  ): FromJS<JSValue>;
+
+  type FromJS<JSValue> = JSValue extends FromJSNoTransform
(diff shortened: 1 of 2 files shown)


jdeniau:
Thanks @KSXGitHub ,I will release this next week probably.

jdeniau:
This has been released in 4.2.4. Thanks !

type-definitions/immutable.d.ts
```diff
@@ -5160,12 +5160,39 @@ declare namespace Immutable {
    */
   function fromJS(
     jsValue: unknown,
-    reviver?: (
+    reviver: (
```
ntucker, line 5163: No way this is correct - I never had problems skipping this argument in runtime.

jdeniau: hi @ntucker . That's an overload for the definition right after. It does say : 
- if the reviver method is set, then the return type is `Collection<unknown, unknown>`
- if the reviver is not set of undefined, then we can calculate the return type

ntucker: https://github.com/immutable-js/immutable-js/issues/1935

https://stackblitz.com/edit/typescript-sk2pyy?file=index.ts

ntucker: This may very well be a bug with typescript, but I'm fairly certain you'd still be interested in supporting typescript versions less than 5.1

jdeniau: Let's continue in #1935 to avoid multiplying conversation in different threads

PR #1932, opened 2023-02-06
manage to handle toJS circular reference.

### LINKED ISSUE #1930: `toJS` breaks type compatibility between records with object values
### What happened

Records with object values that would otherwise be compatible are not because of `toJS`. I've noticed this when fields only differ by optionality.

### How to reproduce

```
import { RecordOf } from 'immutable';

interface Id {}

type A = RecordOf<{ id: Id }>;
type B = RecordOf<{ id?: Id }>;

const a: A = null as any;
const b: B = a;
```

Results in the error
```
Type 'A' is not assignable to type 'B'.
  Type 'A' is not assignable to type 'Record<{ id?: Id | undefined; }>'.
    The types of 'set(...).toJS().id' are incompatible between these types.
      Type 'unknown' is not assignable to type 'Id | undefined'.
```

I think this happens due to the workaround for circular types where objects become `unknown`. Honestly, I think the new types are more trouble than they're worth and would prefer the 4.1 types.

jdeniau:
Fixes #1930

### Related Pull requests

- https://github.com/immutable-js/immutable-js/pull/1922 (handle the `infer Record` did the trick here)

### DIFF
--- a/type-definitions/immutable.d.ts
+++ b/type-definitions/immutable.d.ts
@@ -100,27 +100,25 @@ declare namespace Immutable {
   export type DeepCopy<T> = T extends Record<infer R>
     ? // convert Record to DeepCopy plain JS object
       {
-        [key in keyof R]: R[key] extends object ? unknown : R[key];
+        [key in keyof R]: DeepCopy<R[key]>;
       }
     : T extends Collection.Keyed<infer KeyedKey, infer V>
     ? // convert KeyedCollection to DeepCopy plain JS object
       {
         [key in KeyedKey extends string | number | symbol
           ? KeyedKey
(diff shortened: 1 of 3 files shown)


shaharyar-shamshi:
@jdeniau any update on this PR

GNRSN:
I'm also following for any updates to this 👀

jdeniau:
Merging this as it slightly improve the typing, but feel free to report any issue related to this change !

jdeniau:
For the record I partially reverted this as it does cause a circular reference issue #1957 

I spend 6 hours trying to fix this, but did not found a proper way. I'm forced to rollback.

PR #1937, opened 2023-02-12
Introduce Comparator and PairSorting

giancosta86:
* `Comparator<T>`, together with the `PairSorting` enum, would make sorting more expressive - without breaking backwards compatibility.

### DIFF
--- a/type-definitions/immutable.d.ts
+++ b/type-definitions/immutable.d.ts
@@ -127,6 +127,8 @@ declare namespace Immutable {
 
   /**
    * Describes which item in a pair should be placed first when sorting
+   *
+   * @ignore
    */
   export enum PairSorting {
     LeftThenRight = -1,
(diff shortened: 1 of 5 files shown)


type-definitions/immutable.d.ts
```diff
@@ -125,6 +125,23 @@ declare namespace Immutable {
     : // other case : should be kept as is
       T;
 
+  /**
+   * Describes which item in a pair should be placed first when sorting
```
jdeniau, line 129:
```suggestion
   * Describes which item in a pair should be placed first when sorting
   *
   * @ignore
```

to disable automatic output from in the documentation

type-definitions/immutable.d.ts
```diff
@@ -125,6 +125,23 @@ declare namespace Immutable {
...
+    LeftThenRight = -1,
+    RightThenLeft = +1,
+  }
+
+  /**
+   * Function comparing two items of the same type. It can return:
+   *
+   * * a PairSorting value, to indicate whether the left-hand item or the right-hand item should be placed before the other
+   *
+   * * the traditional numeric return value - especially -1, 0, or 1
```
jdeniau, line 141:
```suggestion
   * * the traditional numeric return value - especially -1, 0, or 1
   *
   * @ignore
```

jdeniau (changes requested):

jdeniau (approved):

PR #1952, opened 2023-08-14
Add install instructions for pnpm and Bun

colinhacks:


jdeniau (commented):
_see above_

jdeniau (changes requested):
Hi,

Thanks for the PR, but it seems that you have an automatic linter or something like that that change all the file, including some text case.
Can you update your PR to revert the linting change?

Thank you

jdeniau (approved):

jdeniau:
Thanks @colinhacks

PR #1962, opened 2023-08-28
Remove tree-shaking possibility (broken)

jdeniau:
Sadly reverting tree-shaking possibility (https://github.com/immutable-js/immutable-js/pull/1888) as it does break. Moreover, as nearly every collection type can be converted to another collection (`Map().toList()` for example), so nearly no code was removed.

I think that if we want to implement this, we might need to rethink this functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ Dates are formatted as YYYY-MM-DD.
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


jdeniau:
@bdurrer I willl merge this quickly to make v5 functional, but I would really appreciate your opinion on this.

PR #1968, opened 2023-10-04
Updated type tests for [Ordered]Set.fromKeys

johnw42:
This change updates the tests to report that the actual type inferred by tsc is correct. It would be easy to make the existing tests pass by changing the order of the overloadings in immutable.d.ts, but in this case I believe tsc is doing the right thing: it can't automatically infer the correct type of the values returned by `fromKeys`, so it's better to infer the type as `unknown`.

### DIFF
--- a/type-definitions/ts-tests/ordered-set.ts
+++ b/type-definitions/ts-tests/ordered-set.ts
@@ -42,7 +42,7 @@ import { OrderedSet, Map } from 'immutable';
 {
   // .fromKeys
 
-  // $ExpectType OrderedSet<string>
+  // $ExpectType OrderedSet<unknown>
   OrderedSet.fromKeys(Map<number, string>());
 
   // $ExpectType OrderedSet<number>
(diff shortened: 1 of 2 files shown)


johnw42:
Odd. I see dtslint is reporting an error under Typescript 4.9, but for me, it's fine with my changes, but without them it reports an error with Typescript 5.0.

johnw42:
Ok, I'm getting more reasonable results after running `npm ci`. The trouble is that either way the test is written, it fails with either TS 4.9 or 5.0. It's still possible to make the tests pass by swapping the order of overloads in immutable.d.ts as lines 1716-1717 and 1939-1940, but I still think that's a bad idea, and the new behavior of tsc is more correct. For now I think the best solution is to remove those two type tests entirely; what do you think?

jdeniau:
Hi !

it seems like the runtime is not that easy:

image

Because if you pass an object, number keys are converted to string, but if you `set`, this stays a number :/

jdeniau:
Hi, I fixed this comportment in #1971 : it seems that the issue is with Collection.Keyed.

About the issue I did point in my previous comment, I think the issue is with the Map constructor and not `Set.fromKeys`

PR #1987, opened 2024-02-20
chore: remove the `tslint.json` file

mrazauskas:
This PR removes the `tslint.json` file.

This file holds configuration / rules for `tslint`. It is not used after #1979. (Or?)


type-definitions/ts-tests/tslint.json
```diff
@@ -1,10 +0,0 @@
-{
-  "extends": "dtslint/dtslint.json",
```
mrazauskas, line 2: Despite having `dtslint` in its name, this file simply holds a list of `tslint` rules, reference: https://github.com/microsoft/dtslint/blob/master/dtslint.json

mrazauskas:
Ah.. `dtslint` is using it. So this could be part of #1986 (or land as a follow up).

/cc @arnfaldur

arnfaldur:
Indeed, removing the file has no effect on the test suite after the changes in https://github.com/immutable-js/immutable-js/pull/1986

jdeniau:
Closing this, as it does not work.

PR #1988, opened 2024-02-21
Migrate type tests to TSTyche

mrazauskas:
Reference https://github.com/immutable-js/immutable-js/issues/1985#issuecomment-1956085576

This PR migrates two type test files to TSTyche. The setup is painless and migration can be done gradually. Hope this makes it easier to review.

The test files were not passed through Prettier before (probably because of some conflict).

Good question: how CI checks should be setup? To test on different versions of TypeScript, I used TSTyche's `target` option. This makes local runs noisy. It would be enough to have `test:types --target 4.5,5.0,current` in CI. @jdeniau what you think?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
           restore-keys: ${{ runner.OS }}-node-
       - run: npm ci
       - run: npm run build
-      - run: npm run unit-test
+      - run: npm run test:unit
       - run: npx size-limit
       - run: npm run check-git-clean
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,8 +50,8 @@ jobs:
       - run: npm run type-check
       - run: npm run check-git-clean
 
-  unit-test:
-    name: 'Build & Unit Test'
+  test:
+    name: 'Build & Unit Test & Type Test'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -66,6 +66,7 @@ jobs:
       - run: npm ci
       - run: npm run build
       - run: npm run test:unit
+      - run: npm run test:types -- --target 4.5,5.0,current
       - run: npx size-limit
       - run: npm run check-git-clean
 
@@ -90,7 +91,7 @@ jobs:
 
   publish:
     name: 'Publish'
-    needs: [lint, type-check, unit-test, website]
+    needs: [lint, type-check, test, website]
     if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
--- a/package.json
+++ b/package.json
@@ -55,7 +55,9 @@
     "npm": ">=7.0.0"
   },
   "scripts": {
-    "test": "run-s format lint type-check build unit-test",
+    "test": "run-s format lint type-check build test:*",
+    "test:unit": "jest",
+    "test:types": "tstyche",
     "format": "npm run lint:format -- --write",
     "lint": "run-s lint:*",
     "lint:format": "prettier --check \"{__tests__,src,type-definitions,website/src,perf,resources}/**/*{.js,.ts,.tsx,.flow,.css}\"",
@@ -69,7 +71,6 @@
     "build:types": "cpy ./type-definitions/immutable.* dist",
     "build:prepare": "./resources/prepare-dist.sh",
     "build:stats": "node ./resources/dist-stats.mjs",
-    "unit-test": "jest",
     "website:build": "cd website && next build && next-sitemap",
     "website:dev": "cd website && next dev",
     "check-git-clean": "./resources/check-git-clean.sh",
@@ -127,6 +128,7 @@
     "rollup": "3.28.1",
     "size-limit": "^8.2.6",
     "transducers-js": "0.4.174",
+    "tstyche": "^1.0.0",
     "typescript": "5.1"
   },
   "size-limit": [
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -55,7 +55,9 @@
     "npm": ">=7.0.0"
   },
   "scripts": {
-    "test": "run-s format lint type-check build unit-test",
+    "test": "run-s format lint type-check build test:*",
+    "test:unit": "jest",
+    "test:types": "tstyche",
```
mrazauskas, line 60: I added `test:types` script to make it easy to pass extra options. For example, in CI the command could be `test:types --target 4.5,5.0,current`. Just a suggestion, it can be set up somehow else.

Currently `type-check` script is used. In early stages I saw TSTyche as a **ty**pe **che**cker (this is where the name is coming from), but it does more. It uses TypeScript to perform static analysis and file are not executed, but with `.only` / `.skip` and other features it feels more like a **type test runner**.

jdeniau: Actually, we do have [this file](https://github.com/immutable-js/immutable-js/blob/d7664bf9d3539da8ea095f2ed08bbe1cd0d46071/type-definitions/ts-tests/index.d.ts#L1) telling that the minimum version is 4.5, and I think that we do only test 4.5 (not really sure about that, we might use the local typescript version).

It might be interesting to test multiple typescript version in CI (but not done actually, so if it can be do, it's great, but not a necessity).

Locally, I think that testing the local version of typescript is good enough.

mrazauskas: Right. I forgot to mention that `current` is currently (or locally) installed TypeScript. This is the default. I mean, `tstyche` is the same as `tstyche --target current`.

.github/workflows/ci.yml
```diff
@@ -65,7 +65,8 @@ jobs:
           restore-keys: ${{ runner.OS }}-node-
       - run: npm ci
       - run: npm run build
-      - run: npm run unit-test
+      - run: npm run test:unit
+      - run: npm run test:types -- --target 4.5,5.0,current
```
mrazauskas, line 69: @jdeniau What you think? Locally `test:types` would test only on the installed TypeScript. The lowest supported version (`4.5`, resolves to `4.5.5`) is checked on CI. I would definitely recommend adding `5.0` (resolves to `5.0.2`), because each major brings in breaking changes. And currently installed.

`latest` could be added. It points to `typescript@latest` in NPM registry, hence it is a changing target. It might start failing with any PR unrelated, i.e. even changes in types. That’s somewhat unpredictable.

mrazauskas: `latest` or even `next` could work as a daily cron job. Just some food for though.

type-definitions/ts-tests/empty.ts
```diff
@@ -1,57 +1,44 @@
...
-  Seq.Keyed();
+  expect(Seq.Keyed<number, string>()).type.toEqual<Seq.Keyed<number, string>>();
 
-  // $ExpectType Keyed<number, string>
-  Seq.Keyed<number, string>();
+  expect(Seq.Set()).type.toEqual<Seq.Set<unknown>>();
 
-  // $ExpectType Set<unknown>
-  Seq.Set();
+  expect(Seq.Set<string>()).type.toEqual<Seq.Set<string>>();
```
mrazauskas, line 19: Interestingly `dtslint` does not see difference between `Set` and `Seq.Set`. Probably this is how it works. I did not know this detail. Learned something new.

type-definitions/ts-tests/es6-collections.ts
```diff
@@ -1,18 +1,23 @@
...
-const setImmutable: ImmutableSet<string> = ImmutableSet<string>();
+test('immutable.js collections', () => {
+  const mapImmutable: ImmutableMap<string, number> = ImmutableMap<
+    string,
+    number
+  >();
+  const setImmutable: ImmutableSet<string> = ImmutableSet<string>();
 
-// $ExpectType Map<string, number>
-mapImmutable.delete('foo');
```
mrazauskas, line 11: Here this is `ImmutableMap` not `Map`. Seems like `dtslint` is not aware of the context.

mrazauskas:
Two more smaller files are added. Sorry, I just couldn’t stop (;

How is it better: to push all tests here? or, to wait until this PR will be merged?

jdeniau:
> Two more smaller files are added. Sorry, I just couldn’t stop (;
> 
> How is it better: to push all tests here? or, to wait until this PR will be merged?

If you are on fire, just push it here, and tell me when you want me to review 😄 
If you handle to remove dtslint, it's perfect 🎉 

The only thing is I need to manually trigger a "run" because of the project configuration, so I'll do it when I can,  but you might be faster than me.

mrazauskas:
Would be nicer, if CI could run automatically.

All checks are green at the moment. Let’s merge this one, if all looks good for you. I see that in larger files I will have some questions. Would be easier to go through these in separate PRs.

type-definitions/ts-tests/empty.ts
jdeniau: Indeed it is 😄

type-definitions/ts-tests/es6-collections.ts
jdeniau: If I understood dtslint, it does compare the string output, so like it is in the `immutable.d.ts` file, so `Map` and not `ImmutableMap`

mrazauskas: Yes. It works correctly, only looks strange.

jdeniau:
Merging this as everything is OK and the transition is promising

mrazauskas:
Wonderful! Thanks for quick review.

Let's iterate. I will open the next PR tomorrow.

PR #1989, opened 2024-02-22
Migrate more type tests to TSTyche

mrazauskas:
Following up #1988

This PR migrates more type tests to TSTyche.


type-definitions/ts-tests/exports.ts
```diff
@@ -14,32 +15,98 @@ import {
   Collection,
 } from 'immutable';
 
-List; // $ExpectType typeof List
-Map; // $ExpectType typeof Map
-OrderedMap; // $ExpectType typeof OrderedMap
-OrderedSet; // $ExpectType typeof OrderedSet
-// TODO: Turn on once https://github.com/Microsoft/dtslint/issues/19 is resolved.
-Range; // $ ExpectType (start?: number | undefined, end?: number | undefined, step?: number | undefined) => Indexed<number>
```
mrazauskas, line 22: Notice that few assertions were disabled by adding a space: `// $ ExpectType` instead of `// $ExpectType`. I think now they work as expected.

mrazauskas:
One question. I work on `List` and `Map` tests. There are cases like this:

```ts
// $ExpectError
List().size = 10;

// $ExpectError
Map().size = 10;
```

This makes sure that `.size` is readonly. Triggering errors is not always the best testing strategy, because it is not clear what is being tested. Currently in TSTyche I can write these tests like this:

```ts
expect(List()).type.toMatch<{ readonly size: number }>();

expect(Map()).type.toMatch<{ readonly size: number }>();
```

Somewhat better, but still indirect. Here is new idea:

```ts
expect(List().size).type.toBeReadonly();

expect(Map().size).type.toBeReadonly();
```

Easy to read (and to implement). @jdeniau What you think?

type-definitions/ts-tests/functional.ts
```diff
@@ -1,84 +1,63 @@
...
 
-  // $ExpectType boolean
-  has([1, 2, 3], 0);
+test('set', () => {
+  expect(set([1, 2, 3], 0, 10)).type.toEqual<number[]>();
 
-  // $ExpectType boolean
-  has({ x: 10, y: 20 }, 'x');
-}
+  expect(set([1, 2, 3], 0, 'a')).type.toRaiseError();
```
mrazauskas, line 25: Here and elsewhere `.not.toBeCallableWith()` matcher should be used instead of `.toRaiseError()`. Keeping it as it was , but I will come back after implementing `.toBeCallableWith()`:

```ts
expect(set).type.not.toBeCallableWith([1, 2, 3], 0, 'a');
```

jdeniau:
```ts
expect(List().size).type.toBeReadonly();
```

seems easy to read, so it's a 👍  for me

mrazauskas:
Thanks for heads up regarding `.toBeReadonly()`. Noted it in my todo list.

If all looks good here, this PR can be merged.

type-definitions/ts-tests/exports.ts
jdeniau: Range [did change in 5.x](https://github.com/immutable-js/immutable-js/blob/5.x/CHANGELOG.md#500-beta5) in that way yes 😉

type-definitions/ts-tests/from-js.ts
```diff
@@ -1,49 +1,42 @@
-import { fromJS, List, Map } from 'immutable';
+import { expect, test } from 'tstyche';
+import { fromJS, Collection, List, Map, MapOf } from 'immutable';
 
-{
-  // fromJS
+test('fromJs', () => {
```
jdeniau, line 4:
```suggestion
test('fromJS', () => {
```

jdeniau (approved):

jdeniau:
Unrelated failing tests  (error in `dist-stats`, probably with an external call)

PR #1993, opened 2024-04-16
Use renamed matchers in type tests

mrazauskas:
This is just a minor change. I noticed that few matchers were not renamed after upgrading TSTyche to v2.


jdeniau (approved):
Thanks!

PR #1994, opened 2024-05-10
Fix Repeat(<value>).equals(undefined) incorrectly returning true

butchler:
It seems like there was a mistake in this old commit, when `deepEqual` was switched from an inherited method to a helper function: https://github.com/immutable-js/immutable-js/commit/79e903b986b2dcef24f5200287fc92c1b7d11810#diff-3ef08c64dab5fc92ebb8aecee330d5e7ae6cc8651f95aab2968b392c228dc606R103

Note that this only happens when calling the `.equals` method directly. `Immutable.is(Immutable.Repeat(<value>), undefined)` properly returns false, so this is a very minor issue that's unlikely to happen in practice. I only noticed this because I was looking at the immutable-js source code.

### DIFF
--- a/__tests__/Repeat.ts
+++ b/__tests__/Repeat.ts
@@ -11,4 +11,8 @@ describe('Repeat', () => {
     expect(v.toArray()).toEqual(['wtf', 'wtf', 'wtf']);
     expect(v.join()).toEqual('wtf,wtf,wtf');
   });
+
+  it('does not claim to be equal to undefined', () => {
+    expect(Repeat(1).equals(undefined)).toEqual(false);
+  });
 });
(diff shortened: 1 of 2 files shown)


jdeniau (approved):
Hi!
Thanks for this fix.

Would you mind adding a test that did previously failed and now works?

Thanks

butchler:
> Would you mind adding a test that did previously failed and now works?

I pushed a commit that adds a test just now.

jdeniau:
Thanks, I did merge but I'm not at home actually. I will push a fix probably next week and on the 5.x branch

jdeniau:
This has been released in v4.3.6

PR #1998, opened 2024-05-17
applitopia's sorted Map and Sorted Set implementation

simonacca:
This is an attempt to upstream applitopia's [implementation](https://github.com/applitopia/immutable-sorted) of sorted map and sorted set.

The code, tests and type definitions are exactly as in the fork, except:

- I have omitted the "partial sort" implementation in the hope of keeping this as small (and therefore as mergeable) as possible
- I have added one commit on top with whatever was needed to make the test suite and linter happy

While I am happy with the result as it is (up to date immutable + sorted map), I am not that familiar with the internals of Immutable, therefore I am not particularly eager to steward this PR all the way to merge. If one of the maintainers is interested and willing to do so that would be much appreciated! Otherwise it can be closed and perhaps serve as a starting point for the next brave soul to attempt this feat :)

### DIFF
--- a/__tests__/SortedMap.ts
+++ b/__tests__/SortedMap.ts
@@ -5,19 +5,10 @@
  *  LICENSE file in the root directory of this source tree.
  */
 
-/**
- * Copyright (c) 2014-present, Facebook, Inc.
- *
- * Original source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- */
-
-///<reference path='../resources/jest.d.ts'/>
+import { is, List, Range, Record, Seq, SortedMap } from 'immutable';
 
 import * as jasmineCheck from 'jasmine-check';
 jasmineCheck.install();
-
-import { is, List, Range, Record, Seq, SortedMap } from '../';
 
 describe('SortedMap', () => {
   it('converts from object', () => {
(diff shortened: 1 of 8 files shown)


bdurrer:
An interesting fork!

Intuitively, I kinda would expect Sorted and Ordered types be the same thing?
Maybe Sorted should replace/enhance the existing Ordered? Sort with an undefined comparator simply means that it doesn't sort upon initialization and uses "add last"?

You can already get what Sorted does by always calling `sort` after an operation. The performance might be much worse tho (didn't test it)

`Set([9,5,7,1]).sort(...);`

and

`.add(3).sort()`

jdeniau:
Possibly related to https://github.com/immutable-js/immutable-js/issues/88

jdeniau:
OK now I understand this PR.
It tooks me some time, but in the end I get it.

To be sure, conceptually, it's the same as calling `.sort()` before each operation?

Whereas `Ordered` only guaranty that the insert order is saved

```ts
Set(['a', 'c', 'b'])
// order might be 'b', 'a', 'c' as Set will optimize the indexing
OrderedSet(['a', 'c', 'b'])
// order will be 'a', 'c', 'b' because Ordered Set keep the order by design
SortedSet(['a', 'c', 'b'])
// order will be 'a', 'b', 'c' as each item will be sorted
```

PR #2004, opened 2024-07-22
Try to fix issue with empty list that is a singleton

### LINKED ISSUE #2003: Singleton `emptyList` is problematic
I just ran down a nasty bug related to this code below:

https://github.com/immutable-js/immutable-js/blob/493afba6ec17d9c999dc5a15ac80c71c6bdba1c3/src/List.js#L413

I did not realize that every time I was creating a new list with `List()`, I was getting the same _instance_ of `List`. This seems counterintuitive in a library based on the idea of immutability. 

Some code in the project was mutating an instance of an empty list which caused all subsequent newly minted lists to be non-empty. Obviously the code in my project was incorrect, but I came here to suggest _not_ using a singleton pattern here regardless.

jdeniau:
Fixes #2003

The EMPTY_LIST singleton is a really old code. It may have been there to avoid calling "makeList", but I don't think that there is a really huge performance impact.

### DIFF
--- a/__tests__/List.ts
+++ b/__tests__/List.ts
@@ -952,7 +952,11 @@ describe('List', () => {
   });
 
   it('Mutating empty list with a JS API should not mutate new instances', () => {
-    Object.assign(List(), List([1, 2]));
+    expect(() => {
+      Object.assign(List(), List([1, 2]));
+    }).toThrow(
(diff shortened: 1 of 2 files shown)


jdeniau:
ping @VirtuaBoza can you review this ?

PR #2011, opened 2024-08-01
check 5.x

jdeniau:
Create a PR to check that all 5.0.0 code is OK before merging in main.


iambumblehead:
Thank you for this great package. The rollup process generates minified and un-minified "umd" modules that are less useful these days. Please consider exporting minified esm instead.

jdeniau:
Hi @iambumblehead 

Theoretically, [the "module" key of package.json](https://github.com/immutable-js/immutable-js/blob/0c2d021552ee74022863f33ff3704f44c4862fde/package.json#L19) does target an ESM build.

I'm not really confident with releasing a non-umd build as there is a lot of users of immutable.
Maybe for the 6.x, but it will probably need a long "beta" time to have feedback from the users.

PR #2014, opened 2024-08-18
issue with tstyche and OrderedMap

jdeniau:
@mrazauskas There seems to be an issue with tstyche here : 

OrderedSet is not equal to Set (which is good), but OrderedMap is reported as same as Map, which is wrong.

Do you have an idea about why ?

Thanks

### DIFF
--- a/type-definitions/ts-tests/map.ts
+++ b/type-definitions/ts-tests/map.ts
@@ -1,5 +1,5 @@
 import { expect, test } from 'tstyche';
-import { Map, List, MapOf } from 'immutable';
+import { Map, List, MapOf, OrderedMap } from 'immutable';
 
 test('#constructor', () => {
   expect(Map()).type.toBe<Map<unknown, unknown>>();
@@ -638,4 +638,8 @@ test('#toJSON', () => {
   expect(Map({ a: Map({ b: 'b' }) }).toJSON()).type.toBe<{
     a: MapOf<{ b: string }>;
   }>();
+});
+
+test('OrderedMap != Map', () => {
+  expect<Map<string, string>>().type.not.toBe<OrderedMap<string, string>>();
 });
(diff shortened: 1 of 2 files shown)


mrazauskas:
Very interesting! I will take a deeper look later.

A quick check:
- `tsd` gave the same result (makes sense, because type compare logic is nearly the same);
- `expect-type` gave the [same result too](https://www.typescriptlang.org/play/?ts=5.5.4#code/JYWwDg9gTgLgBAbzgWQIZgDRwPJQCYCmUBeamcAygTFrocXlfAL5wBmUEIcA5KCAFcYqAEYAbAjwDcAKFCRYiOAQAeYAgGMYAFQCe67GzisOXXqvVaAtDH2SZMi5p13DAHjJuAzjCjAAdgDmWD5+QQB84QAUAJQAdP4QMHEwEACiAI4CqGJ6BmxudEQknqEBwXBlEdExso5qznkE7kzevuWRsQlJKelZOU3uRQytVYGdtTJAA) (it is implemented as a collection generic types, completely different from `tstyche` or `tsd`).

Since different tools agree, there should some difference in implementation of `OrderedMap` and `OrderedSet` that should explain this. I will look at that later.

mrazauskas:
@jdeniau Can I ask what you trying to achieve? (In terms of typings of type tests.)

TypeScript is a structural type system. Hence `interface A { foo: string }` and interface `interface B { foo: string }` are considered identical.

I was manually comparing `Map` and `OrderedMap` and seems like they both have identical properties. Hence they are identical.

In contrary,  the `OrderedSet` interface has `zip()`, `zipAll()`, `zipWith()` that do not exist on the `Set` interface. Hence these interface are reported as not identical.

---

Perhaps [branding](https://www.learningtypescript.com/articles/branded-types) is what you are looking for? Branding could make `Map` and `OrderedMap` not assignable to each others. And they would be not identical too.

mrazauskas:
Isn’t it that `Map` should not be assignable to `OrderedMap`? Hm.. Actually.. There is no difference? Or there is? I don’t know Immutable so well.

In TypeScript codebase there is a type named `Path`. Which is just a `string` with brand property, but if a function takes that normalized `Path`, TypeScript does not allow passing a `string`.

jdeniau:
I think that the typescript type are the same, even if the implementation differ.

But the implementation might differ one day. I don't see a case were branding OrderedMap would be usefull for end users, so I will keep it like this.

PR #2015, opened 2024-08-27
[Flow] Add undefined as a possible return value from find and findLast

Dagur:
The current definition always expects a value to be found. This has caused bugs in projects I have worked on.


jdeniau:
I merged this in 5.x as there probably won't be any version on 4.x anymore.
It will be released in 5.0.0.

PR #2016, opened 2024-08-28
Update flow definitions for first(), last(), find() and findLast()

Dagur:
I've updated `find` and `findLast` again because I think this is closer to the typescript definitions.

I also updated `first` and `last` to reflect the recent change in typescript.


jdeniau:
Hi, can you rebase on the 5.x branch and fix the conflicts?

PR #2020, opened 2024-09-23
upgrade microtime, resolve arm64 install failure

iambumblehead:
immutable-js' npm install succeeds on arm64 after microtime is incremented


jdeniau:
Merged in 5.x via commit [778c0d0](https://github.com/immutable-js/immutable-js/commit/778c0d0eaf7c00ad56400150b219d889a6293500)

PR #2024, opened 2024-10-07
Bump TSTyche

mrazauskas:
@jdeniau This is still fresh idea, but I am thinking to deprecated `.toMatch()` in the next version of TSTyche. It works well, but its behaviour is hard to explain.

It feels like `Omit` and `Pick` utility types should be enough for subtype testing. Well.. These work only with types so, I thought to add `omit()` and `pick()` utilities that are reshaping type of the given object.

Could I ask to take a quick look at updated tests? Isn’t it more clear what is going on? It would be very useful to get feedback before releasing RC.

---

As a side note, this assertion would pass:

```ts
expect<{ a: string; b: number }>().type.toMatch<{ a?: string | number | undefined }>();
```

Hm.. What? Yep, these are those quirks of `.toMatch()`. Nothing similar can happen with `Pick`.

Another aspect: user has to learn what `.toMatch()` does, in contrary `Pick` is already familiar.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,7 +52,7 @@
         "rollup": "3.28.1",
         "size-limit": "^8.2.6",
         "transducers-js": "0.4.174",
-        "tstyche": "^2.0.0",
+        "tstyche": "^3.0.0-beta.3",
         "typescript": "5.1"
       },
       "engines": {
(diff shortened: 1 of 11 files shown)


jdeniau:
Hi,
I don't see a real difference from my point of vue, but I do understand that Pick is a known TS utility type, and that it does fix a bug.

Does `pick` utility accept more than one value?
```ts
expect(pick(variable, 'size' | 'items')).toMatch...
```

mrazauskas:
> Does `pick` utility accept more than one value?

Yes, like this: `pick(object, 'foo', 'bar', 'baz')`. I have to document that, thanks for the input!

PR #2030, opened 2024-10-27
Fix List.VNode.removeAfter() bug

alexvictoor:
Hello!
I have added a failing test. I was not sure where to write it, let me know if it should be located elsewhere. I guess the bug comes from a bad copy/paste between VNode.removeBefore() & VNode.removeAfter().
Let me know if there is something missing in this PR.

### DIFF
--- a/__tests__/List.ts
+++ b/__tests__/List.ts
@@ -722,9 +722,17 @@ describe('List', () => {
     expect(o.get(0)).toBe('f');
   });
 
-  it('works with push and insert without phantom values', () => {
+  it('works with push, set and insert without phantom values', () => {
     const v = List.of().set(287, 287).push(42).insert(33, 33);
     expect(v.toJS().filter(item => item === 287)).toHaveLength(1);
+    const v2 = List.of().push(0).unshift(-1).unshift(-2).pop().pop().set(2, 2);
+    expect(v2.toJS()).toEqual([-2, undefined, 2]);
+    const v3 = List.of().set(447, 447).push(0).insert(65, 65);
+    expect(v3.toJS().filter(item => item === 447)).toHaveLength(1);
+    const v4 = List.of().set(-28, -28).push(0).shift().set(-30, -30);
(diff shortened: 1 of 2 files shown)


jdeniau (approved):

jdeniau:
Wow, this is an incredibly specific bug ! 😲 
Nice catch !

For the record I did not write that code, and I do not fully understand this method and the left shift operator.

If you did understand this: 
- could you explain what you did understand ?
- do you have another reproductible cas as I could not reproduce the issue with other cases than 287 and 33

alexvictoor:
Hello @jdeniau !
I guess you were right to ask, my PR is far from being perfect. I have just noticed some cases where not handle. I will fix that shortly. Also I might have found another issue in the removeBefore method.
To answer your question, I wanted to understand deeply the algorithm and the data structures used under the cover by ImmutableJS. Hence I have started to reimplement an immutable list from scratch (https://github.com/alexvictoor/ts-immutable). 
A few weeks ago I had the feeling that my implementation had the main features and I started to compare the behaviours of the main methods with the original ImmutableJS List using property based testing and wonderful fast-check library (thanks @dubzzz). I found a lot of bugs in my implementation using fast-check and then I have found this bug :)
Might be difficult to explain the code incriminated. If you like, since we are both french, we can schedule a google meet. FYI I  am willing to explain how ImmutableJS works under the cover at Paris Typescript meetup in a few months

jdeniau:
@alexvictoor great, I sent you an invite.
For the record I have the secret dream of converting immutable to TS 🤞

jdeniau:
@alexvictoor For the record, I did add the same test as `v` but testing the 10 000 first integers, and here is the list of the number that was failing, and that do pass now with your code:

``
33, 42, 95, 127, 159, 191, 223, 255, 287, 319, 351, 383, 415, 447, 479, 511, 543, 575, 607, 639, 671, 703, 735, 767, 799, 831, 863, 895, 927, 959, 991, 1023
```

jdeniau:
released in 5.0.3

PR #2034, opened 2024-11-04
Upgrade TSTyche

mrazauskas:
First of all, it should feel amazing to release `immutable@5.0.0`. Congrats!

TSTyche 3.0 is coming out later this week. Seems like current release candidate will be published as it is. Release notes can be found [here](https://tstyche.org/releases/tstyche-3).

As I was mentioning before, the `.toMatch()` matcher is deprecated. The `pick()` type utility can be used instead.

By the way, lowest supported Node.js version now is 18.19. That shouldn’t be a problem, or?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,7 +52,7 @@
         "rollup": "3.28.1",
         "size-limit": "^8.2.6",
         "transducers-js": "0.4.174",
-        "tstyche": "^2.1.1",
+        "tstyche": "^3.0.0-rc.2",
         "typescript": "5.1"
       },
       "engines": {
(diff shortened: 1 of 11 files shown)


jdeniau:
> First of all, it should feel amazing to release immutable@5.0.0. Congrats!

Thank you ! 

No problem for node 18, immutable is not use node api so it should work with old version of node.
For the record, our test stack does use [node 20](https://github.com/immutable-js/immutable-js/blob/5f0de66fccc33435c88477560869768229845ed3/.github/workflows/ci.yml#L21).

mrazauskas:
Thanks for quick review!

PR #2041, opened 2025-01-04
Avoid return in constructors

alexvictoor:
Hello!
Happy new year 2025!
This PR removes most "return" in constructors. The only one left is in Record.js.  
I have started to work on TypeScript JSDoc comments and noticed that return in constructor would make the migration of the codebase really difficult.  
Also my personal opinion is that removing return in constructors increase readability.  
Anyway, this PR is a prerequisite for a second PR that will come next which introduce JSDoc TypeScript comments in the codebase.  
I have tried to reduce as much as possible the impact on the API but there are still some small breaking changes:
Map, List, Set are now functions. They are not classes anymore. The related classes are MapImpl, ListImpl and SetImpl. Thoses new classes are not exported so IMHO the impact for ImmutableJS users is limited. 
The only issue is illustrated by the following code fragment:
```typescript
m instanceof Map;    // does not work anymore
Map.isMap(m);          // should be used instead
```

### DIFF
--- a/__tests__/List.ts
+++ b/__tests__/List.ts
@@ -111,7 +111,7 @@ describe('List', () => {
     expect(out.getIn(['c', 0])).toEqual('v');
     expect(out.get('a')).toBeInstanceOf(Array);
     expect(out.get('b')).toBeInstanceOf(Array);
-    expect(out.get('c')).toBeInstanceOf(Map);
+    expect(Map.isMap(out.get('c'))).toBe(true);
     expect(out.get('c')?.keySeq().first()).toBe(0);
   });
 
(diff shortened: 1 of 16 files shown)


jdeniau:
Map, List and Set classes are exported. See https://github.com/immutable-js/immutable-js/blob/d64444079618806d8ee6379177ec23b212230b24/src/Immutable.js#L60-L64

So unfortunately, I don't think that we should remove support for this.

alexvictoor:
Hello @jdeniau 
For the end user, everything still works as expected regarding Map, List and Set except "instanceof".  I am aware this is a breaking change but IMO this is a small one and most of all I have not found any better way to get rid of return statements in constructors. 
I think this is really important to remove these return statements because once this is done, TypeScript JsDoc annotations can be added to the code quite easily.

jdeniau:
@alexvictoor I did merge this on the 6.x branch

PR #2044, opened 2025-02-11
Detect change in dist file when pushing TS files

jdeniau:
This PR prepare the migration to TS by testing the output file :

It build the base branch and the current PR, and will fail if the builded file is different in than the base branch.

The purpose of that is to migrate files to TypeScript slowly without altering the runtime file. 
I hope that we can work this way without breaking runtime.

If we _need_ to break runtime, then we might want to add `@ts-expect-error` in main branch, and remove / fix it in the next major.

### DIFF
--- a/.github/workflows/output_diff.yml
+++ b/.github/workflows/output_diff.yml
@@ -1,0 +1,61 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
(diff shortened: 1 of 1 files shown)


jdeniau:
cc @alexvictoor

PR #2067, opened 2025-03-17
npm audit fix

mazerty:
i noticed during `npm install` several warnings due to "vulnerabilities"
the first commit is just running `npm audit fix` but a few errors remained
the second commit is with `npm audit fix --force` which updated the `package.json` file too

let me know if you're interested, i can remove the second commit if you prefer :)


mazerty:
i'm looking at the ci errors

PR #2072, opened 2025-03-19
Drop TS 4 support and add `const P extends` for getIn method types

jdeniau:
Use [const type parameters](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#const-type-parameters) for getIn path retrieval.

This feature has been introduced in TS 5.0, which is more than two years old now. This PR drops support for TS < 5.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
       - run: npm ci
       - run: npm run build
       - run: npm run test:unit
-      - run: npm run test:types -- --target 4.5,5.0,current
+      - run: npm run test:types -- --target 5.0,current
       - run: npx size-limit
       - run: npm run check-git-clean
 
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       - run: npm ci
       - run: npm run build
       - run: npm run test:unit
-      - run: npm run test:types -- --target 4.5,5.0,current
+      - run: npm run test:types -- --target 5.0,current
```
mrazauskas, line 69: Nice! To draw your attention, `tstyche@3.1.1` and above makes it possible to test agains a range of versions:

```suggestion
      - run: npm run test:types -- --target '>=5.0'
```

jdeniau: Nice, but doing that the tests ran 5 times slower. 
I do not know if it's worth it for now (but I do keep that in mind)

PR #2075, opened 2025-03-25
Update Range.js

ustinvaskin:
just added ternary "?" for better readability. Hope its all right.


jdeniau:
Thank you @ustinvaskin

PR #2085, opened 2025-03-27
Migrate from jest to vitest

jdeniau:
Now that jest-jasmine has been removed, let's migrate from jest to vitest.

## Why ?

- Because jest seems to be moving really slowly now (jest 29 is two years old, first alpha of 30 is 1.5 year old). Even if there is a recent alpha of 30, we don't know if it will resolve all pain of jest configuration.
- Because we do bundle immutable with rollup, and vitest does use rollup too, so the configuration is pretty straightforward, and we can remove a lot of hack.


jdeniau:
Superseeded by https://github.com/immutable-js/immutable-js/pull/2087

PR #2093, opened 2025-04-10
chore: update type tests

mrazauskas:
Primitive type matchers like `.toBeString()` or `.toBeBoolean()` will be removed in the next major TSTyche release. Reference: https://tstyche.org/releases/tstyche-4#matchers

This PR is replacing them with `.toBe()`.

---

I think that `.toBe<number>()` looks more constant. Makes it is easier to move from `.toBe<number>()` to `.toBe<Promise<number>>()`, if the behaviour is changing. It was not clear why using `.toBeNumber()` is anyhow better.

### DIFF
--- a/type-definitions/ts-tests/from-js.ts
+++ b/type-definitions/ts-tests/from-js.ts
@@ -33,13 +33,13 @@ test('fromJS', () => {
 });
 
 test('fromJS in an array of function', () => {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  /* eslint-disable @typescript-eslint/no-explicit-any */
   const create = [(data: any) => data, fromJS][1];
 
   expect(create({ a: 'A' })).type.toBe<any>();
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const createConst = ([(data: any) => data, fromJS] as const)[1];
+  /* eslint-enable @typescript-eslint/no-explicit-any */
 
   expect(createConst({ a: 'A' })).type.toBe<Map<'a', string>>();
 });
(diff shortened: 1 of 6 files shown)


type-definitions/ts-tests/from-js.ts
```diff
@@ -33,13 +33,13 @@ test('fromJS', () => {
 });
 
 test('fromJS in an array of function', () => {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  /* eslint-disable @typescript-eslint/no-explicit-any */
   const create = [(data: any) => data, fromJS][1];
 
-  expect(create({ a: 'A' })).type.toBeAny();
+  expect(create({ a: 'A' })).type.toBe<any>();
```
mrazauskas, line 39: Also lint rules apply for `.toBe<any>()`. That looks like a win, actually.

type-definitions/ts-tests/from-js.ts
```diff
@@ -33,13 +33,13 @@ test('fromJS', () => {
...
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const createConst = ([(data: any) => data, fromJS] as const)[1];
+  /* eslint-enable @typescript-eslint/no-explicit-any */
```
jdeniau, line 42: This seems weird 🤔

jdeniau (approved):

type-definitions/ts-tests/from-js.ts
mrazauskas: I though two comment lines are less than three (;

This is the same like `// eslint-disable-next-line` would be used on every line between `/* eslint-disable */` and `/* eslint-enable */`. That’s just another syntax. I can bring back `// eslint-disable-next-line` if you prefer.

jdeniau: I prefer the `next-line` version because it does not affect a whole range that can grow bigger in the future.

jdeniau:
Thanks @mrazauskas

PR #2100, opened 2025-04-29
add simple test to reproduce groupBy regression

giggo1604:
Hello,

While updating our codebase to the latest version of Immutable, we encountered what seems to be a regression in the behavior of `groupBy`.

This change was introduced in [Commit 53237d2](https://github.com/immutable-js/immutable-js/commit/53237d24a707e655ad060dd744a9bb9893915cba).

Previously, grouping by a function that might return `undefined` was tolerated, but in the latest version, this now throws. While the new behavior arguably makes sense—since grouping by `undefined` isn’t very clean—it does appear to be a breaking change, violating semantic versioning.

If this change is considered intentional, it should also be reflected in the type system: the `grouper` function should not be allowed to return `undefined`.

We’d like to clarify the intended direction:

- Was this change considered a bug fix?
- Is there any possibility that the previous behavior might be restored (i.e., loosening the new sanity check)?
- Or should we assume this is the expected behavior going forward and update our code accordingly?

Thanks in advance for your guidance, and thanks for maintaining the project!

### DIFF
--- a/__tests__/groupBy.ts
+++ b/__tests__/groupBy.ts
@@ -75,6 +75,13 @@ describe('groupBy', () => {
     expect(group.toJS()).toEqual({ odd: [1, 3, 5], even: [2, 4, 6] });
   });
 
+  it('allows `undefined` as a key', () => {
+    const group = Seq([1, 2, 3, 4, 5, 6]).groupBy((x) =>
+      x % 2 ? undefined : 'even'
+    );
+    expect(group.toJS()).toEqual({ undefined: [1, 3, 5], even: [2, 4, 6] });
+  });
+
   it('groups indexed sequences, maintaining indicies when keyed sequences', () => {
     const group = Seq([1, 2, 3, 4, 5, 6]).groupBy((x) => x % 2);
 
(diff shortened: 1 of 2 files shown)


jdeniau:
Hi,

Thanks for the bug report. I added that because I trusted typescript, but as immutable accept any value as key, having `undefined` as key should not be a problem.

See https://immutable-js.com/play/#IE1hcChbW3VuZGVmaW5lZCwgJ2EnIF1dICkKLmdldCh1bmRlZmluZWQp

TS might be more strict on undefined values though. I will check that, but it's clearly a regression.
I do not have access to a computer in the next few days though.

giggo1604:
Hi,

Thanks for your quick answer.

I added a quick fix and a test that caught the regression. Used @ts-expect-error for now since doing a proper fix would probably be a lot more work.

Long-term, it might make more sense to disallow undefined keys and introduce a breaking change. As this is more in line with how typescript works.

jdeniau:
Output diff failing test is under control as we change a TS file that does change the builded output.
This has been set to check that file migration from JS to TS does not alter the output in v5.

PR #2102, opened 2025-05-05
Revert previous assertion as it introduced a regression

jdeniau:
Superseed https://github.com/immutable-js/immutable-js/pull/2100

Revert a change introduced in [Commit 53237d2](https://github.com/immutable-js/immutable-js/commit/53237d24a707e655ad060dd744a9bb9893915cba).

Previously, grouping by a function that might return `undefined` was tolerated, but in the latest version, this now throws.

This PR does add @giggo1604 changes, but improve the TS typing directly


jdeniau:
Output diff failing test is under control as we change a TS file that does change the builded output.
This has been set to check that file migration from JS to TS does not alter the output in v5.

PR #2131, opened 2025-06-30
Extract CollectionHelperMethods

jdeniau:
Take https://github.com/immutable-js/immutable-js/commit/d7a41f78eda15f7d54e6f9b6ea518f1671d38bf2#diff-e07884517d664b4d4195610350d28217383ae084398b98be4eba6d0658548489 from 6.x into main. 
Extract some other utils that can be converted to TS.


jdeniau:
Output diff is handled (simple position change)

PR #2134, opened 2025-07-05
fix(src):fix some ts error hint

JayMeDotDot:
`src/Collection.ts`
I changed generic type `T` to `unknown`; otherwise, ts will lint a `Type '(Iterable<T> | ArrayLike<T>) & IndexedCollectionImpl<unknown>' is not assignable to type 'IndexedCollectionImpl<T>` error message.

`src/predicates/isKeyed.ts`
You left a to-do, so I changed it.

`src/utils/mixin.ts`
`PropertyKey` has a type of `number | string | symbol`.
I don't know why `Object.getOwnPropertySymbols` needs to be an assertion.
Removing it, the ESLint error message automatically disappears.

### DIFF
--- a/src/Collection.ts
+++ b/src/Collection.ts
@@ -107,9 +107,9 @@ export function KeyedCollection(
 
 export class KeyedCollectionImpl<K, V> extends CollectionImpl<K, V> {}
 
-export function IndexedCollection<T>(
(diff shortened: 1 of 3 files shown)


jdeniau:
OK I got the error in TS 5.9. This should be resolve in #2139

PR #2142, opened 2025-09-15
Remove empty collection singletons

### LINKED ISSUE #2140: Undocumented Breaking Change in v5: Inconsistent factory references for immutable collections
As per the documentation, every immutable collection (`Map`, `List`, `Set`, etc) has its equivalent factory function that creates a **new** immutable object. However, there appear to be inconsistencies across versions.

In version 4:

```
List() === List() // True
Map() === Map() // True
Set() === Set() // True
...
```
All return the same value, but each of this calls should generate a new immutable object for which reference equality should be different.

In version 5:

```
List() === List() // False
Map() === Map() // True
Set() === Set() // True
...
```
`List` now correctly returns a new object, but `Map` and `Set` still return the same reference. This is closer to what you can expect from the documentation, but still is a possible breaking change for `List` that wasn’t documented in the release notes.

It might seem like a no issue because most of the time any manual comparison will and should be done with `equals()` or `is()`. However, many common libraries like React, Angular or Redux rely on reference equality for performance optimizations and change detection. A common scenario where this can be a problem is when dealing with default values: they very often use empty collections and this can cause subtle bugs that are quite tricky to catch.

6uzm4n:
Fixes #2140 

All empty collection singletons have been removed. Added tests for instance inequality on  immutable objects with value equality.

### DIFF
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -26,6 +26,13 @@ describe('Map', () => {
     expect(m.get('c')).toBe('C');
   });
 
+  it('constructor provides different instances', () => {
+    expect(Map()).not.toBe(Map());
+    expect(Map()).toEqual(Map());
+    expect(Map({ a: 'A' })).not.toBe(Map({ a: 'A' }));
+    expect(Map({ a: 'A' })).toEqual(Map({ a: 'A' }));
+  });
+
   it('constructor provides initial values', () => {
     const m = Map({ a: 'A', b: 'B', c: 'C' });
     expect(m.size).toBe(3);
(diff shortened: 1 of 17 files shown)


jdeniau:
Sorry I forgot to mention that there is a `6.x` branch to target here!
Can your rebase your code? You may have some conflict though as we are migrating the codebase to Typescript though.

6uzm4n:
Whoops, my bad! Now the PR is dirtier than what I would like but I think I didn't miss anything while rebasing

jdeniau:
Thanks @6uzm4n 👍

PR #2266, opened 2026-08-16
Resolve the three remaining likely-bugs findings (R1-R4)

jdeniau:
Each finding from .agents/migration-status.md was traced back to its originating PR, then fixed or confirmed as an intentional change:

- R1 (isSubset string argument, from #2204): keep the new character-iteration semantics, consistent with isSuperset and with Collection('abc'); documented as a v6 breaking change in the CHANGELOG and pinned by multi-char unit tests.
- R2 ('@@iterator' fallback removal, from #2127): confirmed intended modernisation; explicit CHANGELOG entry, and the two immutable.d.ts docstrings no longer mention @@iterator. immutable.js.flow is left untouched since Flow's @@iterator syntax denotes Symbol.iterator.
- R3 (reverseFactory reverse-iteration keys, ?? 0 from #2206, NaN keys already in v5): fixed by counting down from the number returned by ensureSize(collection) - the reversed sequence has the same size as its source. Regression tests cover both __iterate (reduceRight) and __iterator (keyed reverse) paths; CHANGELOG entry since the NaN behaviour shipped in v5.
- R4 (wholeSlice with begin === undefined, from #2128): restored the faithful v5 form (begin !== undefined && begin <= -size) so slice() with no arguments goes through sliceFactory again, on empty collections too; identity regression test added.

Claude-Session: https://claude.ai/code/session_0122shfP2ngkkjuXZBWoeFGW

### DIFF
--- a/__tests__/Set.ts
+++ b/__tests__/Set.ts
@@ -301,6 +301,16 @@ describe('Set', () => {
     expect(Set.of('a', 'b').isSuperset('ab')).toBe(true);
   });
 
+  it('treats a string argument of isSubset as a collection of characters', () => {
+    // Since v6 a string is iterated character by character, like everywhere
+    // else in Immutable (and like `isSuperset` does). It is no longer passed
+    // to `String.prototype.includes`, which had substring semantics:
+    // 'ab' is not one of the characters of 'abc'.
+    expect(Set.of('ab').isSubset('abc')).toBe(false);
+    expect(Set('abc').isSubset('abc')).toBe(true);
+    expect(Set.of('ab').isSuperset('abc')).toBe(false);
+  });
+
   describe('accepts Symbol as entry #579', () => {
     it('operates on small number of symbols, preserving set uniqueness', () => {
       const a = Symbol();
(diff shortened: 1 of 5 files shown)


__tests__/Set.ts
```diff
@@ -301,6 +301,16 @@ describe('Set', () => {
...
   });
 
+  it('treats a string argument of isSubset as a collection of characters', () => {
+    // Since v6 a string is iterated character by character, like everywhere
+    // else in Immutable (and like `isSuperset` does). It is no longer passed
+    // to `String.prototype.includes`, which had substring semantics:
+    // 'ab' is not one of the characters of 'abc'.
+    expect(Set.of('ab').isSubset('abc')).toBe(false);
+    expect(Set('abc').isSubset('abc')).toBe(true);
+    expect(Set.of('ab').isSuperset('abc')).toBe(false);
```
jdeniau:
```suggestion
    expect(Set(['ab']).isSubset('abc')).toBe(false);
    expect(Set('abc').isSubset('abc')).toBe(true);
    expect(Set(['ab']).isSuperset('abc')).toBe(false);
```

PR #2284, opened 2026-09-08
Fix Map.update so a stored __proto__ key can change

### LINKED ISSUE #2283: Update function does not work with __proto__ key
# What happened

A property-based test caught this edge case in my project and I traced the cause to Immutable Map update function.

# How to reproduce

```ts
Map()
  .set("__proto__", "set")
  .update("__proto__", (u) => "updated")
  .get("__proto__") // expected: "updated", actual: "set"

```

https://immutable-js.com/play/#TWFwKCkKICAuc2V0KCJfX3Byb3RvX18iLCAic2V0IikKICAudXBkYXRlKCJfX3Byb3RvX18iLCAodSkgPT4gInVwZGF0ZWQiKQogIC5nZXQoIl9fcHJvdG9fXyIpIC8vIGV4cGVjdGVkOiAidXBkYXRlZCIsIGFjdHVhbDogInNldCIK

official-burak:
`Map.set('__proto__', ...)` already stores that key. `update()` goes through the shared functional `set()`, which skipped `__proto__` and `constructor` on every collection, so the later update was a no-op.

The skip stays for plain objects, where it still blocks prototype pollution.

Fixes #2283

### DIFF
--- a/__tests__/Map.ts
+++ b/__tests__/Map.ts
@@ -608,6 +608,21 @@ describe('Map', () => {
     expect(m.toObject().admin).toBeUndefined();
     expect(m.toJS().admin).toBeUndefined();
   });
+
+  it('update can change a __proto__ key stored on the map', () => {
+    const m = Map()
+      .set('__proto__', 'set')
+      .update('__proto__', () => 'updated');
+    expect(m.get('__proto__')).toBe('updated');
+    expect(m.size).toBe(1);
+  });
+
+  it('update can change a constructor key stored on the map', () => {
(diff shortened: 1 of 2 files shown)


lbesecker195 (approved):
Verified: the `isProtoKey` early return sat before the `isImmutable(collection)` branch, so `update('__proto__', ...)` on a Map that already stores that key was a silent no-op while `set('__proto__', ...)` worked. Moving the guard below that branch keeps the plain-object protection intact. Tests cover both keys. LGTM.

